PR #78, opened 2019-05-22
SxS In-Proc Support for Multiple Profiler Instances

jander-msft:
Allow multiple instances of CProfilerManager to be created and initialized.
Pass CProfilerManager instances rather than using static instance.
Change CMethodInfo static cache into instance field.
Remove unused statics.


jander-msft:
Changes in this PR addressed issue #140

PR #99, opened 2019-06-29
Migrate private EULA to public MIT License for OSS

WilliamXieMSFT:


LICENSE
```diff
@@ -1,51 +1,23 @@
-MICROSOFT SOFTWARE LICENSE TERMS
+The MIT License (MIT)
 
-MICROSOFT INSTRUMENTATION ENGINE SOFTWARE DEVELOPMENT KIT
+Copyright (c) .NET Foundation and Contributors
```
delmyers: Why .NET foundation? Shouldn't this be Microsoft?

WilliamXieMSFT:
Consider adding pre-commit hooks to verify/check files contain the header

LICENSE
WilliamXieMSFT: Correct, thanks for the catch!

verifyCopyrightHeaders.ps1
```diff
@@ -0,0 +1,114 @@
...
+    '.snk'
+    '.txt'
+    '.user'
+)
+
+$excludedFiles = @(
+    # external
+    'corprof.h'
+    'dbgmsg.h'
+
```
delmyers: Do we not have a separate folder for externals?

verifyCopyrightHeaders.ps1
```diff
@@ -0,0 +1,114 @@
...
+Write-Verbose "    $copyrightHeader"
+Write-Verbose "    $mitLicenseHeader"
+Write-Verbose ""
+
+Write-Verbose "Checking files under directory '$PSScriptRoot' ..."
+Write-Verbose "Non-compliant files will be stored in $resultFile"
+
+Get-ChildItem -Path $PSScriptRoot -Recurse -File | 
+    Where-Object { $_.PSParentPath -notmatch "\\$([String]::Join('|', $excludedDirs))" } | 
+    Where-Object { !$excludedFiles.Contains($_.Name) } |
```
delmyers: I thought that the standard syntax was -not

verifyCopyrightHeaders.ps1
```diff
@@ -0,0 +1,114 @@
...
+                $lines[0].ToString().Contains($shellHeader))
+            {
+                $copyrightIndex = 1
+                $mitLicenseIndex = 2
+            }
+
+            if (!$lines[$copyrightIndex].ToString().Contains($copyrightHeader) -or
+                !$lines[$mitLicenseIndex].ToString().Contains($mitLicenseHeader))
+            {
+                "Could not find OSS header in $($_.FullName)" > $resultFile
```
delmyers: You may consider just writing to stdout and stderr, and piping to a file instead of writing directly to the file.

verifyCopyrightHeaders.ps1
WilliamXieMSFT: Not at the moment. I can move these into one update the refs in a separate checkin.

PR #101, opened 2019-07-02
Use separate event for signaling finish event logging processing.

jander-msft:
Use separate event for signaling finish event logging processing.
Have file sink only report flags if file was opened.

### DIFF
--- a/src/Common.Lib/EventLoggingBase.cpp
+++ b/src/Common.Lib/EventLoggingBase.cpp
@@ -13,7 +13,11 @@ CEventLoggingBase::CEventLoggingBase() :
     // released quickly and we avoid context switching.
     m_cs(10),
     m_eventQueueLength(0),
-    m_hEventQueueEvent(FALSE, FALSE), // auto-reset, initially unsignaled
+    // Event used to tell the processing thread to process the current items
+    // on the queue.
+    m_hEventQueueProcessEvent(FALSE, FALSE), // auto-reset, initially unsignaled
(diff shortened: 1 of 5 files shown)


src/Common.Lib/EventLoggingBase.cpp
```diff
@@ -13,7 +13,12 @@ CEventLoggingBase::CEventLoggingBase() :
     // released quickly and we avoid context switching.
     m_cs(10),
     m_eventQueueLength(0),
-    m_hEventQueueEvent(FALSE, FALSE), // auto-reset, initially unsignaled
+    // Event used to tell the processing thread to process the current items
+    // on the queue.
+    m_hEventQueueProcessEvent(FALSE, FALSE), // auto-reset, initially unsignaled
+    // Event is initially signaled to allow TerminateEventSource to complete successfully
+    // if event source thread is not started.
+    m_hEventQueueFinishedEvent(TRUE, TRUE), // manual-reset, initially signaled
```
WilliamXieMSFT: Shouldn't this be initially unsignaled?

jander-msft: If it's initially unsignaled and the event source thread is never created (because, for example, error logging is NOT enabled), then shutdown will hang waiting for the thread to signal this event.

jander-msft: Fixed this by checking if the thread was created before waiting on the event. I've changed the event back to being unsignaled initially.

PR #104, opened 2019-07-11
Implement ICorProfilerInfo10

jander-msft:


jander-msft:
@delmyers, I wanted to verify with you that I did not need to bump CLR_INSTRUMENTATION_ENGINE_API_VER because there isn't any API surface change to the Instrumentation Engine, correct?

jander-msft:
It looks like ICorProfilerInfo10 is still in flux. Over time, people have added and remove methods from it during the .NET Core 3.0 time frame. Do we need to consider that here somehow? We probably don't want to ship the Instrumentation Engine with this interface potentially in-flux.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Please format the changes as follows:
 + BugFixes:
 + Updates:
 
+# 1.0.26
```
WilliamXieMSFT: I recently version bumped to 1.0.25 , we don't need to version bump for this change since master has 1.0.24 still

jander-msft: This is targeting master, thus master has 1.0.25 before this change. I'm bumping the version number so that we produce packages in higher precedence that what was released as 1.0.25

inc/clr/prof/corprof.h
```diff
@@ -14868,537 +14876,1274 @@ EXTERN_C const IID IID_ICorProfilerInfo9;
...
+#ifndef __ICorProfilerInfo10_INTERFACE_DEFINED__
+#define __ICorProfilerInfo10_INTERFACE_DEFINED__
 
-/* interface ICorProfilerMethodEnum */
+/* interface ICorProfilerInfo10 */
 /* [local][unique][uuid][object] */ 
 
 
-EXTERN_C const IID IID_ICorProfilerMethodEnum;
+EXTERN_C const IID IID_ICorProfilerInfo10;
```
delmyers: >EXTERN_C const IID IID_ICorProfilerInfo10 [](start = 0, length = 41)

Are we updating the version of the PAL as well? If we aren't then I'm unsure whether or not IID_ICorProfilerInfo10 will be linked into the final .a file, which may cause undefined behavior.

delmyers:
Good question. I could go either way on this. There is a new api surface in that we support ICorProfilerInfo10.  But you could also figure that out just by QI'ing, which we will need to do anyway.

Maybe we should stick to only bumping the api version when we make idl changes?

---
In reply to: [510297969](https://github.com/microsoft/CLRInstrumentationEngine/pull/104#issuecomment-510297969) [](ancestors = 510297969)

CHANGELOG.md
WilliamXieMSFT: This PR is targeting develop branch, which is should be fine for testing since the only difference between develop and master are the MIT license change and nit bugfixes.

delmyers:
@noahfalk is added to the review. #Closed

delmyers:
Check with Noah about it. There are already Core 3 previews, so I doubt that the API surface should be changing.

---
In reply to: [510303048](https://github.com/microsoft/CLRInstrumentationEngine/pull/104#issuecomment-510303048) [](ancestors = 510303048)

CHANGELOG.md
jander-msft: Sorry, yes, I meant develop. I sometimes get tripped up because of the similar-yet-different branch name schemes among various repos.

jander-msft:
I agree.

---
In reply to: [510559890](https://github.com/microsoft/CLRInstrumentationEngine/pull/104#issuecomment-510559890) [](ancestors = 510559890,510297969)

noahfalk:
cc @davmason (the main dev on our team doing profiler work)

Our general .Net Core policy is that breaking changes between preview releases is legal, but at some point nearing the end of the release we will declare a point where breaking changes are no longer allowed. In .Net Core 3.0 we promised to stop breaking changes when we delivered Preview7, which should be occurring any day now. I believe the interface you have here matches the one that will be shipped in Preview7 and thus you can rely on it being final.

However this does leave open the question of how you want to handle future releases. If you are shipping CLRIE with access to preview .Net Core features then you may want to treat those versions of CLRIE as also being previews with the same policy about allowing breaking changes in the preview time period.

inc/clr/prof/corprof.h
jander-msft: The latest PAL does not have ICorProfilerInfo10 defined. I don't think we are using this symbol anywhere, so what I'm thinking is that I can try to undefine it, make it a compiler error if used (not sure if that it possible), or redefine it with another header file (we know what the value should be). I'm thinking the last option.

delmyers: We may or may not be using the symbol. It depends on the way that uuidof is implemented in our version of clang. I think some older versions just defined it as:

#define uuidof(interface) IID_##interface

so, uuidof(ICorProfilerInfo10) would just resolve to IID_ICorProfilerInfo10.

You can't undefine a symbol, unfortunately. The problem will be with the linker. If there is no value set for the extern, then there will just be a random value put in place of the address of the const variable, and your behavior is undefined. The most likely thing is that the QI will just fail because you'll have a garbage GUID, but that is no guarantee.

You could just put it in our own cpp file. I think the linker will complain if there is an attempt to link the same symbol more than once.

jander-msft: If I don't define it and we try to use it in our code, then we'll get a linker error when building on Windows stating "error LNK2001: unresolved external symbol IID_ICorProfilerInfo10"; on Linux, there will not be an error because of the delayed symbol resolution.

If I do define it, there's nothing that is using it, so I don't have anything that actually verifies that the definition satisfies the extern declaration. Ignoring that, I'm not sure I can put it in a shared cpp file since I have to reference stdafx.h, which is in a different location for each project. Creating another library project just for this seems a little overkill.

I'm reasonably certain that using CComQIPtr<ICorProfilerInfo10> or QueryInterface with __uuidof(ICorProfilerInfo10) works without this being defined for both Windows and Linux (I'll reconfirm that).

I'm leaning toward not defining this since we aren't using it and we would very likely get a linker error when we do start using it.

delmyers: We are only not using the symbol because we are using uuidof. If clang actually looks at the declspec in our version, then we will continue to not use it. But, if it doesn't, then we are already using it in the Linux build in CorProfilerInfoWrapper.cpp. That's where I'm worried about the undefined behavior.

delmyers: After some discussion... it seems that clang has support for __uuidof? I can't find anything in the official docs about it, though. I'll sign off.

PR #105, opened 2019-07-16
Fix bug where preinstalled site extension is empty

WilliamXieMSFT:
Without this fix, the preinstalled site extension .zip file can't resolve filepaths.

### DIFF
--- a/src/InstrumentationEngine.Preinstall/InstrumentationEngine.Preinstall.csproj
+++ b/src/InstrumentationEngine.Preinstall/InstrumentationEngine.Preinstall.csproj
@@ -47,27 +47,29 @@
         <Destination>$(PackageVersion)</Destination>
       </ZipItem>
       <!-- Instrumentationx86Files -->
-      <ZipItem Include="$(InputBinCfgRoot)\x86\InstrumentationEngine\**"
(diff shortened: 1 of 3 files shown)


src/InstrumentationEngine.Preinstall/InstrumentationEngine.Preinstall.csproj
```diff
@@ -3,6 +3,7 @@
       -->
 <Project>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), EnlistmentRoot.marker))\build\Common.props" />
+  <Import Project="$(EnlistmentRoot)\build\Packaging.props" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <PropertyGroup>
```
wiktork: did you need to version bump?

WilliamXieMSFT: Luckily not for this change

WilliamXieMSFT: It appears this only fixes local build but not from ADO. Currently investigating...

PR #108, opened 2019-07-18
Fixed bug in duplicate file check

WilliamXieMSFT:
Check against the final destinationFileName+destinationFolderName instead of just destinationFileName as duplicate files can be dropped in different destination folders.

### DIFF
--- a/build/Tasks/ZipArchiveTask.cs
+++ b/build/Tasks/ZipArchiveTask.cs
@@ -41,6 +41,12 @@ namespace Microsoft.VisualStudio.ProductionDiagnostics.BuildTasks
                     foreach (ITaskItem fileItem in Files)
                     {
                         string destinationFileName = new FileInfo(fileItem.ItemSpec).Name;
+                        string destinationFolder = fileItem.GetMetadata("Destination");
+                        string type = fileItem.GetMetadata("Type");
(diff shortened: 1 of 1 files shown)


WilliamXieMSFT:
I'll be adding a package verification script so these sort of issues don't come up again.

build/Tasks/ZipArchiveTask.cs
```diff
@@ -41,6 +41,12 @@ public override bool Execute()
...
                         string destinationFileName = new FileInfo(fileItem.ItemSpec).Name;
+                        string destinationFolder = fileItem.GetMetadata("Destination");
+                        string type = fileItem.GetMetadata("Type");
+                        if (!string.IsNullOrEmpty(destinationFolder))
+                        {
+                            destinationFileName  = Path.Combine(destinationFolder, destinationFileName);
+                        }
 
                         // Prevent duplicate files from getting zipped.
                         if (uniqueFiles.Contains(destinationFileName))
```
wiktork: Perhaps this should be an error instead?

WilliamXieMSFT: For local builds, we get two copies of each file since one appears in the x64\ folder and another aperas in x64\InstrumentationEngine. I can address your comment in the next iteration.

PR #109, opened 2019-07-18
Remove Shutdown() calls from destructors

WilliamXieMSFT:
This causes deadlocking on EventLoggingBase.cpp's TerminateEventSource() which attempts to get a handle to the CCriticalSection that's been discarded prior to static deconstructor call. This fix here is to remove the static deconstructors from calling shutdown.

### DIFF
--- a/src/Common.Lib/EventLoggingBase.cpp
+++ b/src/Common.Lib/EventLoggingBase.cpp
@@ -25,12 +25,6 @@ CEventLoggingBase::CEventLoggingBase() :
 
 CEventLoggingBase::~CEventLoggingBase()
 {
-    CCriticalSectionHolder holder(&m_cs);
-
-    if (m_hEventSource)
-    {
-        TerminateEventSource();
-    }
 }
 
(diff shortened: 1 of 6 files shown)


src/Common.Lib/EventLoggingBase.cpp
```diff
@@ -155,16 +138,8 @@ HRESULT CEventLoggingBase::TerminateEventSource()
         }
     }
 
-    // Allow thread to drain the queue. Normally, we'd just wait for the thread
-    // to finish (wait on the thread) but the thread is deadlocking on CoUninitialize.
-    // As a workaround, use a seperate event that the thread will signal when it is
-    // finished processing items and no longer using the event source.
-    // Only wait on the handle if the thread was created. This avoids situations where
-    // event logging was not initialized but TerminateEventSource was still called.
-    if (m_hEventQueueThread)
```
jander-msft: I would still check that the thread exists before waiting on the thread just in case the thread wasn't started or failed to start.

PR #110, opened 2019-07-18
RI master back to develop

WilliamXieMSFT:


WilliamXieMSFT:
This PR is unnecessary, we should be able to directly RI merge master to develop

PR #112, opened 2019-07-26
Add null checks for MethodName BSTRs

WilliamXieMSFT:
Obfuscated dlls can caus method names to be empty -- MetadataImport::GetMethodProps() returns nullptr

Thank you Michael Baranov and the SiteCore team for helping us uncover this bug!

### DIFF
--- a/src/InstrumentationEngine/MethodInfo.cpp
+++ b/src/InstrumentationEngine/MethodInfo.cpp
@@ -205,7 +205,7 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::InitializeFullName()
 {
     HRESULT hr = S_OK;
 
-    if (m_bstrMethodFullName.Length() != 0)
+    if (m_bstrMethodFullName != nullptr && m_bstrMethodFullName.Length() != 0)
     {
         return S_OK;
     }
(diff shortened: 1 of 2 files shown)


src/InstrumentationEngine/MethodInfo.cpp
```diff
@@ -246,14 +246,24 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::InitializeFullName()
...
+            declaringTypeName != nullptr &&
+            declaringTypeName.Length() > 0)
         {
             nameBuilder << (LPWSTR)declaringTypeName << _T(".");
         }
     }
-    nameBuilder << (LPWSTR)m_bstrMethodName;
-    m_bstrMethodFullName = nameBuilder.str().c_str();
 
+    if (m_bstrMethodName != nullptr)
```
wiktork: Can we document/verify the expectations around m_bstrMethodName and m_bstrMethodFullName? It seems like m_bstrMethodName can be null but m_bstrMethodFullName is assigned the empty string. What do we want IMEthodInfo callers to get for these cases when they request the method name?

WilliamXieMSFT: I'm refactoring this so that m_bstrMethodName will never be null (ie. set to empty string). This essentially ensures all callers will get an empty string.

PR #117, opened 2019-08-02
Reduce error logging noise for .NET framework < 4.8

WilliamXieMSFT:


src/InstrumentationEngine/MethodJitInfo.cpp
```diff
@@ -79,7 +79,16 @@ namespace MicrosoftInstrumentationEngine
         {
             // Note, this will return E_NOTIMPL before framework 4.8, so we don't
             // assert on failure.
-            IfFailRet(pProfilerInfo4->GetILToNativeMapping2(m_functionId, m_rejitId, cMap, pcNeeded, pMap));
+            if (FAILED(hr = (pProfilerInfo4->GetILToNativeMapping2(m_functionId, m_rejitId, cMap, pcNeeded, pMap))))
+            {
+                if (hr != E_NOTIMPL)
+                {
+                    LogErrorFor(pProfilerInfo4->GetILToNativeMapping2(m_functionId, m_rejitId, cMap, pcNeeded, pMap));
```
wiktork: I think it's perfectly ok to just say:
if (hr != E_NOTIMPL)
{
   CLogging::LogError(_T("Unexected error when calling GetILToNativeMapping…%80x"), hr) or something similar.
}

The current code looks like you are calling the function twice.

WilliamXieMSFT: I agree with your point that the code looks like it's invoking the function twice; the macro just stringifies it but it's not obvious.

I've also considered modifying IfFailRet to take in an "excluded" hr and write error if it's not excluded.

---
In reply to: [310199521](https://github.com/microsoft/CLRInstrumentationEngine/pull/117#discussion_r310199521) [](ancestors = 310199521)

PR #127, opened 2019-08-23
Add bitness to the RawProfilerHook variables

WilliamXieMSFT:
Also allow RPH even if no IM config files found.


docs/getting_started.md
```diff
@@ -78,4 +78,10 @@ A simple example of an InstrumentationMethod can be found in
...
 
-See [Configuration](configuration.md) for more details
\ No newline at end of file
+See [Configuration](configuration.md) for more details
+
+### How do I use the Raw Profiler Hook?
+
+The Raw Profiler Hook is the
+
+TODO willxie
```
WilliamXieMSFT: Adding documentation in next iteration

src/ExtensionsHostLib/RawProfilerHookSettingsReader.cpp
```diff
@@ -12,8 +12,13 @@ namespace Agent
 {
 namespace Host
 {
-    LPCWSTR cszRawProfilerHookVariableName = L"MicrosoftInstrumentationEngine_RawProfilerHook";
-    LPCWSTR cszRawProfilerHookPathVariableName = L"MicrosoftInstrumentationEngine_RawProfilerHookPath";
```
jander-msft: This is technically a breaking change, but I don't know how useful this was previously considering that it could try to load the assembly with a different bitness and fail. Would it be useful to check for the variable with the bitness in its name first and then fallback to the one without the bitness in its name, to preserve backward compatibility?

tests/RawProfilerHook/RawProfilerHook.Tests/TestEngine/TestEngineNoBitness.cs
```diff
@@ -0,0 +1,202 @@
...
+    using System.IO;
+    using System.Text;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using RemoteUnitTestExecutor;
+    using System.Threading;
+
+    /// <summary>
+    ///     Summary description for TestsExtensionHostLoading
+    /// </summary>
+    public class TestEngineNoBitness : RemoteUnitTestExecutor.TestEngineBase
```
jander-msft: If I understand correctly, this class is the same as TestEngine, just will slightly different environment variables. Could you fold this back into TestEngine and just have the test configure whether it should use bitness variables or not (maybe as a boolean property on TestEngine, which is read in by OnBeforeStarted)?

src/ExtensionsHostLib/RawProfilerHookSettingsReader.cpp
```diff
@@ -13,7 +13,12 @@ namespace Agent
 namespace Host
 {
     LPCWSTR cszRawProfilerHookVariableName = L"MicrosoftInstrumentationEngine_RawProfilerHook";
-    LPCWSTR cszRawProfilerHookPathVariableName = L"MicrosoftInstrumentationEngine_RawProfilerHookPath";
```
delmyers: >cszRawProfilerHookPathVariableName [](start = 12, length = 34)

Remove this. You overwrite it below.

src/ExtensionsHostLib/RawProfilerHookSettingsReader.cpp
```diff
@@ -52,18 +55,31 @@ namespace Host
...
-            L"Examining environment variable, name:",
+            L"Examining environment variable",
             Param(L"name", cszRawProfilerHookPathVariableName));
 
         IfFailRetHresult(
             env.GetEnvironmentVariable(
                 cszRawProfilerHookPathVariableName,
                 strRawProfilerHookModulePath), hr);
+
+        if (S_FALSE == hr)
```
delmyers: >S_FALSE == hr [](start = 12, length = 13)

Add a comment that you are doing this for backward compatibility.

Issue #130, opened 2019-09-06
Can't debug instrumentation method initialization when using retail binaries.


delmyers:
ClrInstrumentationEngine allows you to debug your instrumentation method by setting an environment variable to wait for a debugger attach. This is not practical for customers, though, because the ProfilerManager.cpp sets up a wait time that is only overridden for debugging when the dll is compiled for debugging.

Though clients can attempt to locate the correct release tag, and rebuild with debug if they are on Windows, I expect most clients would rather just use retail dlls, and debug their own code.

I suggest that we remove the #ifdef Debug for the override of the wait time.

PR #131, opened 2019-09-06
Allow clients to debug Instrumentation Method initialization with retail ClrInstrumentationEngine

delmyers:


delmyers:
This is a fix for bug #130

PR #134, opened 2019-09-09
Publish symbols to symweb

WilliamXieMSFT:
This publishes symbols to the Azure DevOps project associated with the build. For now, that build is internal to Microsoft. Once we have onboarded to the public dev.azure.com/ms/clrinstrumentationengine project, these symbols should be accessible by anyone.


build/yaml/jobs/windows/binaries.yaml
```diff
@@ -5,6 +5,7 @@
 
 parameters:
   IndexSources: false
+  PublishSymbols: false
```
delmyers: >PublishSymbols: false [](start = 2, length = 21)

Is there a pipeline that sets this to true? I don't see it in the code. If there isn't a pipeline that sets this, then we will never publish symbols.

WilliamXieMSFT: Yes, this is set in the Jobs\binaries.yaml and is added for Signed Release builds

---
In reply to: [333118445](https://github.com/microsoft/CLRInstrumentationEngine/pull/134#discussion_r333118445) [](ancestors = 333118445)

build/yaml/jobs/binaries.yaml
```diff
@@ -19,8 +19,9 @@ jobs:
       CopyFilters: |
         **
         !**\Nagler*
-        !**\*Tests*
-        !**\*Tests*\**
+        !**\*Test*
+        !**\*Test*\**
+      PublishSymbols: true
```
WilliamXieMSFT: >PublishSymbols: true [](start = 6, length = 20)

I'll look into adding this for Linux as well.

delmyers: Does this mean that we will be publishing symbols for all builds? Even PRs and CIs? Seems like we should only be doing it for the real signed build.

---
In reply to: [333181904](https://github.com/microsoft/CLRInstrumentationEngine/pull/134#discussion_r333181904) [](ancestors = 333181904)

WilliamXieMSFT: No, this is only for when SignType is set (to Test or Real). I can make it strictly for Real signed.

---
In reply to: [333184937](https://github.com/microsoft/CLRInstrumentationEngine/pull/134#discussion_r333184937) [](ancestors = 333184937,333181904)

delmyers: Didn't see the if statement above.

---
In reply to: [333184937](https://github.com/microsoft/CLRInstrumentationEngine/pull/134#discussion_r333184937) [](ancestors = 333184937,333181904)

WilliamXieMSFT: I'll add more comments here for clarity

---
In reply to: [333216207](https://github.com/microsoft/CLRInstrumentationEngine/pull/134#discussion_r333216207) [](ancestors = 333216207,333184937,333181904)

WilliamXieMSFT:
>           Platform: 'x86'

We can build the name of each strategy based on these parameters.

---
Refers to: build/yaml/jobs/binaries.yaml:29 in fb2983e. [](commit_id = fb2983e2c382144305d342d428da8e6b7b385af0, deletion_comment = False)

WilliamXieMSFT:
>   Configuration: Both

Remove both, so configuration will have to be passed. This will reduce the amount of duplicates in the matrix strategy parameter below.

---
Refers to: build/yaml/jobs/binaries.yaml:10 in fb2983e. [](commit_id = fb2983e2c382144305d342d428da8e6b7b385af0, deletion_comment = False)

build/yaml/jobs/binaries.yaml
WilliamXieMSFT: Regarding the Linux symbols, I'll hold off as Aerobee doesn't do this either at the moment.

---
In reply to: [333242781](https://github.com/microsoft/CLRInstrumentationEngine/pull/134#discussion_r333242781) [](ancestors = 333242781,333216207,333184937,333181904)

Issue #143, opened 2019-10-03
Migrate CLRIE to VS 2019


WilliamXieMSFT:
CLRIE currently depends on VS 2017 toolsets and builds. VS 2019 was released on April 2nd, 2019 and we are internally using it for daily development. 

Unless there's strong pushback to continue developing using VS 2017, we will migrate this project to build with the latest tools provided by VS 2019.

PR #144, opened 2019-10-03
Migrate CLRIE to build with VS2019

WilliamXieMSFT:
This is to address issue #143

### DIFF
--- a/build/Unmanaged.props
+++ b/build/Unmanaged.props
@@ -48,7 +48,7 @@
     <UseOfAtl>Static</UseOfAtl>
     <UseOfMfc>false</UseOfMfc>
     <CharacterSet>Unicode</CharacterSet>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <WindowsTargetPlatformVersion>10.0.14393.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
 
--- a/build/Unmanaged.props
+++ b/build/Unmanaged.props
@@ -49,7 +49,7 @@
     <UseOfMfc>false</UseOfMfc>
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v142</PlatformToolset>
-    <WindowsTargetPlatformVersion>10.0.14393.0</WindowsTargetPlatformVersion>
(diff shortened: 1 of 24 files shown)


build/Unmanaged.props
```diff
@@ -48,8 +48,8 @@
     <UseOfAtl>Static</UseOfAtl>
     <UseOfMfc>false</UseOfMfc>
     <CharacterSet>Unicode</CharacterSet>
-    <PlatformToolset>v141</PlatformToolset>
-    <WindowsTargetPlatformVersion>10.0.14393.0</WindowsTargetPlatformVersion>
+    <PlatformToolset>v142</PlatformToolset>
+    <WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>
```
wiktork: >18362 [](start = 39, length = 5)

We're using an older version for aerobee (10.0.17763.0) iirc. It would be nice to synch these up.

src/TestHostExtension/stdafx.h
```diff
@@ -8,6 +8,9 @@
 
 #pragma once
 
+#define _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING
+#include <experimental/filesystem>
```
wiktork: >experimental [](start = 10, length = 12)

Was filesystem moved to non-experimental? Isn't that the one we would want to be using?

build/Unmanaged.props
WilliamXieMSFT: Sure thing!

WilliamXieMSFT: Discussed offline, we should actually upgrade Aerobee

src/TestHostExtension/stdafx.h
WilliamXieMSFT: The <filesystem> in the 141 toolset was actually the experimental one. I'll create an issue

WilliamXieMSFT: https://docs.microsoft.com/en-us/cpp/standard-library/filesystem?view=vs-2019

WilliamXieMSFT: #145

PR #146, opened 2019-10-03
Create ProfilerManagerWrapper for each Instrumentation Method

WilliamXieMSFT:
Current Changes
* This initial implementation appends the Method ClassId to the logging statement.
* Nit fixes around patchXdt, LocalWindowsTesting scripts
* Nit fix around printing PID in TraceError

Future Change
* Allow InstrumentationMethods to declare their logging levels independent of CLRIE


WilliamXieMSFT:
I'll update the CHANGELOG as well

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -859,7 +862,7 @@ HRESULT MicrosoftInstrumentationEngine::CProfilerManager::SetEventMask2(_In_ DWO
     return S_OK;
 }
 
-MicrosoftInstrumentationEngine::ClrVersion MicrosoftInstrumentationEngine::CProfilerManager::GetAttachedClrVersion()
```
WilliamXieMSFT: >MicrosoftInstrumentationEngine [](start = 43, length = 30)

I can revert this... or update the entire file to remove this namespace since it makes this file harder to read.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    HRESULT hr = S_OK;
+    BOOL fRegistered;
+    IfFailRet(m_pProfilerManager->IsInstrumentationMethodRegistered(clsid, &fRegistered));
+    *pfRegistered = fRegistered;
+    return S_OK;
+}
+
+// IProfilerManagerLogging Methods
+/*
+*/
```
WilliamXieMSFT: Add Comments to these

src/InstrumentationEngine/ProfilerManagerWrapper.h
```diff
@@ -0,0 +1,129 @@
...
+        STDMETHOD(LogDumpMessage)(_In_ const WCHAR* wszMessage);
+
+        // Called to cause logging to be written to the debug output port (via DebugOutputString) as well
+        // as to the host.
+        STDMETHOD(EnableDiagnosticLogToDebugPort)(_In_ BOOL enable);
+
+        // Allows instrumentation methods and hosts to ask for the current logging flags
+        STDMETHOD(GetLoggingFlags)(_Out_ LoggingFlags* pLoggingFlags);
+
+        // Allows instrumentation methods and hosts to modify the current logging flags
```
WilliamXieMSFT: These comments were copy-pasted from CProfilerManager, I can clean them up

src/Scripts/WindowsLocalTest.ps1
```diff
@@ -187,4 +192,8 @@ param(
     }
 }
 
-Start-Process $ApplicationPath
\ No newline at end of file
+Start-Process $ApplicationPath
+
+# This reverts the state of the caller to not be profiled
+$env:COR_ENABLE_PROFILING = 0
+$env:CORECLR_ENABLE_PROFILING = 0
```
WilliamXieMSFT: I'm considering resetting every variable

build/Scripts/PrivateXdt/GeneratePrivateXdt.ps1
```diff
@@ -31,6 +31,7 @@ param(
 )
 
 Import-Module "$PSScriptRoot\XdtHelper.psm1" -Force
+Add-Type -AssemblyName 'System.Xml.Linq'
```
jander-msft: nit: Could you make these changes in a separate PR so as not to conflate the two sets of changes?

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -506,7 +507,9 @@ HRESULT MicrosoftInstrumentationEngine::CProfilerManager::AddInstrumentationMeth
             // InstrumentationMethod. Since we are using global lock - InstrumentationMethod can only set flags associated
             // with it in this thread.
             CInitializeInstrumentationMethodHolder initHolder(this);
-            hr = pInstrumentationMethod->Initialize(this, m_bValidateCodeSignature);
+            CComPtr<CProfilerManagerWrapper> pProfilerManagerWrapper;
+            pProfilerManagerWrapper.Attach(new (nothrow) CProfilerManagerWrapper(pInstrumentationMethod->GetClassId(), this));
```
jander-msft: Since you have "nothrow", you should check that pProfilerManagerWrapper is not null: IfNullRet(pProfilerManagerWrapper);

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -506,7 +507,9 @@ HRESULT MicrosoftInstrumentationEngine::CProfilerManager::AddInstrumentationMeth
...
+            hr = pInstrumentationMethod->Initialize(pProfilerManagerWrapper.Detach(), m_bValidateCodeSignature);
```
jander-msft: I think you should just pass in pProfilerManagerWrapper without detaching.

IIRC, constructing the CProfilerManagerWrapper has the reference count set to 1. CComPtr::Attach/Detach don't change the reference count. The instrumentation method is going to AddRef and Release as needed, but nothing else is holding onto the CProfilerManagerWrapper, so it gets leaked because the initial count never gets released.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::RemoveRawProfilerHook()
+{
+    return m_pProfilerManager->RemoveRawProfilerHook();
+}
+
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::GetCorProfilerInfo(_Outptr_ IUnknown **ppCorProfiler)
+{
+    HRESULT hr = S_OK;
+    CComPtr<IUnknown> pCorProfiler;
```
jander-msft: A lot of these methods don't really do anything... why not just forward the call directly to the underlying m_pProfilerManager and not worry about Attach/Detach, local variables, etc?

src/InstrumentationEngine/ProfilerManagerWrapper.h
jander-msft: I imagine most of these are just declaring what's on the the IProfilerManager* interfaces. If so, you shouldn't need to have these comments here (they'll just get out of sync).

src/InstrumentationEngine/dllmain.cpp
```diff
@@ -3,7 +3,7 @@
 
 #include "stdafx.h"
 
-#include "ProfilerManager.h"
+#include "ProfilerManagerWrapper.h"
```
jander-msft: Why does dllmain.cpp need to include this?

src/Scripts/WindowsLocalTest.ps1
```diff
@@ -138,6 +138,11 @@ param(
         value  = 1
         enable = $DisableSignatureValidation
     }
+    @{
```
jander-msft: Please separate into another PR

src/InstrumentationEngine/ProfilerManagerWrapper.h
```diff
@@ -0,0 +1,129 @@
...
+        public IProfilerManager2,
+        public IProfilerManager3,
+        public IProfilerManager4,
+        public IProfilerManager5,
+        public IProfilerManagerLogging,
+        public CModuleRefCount
+    {
+    public:
+        CProfilerManagerWrapper(GUID instrumentationMethodGuid, CProfilerManager* pProfilerManager);
+        ~CProfilerManagerWrapper();
```
jander-msft: The dtor doesn't do anything, so please remove.

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -241,11 +238,8 @@ namespace MicrosoftInstrumentationEngine
             _In_  LPVOID lpParameter
             );
 
-
         HRESULT LoadInstrumentationMethods(_In_ BSTR bstrConfigPath);
 
-        HRESULT ApplyInstrumentationMethodInstrumentation(_In_ CMethodInfo* pMethodInfo);
```
jander-msft: I assume this didn't have an implementation.

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -200,12 +200,9 @@ namespace MicrosoftInstrumentationEngine
 
         HRESULT CreateNewMethodInfo(_In_ FunctionID functionId, _Out_ CMethodInfo** ppMethodInfo);
 
-        CLogging* GetLogging();
```
jander-msft: I assume this didn't have an implementation.

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: You can go either way as long as it is consistent. I prefer not having the namespace everywhere (your latter suggestion).

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    wcscpy_s(wszPrefix, m_wszInstrumentationMethodGuid.c_str());
+    wcscat_s(wszPrefix, wszError);
+    return m_pProfilerManager->LogError(wszPrefix);
+}
+
+/*
+*/
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::LogDumpMessage(_In_ const WCHAR* wszMessage)
+{
+    WCHAR wszPrefix[MAX_PATH];
```
jander-msft: You'll have to check if the tests are okay with this change. The logging system and tests treat dump logging slightly differently than error and message logging.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    return S_OK;
+}
+
+// IProfilerManagerLogging Methods
+/*
+*/
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
+{
+    WCHAR wszPrefix[MAX_PATH];
+    wcscpy_s(wszPrefix, m_wszInstrumentationMethodGuid.c_str());
```
jander-msft: I think it would be better if these were more formatted, something like "[IM:00000000-0000-0000-0000-000000000000]..." so that it's easier to know what the GUID is for and easier to parse out this information. I would like to do this in general for most of the information, such as changing the "LogError" and "LogMessage" to a similar format, but that would be beyond the scope of this change.

build/Scripts/PrivateXdt/GeneratePrivateXdt.ps1
WilliamXieMSFT: Sure, I'll separate them out as separate PRs

src/InstrumentationEngine/ProfilerManager.h
WilliamXieMSFT: That's correct, there were two methods that did not have implementation so I've removed them.

src/InstrumentationEngine/ProfilerManager.h
WilliamXieMSFT: Correct.

src/InstrumentationEngine/dllmain.cpp
WilliamXieMSFT: Artifact of past change, reverting

---
In reply to: [331295942](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331295942) [](ancestors = 331295942)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: I was not thinking clearly

---
In reply to: [331295260](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331295260) [](ancestors = 331295260)

src/InstrumentationEngine/ProfilerManager.cpp
delmyers: And return E_OUTOFMEMORY

---
In reply to: [331293268](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331293268) [](ancestors = 331293268)

src/InstrumentationEngine/ProfilerManager.cpp
delmyers: If you are going to make this change, can you update the rest of the file as well?

---
In reply to: [331296901](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331296901) [](ancestors = 331296901)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    wcscat_s(wszPrefix, wszMessage);
+    return m_pProfilerManager->LogMessage(wszPrefix);
+}
+
+/*
+*/
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::LogError(_In_ const WCHAR* wszError)
+{
+    WCHAR wszPrefix[MAX_PATH];
+    wcscpy_s(wszPrefix, m_wszInstrumentationMethodGuid.c_str());
```
delmyers: >wszPrefix [](start = 13, length = 9)

There is a potential buffer overrun here if, somehow the guid is corrupted and the underlying size of the string in memory gets calculated as larger than MAX_PATH. Please replace string manipulation functions with ones that take a destination size as a parameter.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: I agree, a better design would be WriteLog(LogLevel, IMGuid, MessageStr) to be more explicit about these formats. It would require a new interface.

---
In reply to: [331297825](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331297825) [](ancestors = 331297825)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Good catch

---
In reply to: [331294863](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331294863) [](ancestors = 331294863)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Thanks!

---
In reply to: [331673239](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331673239) [](ancestors = 331673239,331293268)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: You're correct, the Dumps are used to test the result of the output. However, our tests currently do not track InstrumentationMethod logging. I will add new tests to incorporate those changes.

---
In reply to: [331297278](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331297278) [](ancestors = 331297278)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: The current implementation with Detach fails RuntimeExceptionCallbacksTest with error -1073741819.

Removing the detach causes the RuntimeExceptionCallbacks_Debug_x64.exe to crash due to access violation during release. 

I'm going to reactivate this comment and investigate further.

---
In reply to: [331680571](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331680571) [](ancestors = 331680571,331294863)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    wcscat_s(wszPrefix, wszError);
```
delmyers: >wszPrefix [](start = 13, length = 9)

same as above, if the size(wszError + Guid) > MAX_PATH, then you will have a buffer overrun, which is a security issue.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    wcscpy_s(wszPrefix, m_wszInstrumentationMethodGuid.c_str());
+    wcscat_s(wszPrefix, wszMessage);
+    return m_pProfilerManager->LogMessage(wszPrefix);
+}
+
+/*
+*/
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::LogError(_In_ const WCHAR* wszError)
+{
+    WCHAR wszPrefix[MAX_PATH];
```
delmyers: >wszPrefix[MAX_PATH] [](start = 10, length = 19)

Why MAX_PATH?

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
delmyers: Never mind, Jackson indicated to me that this does use the size, but it is inferred from the size of the parameter that is passed in. Internally, this calls wcscat_s(wszPrefix, array_size(wszPrefix), wszError).

---
In reply to: [331698666](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331698666) [](ancestors = 331698666)

src/InstrumentationEngine/ProfilerManagerWrapper.h
```diff
@@ -0,0 +1,129 @@
...
+
+namespace MicrosoftInstrumentationEngine
+{
+
+    class CProfilerManagerWrapper :
+        public IProfilerManager,
+        public IProfilerManager2,
+        public IProfilerManager3,
+        public IProfilerManager4,
+        public IProfilerManager5,
```
wiktork: >IProfilerManager5 [](start = 15, length = 17)

so if we add IProfilerManager6, this class won't be able to QI for it by default. Might be worthwhile to add comments into the idl to make sure this is updated as well.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: It's the convention we use everywhere.

---
In reply to: [331698879](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331698879) [](ancestors = 331698879)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// IProfilerManagerLogging Methods
+/*
+*/
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
+{
+    WCHAR wszPrefix[MAX_PATH];
+    wcscpy_s(wszPrefix, m_wszInstrumentationMethodGuid.c_str());
+    wcscat_s(wszPrefix, wszMessage);
```
wiktork: >wcscat_s [](start = 4, length = 8)

Did i miss something here? What if the message is bigger than MAX_PATH?

jander-msft: FYI, if you look deeper into the CLogging/CLoggerService implementation, the buffer size is 4096.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+    BOOL fRegistered;
+    IfFailRet(m_pProfilerManager->IsInstrumentationMethodRegistered(clsid, &fRegistered));
+    *pfRegistered = fRegistered;
+    return S_OK;
+}
+
+// IProfilerManagerLogging Methods
+/*
+*/
+HRESULT MicrosoftInstrumentationEngine::CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
```
WilliamXieMSFT: >LogMessage [](start = 65, length = 10)

Pass format string and guid instead.

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Turns out the test instrumentation method doesn't addref the profilermanager so it gets released causing the issue.

---
In reply to: [331696295](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331696295) [](ancestors = 331696295,331680571,331294863)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    WCHAR wszPrefix[MAX_PATH];
+    wcscpy_s(wszPrefix, m_wszInstrumentationMethodGuid.c_str());
+    wcscat_s(wszPrefix, wszMessage);
+    return m_pProfilerManager->LogMessage(wszPrefix);
```
WilliamXieMSFT: >wszPrefix [](start = 42, length = 9)

This is only logging the prefix

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: DIscussed this offline, for internal interface, it's fairly trivial.

The harder task would be if we were to support this interface publicly and how to handle dll boundaries, unnecessary memory allocations, etc. I'll create a task/issue tracking the latter discussion.

---
In reply to: [331712326](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331712326) [](ancestors = 331712326)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: *almost everywhere. I'm fixing this to be consistent with the rest of the logging infrastructure (4096 characters)

---
In reply to: [331701234](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331701234) [](ancestors = 331701234,331698879)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: Worked around this by making NaglerInstrumentationMethod depend on the global logging instance rather than the profiler manager wrapper. 

---
In reply to: [331297278](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331297278) [](ancestors = 331297278)

src/InstrumentationEngine/ProfilerManagerWrapper.h
WilliamXieMSFT: I've added the comment in CProfilerManager since we only want to update CProfilerManagerWrapper if the Manager is updated.

---
In reply to: [331700175](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r331700175) [](ancestors = 331700175)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+// IProfilerManagerLogging Methods
+/*
+ * Inject Instrumentation Method ClassId
+ */
+HRESULT CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
+{
+    HRESULT hr = S_OK;
+    WCHAR wszBuffer[LogEntryMaxSize] = _T("[IM:%s]");
+    IfFailRetErrno(wcscat_s(wszBuffer, wszMessage));
```
jander-msft: This is assuming that there are no residual format specifiers or that the caller was intentionally try to log something that looks like a format specified. Do we need to escape those, or do some kind of replacement before/after formatting the string?

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// IProfilerManagerLogging Methods
+/*
+ * Inject Instrumentation Method ClassId
+ */
+HRESULT CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
+{
+    HRESULT hr = S_OK;
+    WCHAR wszBuffer[LogEntryMaxSize] = _T("[IM:%s]");
```
jander-msft: Can you initialize with wszLogFormattablePrefix? Or initially leave it empty and then append the const string?

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -26,9 +26,9 @@ HRESULT CInstrumentationMethod::Initialize(_In_ IProfilerManager* pProfilerManag
         pProfilerManager->GetCorProfilerInfo((IUnknown**)&pCorProfilerInfo);
         pCorProfilerInfo->SetEventMask(COR_PRF_MONITOR_EXCEPTIONS | COR_PRF_ENABLE_STACK_SNAPSHOT);
 
-        CComPtr<IProfilerManagerLogging> spLogger;
-        pProfilerManager->GetLoggingInstance(&spLogger);
-        spLogger->LogDumpMessage(_T("<ExceptionTrace>\r\n"));
+        CComQIPtr<IProfilerManager4> pProfilerManager4 = pProfilerManager;
```
jander-msft: Why was this changed?

delmyers:
>                      public IProfilerManager5,

If you need two classes to implement the same interfaces, you can create a new class that both classes extend.

```
class IProfilerManagerContract :
    public IProfilerManager,
    public IProfilerManager2,
    public IProfilerManager3,
    public IProfilerManager4,
    public IProfilerManager5
{
/// ...
}

```

---
Refers to: src/InstrumentationEngine/ProfilerManager.h:35 in 53d1f69. [](commit_id = 53d1f69b961dbde3c064d91c86887432c50e5f6b, deletion_comment = False)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
delmyers: >_T("[IM:%s]"); [](start = 39, length = 14)

wszLogFormattablePrefix? + all occurrences

delmyers: He can't directly initialize with wszLogFormattablePrefix because you can't initialize an array with a pointer. He could initialize with the const, if he wanted to use the tstring class, but that would incur allocation cost. The easiest thing to do to enable reuse is to use a macro.

---
In reply to: [333780025](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r333780025) [](ancestors = 333780025)

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
WilliamXieMSFT: The GetLoggingInstance appends the IM prefix, however the GetGlobalLoggingInstance does not. This was the quickest way to unblock the Tests since they assume the logging output generates correct XML (the IM prefix breaks this).

I'll add IM-specific logging tests in the next iteration

---
In reply to: [333780324](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r333780324) [](ancestors = 333780324)

WilliamXieMSFT:
>                      public IProfilerManager5,

Great suggestion!

---
In reply to: [541166839](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#issuecomment-541166839) [](ancestors = 541166839)

---
Refers to: src/InstrumentationEngine/ProfilerManager.h:35 in 53d1f69. [](commit_id = 53d1f69b961dbde3c064d91c86887432c50e5f6b, deletion_comment = False)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: I'll remove the wszLogFormattablePrefix and define a macro as Del suggested.

---
In reply to: [334116304](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r334116304) [](ancestors = 334116304,333780025)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: Good point, This interface does not assume the string has format specifiers, I'll have to escape `%` in the original string.

---
In reply to: [333779838](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r333779838) [](ancestors = 333779838)

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
WilliamXieMSFT: Turns out there was a bug in the XML comparison (it compared the xml to itself so always succeeds). 

---
In reply to: [336672890](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r336672890) [](ancestors = 336672890,333780324)

tests/InstrEngineTests/InstrEngineTests/TestConstants.cs
```diff
@@ -11,6 +11,6 @@ namespace InstrEngineTests
 {
     internal class TestConstants
     {
-        public const int TestTimeout = 60000; //Int32.MaxValue;
+        public const int TestTimeout = Int32.MaxValue; //60000
```
WilliamXieMSFT: >60000 [](start = 57, length = 5)

reverting

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
```diff
@@ -132,25 +132,32 @@ public static void DiffResultToBaseline(string output, string baseline)
...
 
             using (StreamReader baselineStream = new StreamReader(baselinePath))
+            using (StreamReader outputStream = new StreamReader(outputPath))
             {
-                using (StreamReader outputStream = new StreamReader(outputPath))
+                string tmpStr;
+                StringBuilder strBuilder = new StringBuilder();
+                while ((tmpStr = baselineStream.ReadLine()) != null)
                 {
-                    baselineStr = baselineStream.ReadToEnd();
```
WilliamXieMSFT: >ReadToEnd [](start = 49, length = 9)

This keeps newline characters whereas ReadLine() does not.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,237 @@
...
+HRESULT CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
+{
+    HRESULT hr = S_OK;
+    WCHAR wszBuffer[LogEntryMaxSize] = { 0 };
+    tstring tsMessage = wszMessage;
+    tstring tsEscapedMessage;
+    for (tstring::iterator it = tsMessage.begin(); it < tsMessage.end(); it++)
+    {
+        // Escape any existing format strings to avoid conflicts with the FORMATTABLE_PREFIX
+        if (*it == L'%')
```
jander-msft: Maybe add a better comment as to what is happening here. If I understand correctly, you are duplicating the % characters since %% is how you escape a single % character in a format string that you don't want treated as a format specifier.

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,237 @@
...
+/*
+ * Inject Instrumentation Method ClassId
+ */
+HRESULT CProfilerManagerWrapper::LogMessage(_In_ const WCHAR* wszMessage)
+{
+    HRESULT hr = S_OK;
+    WCHAR wszBuffer[LogEntryMaxSize] = { 0 };
+    tstring tsMessage = wszMessage;
+    tstring tsEscapedMessage;
+    for (tstring::iterator it = tsMessage.begin(); it < tsMessage.end(); it++)
```
jander-msft: Please refactor all three of these copies into another method.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
jander-msft: Could you revert this?

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1142,40 +1158,30 @@ HRESULT CInstrumentationMethod::GetType(IModuleInfo* pModuleInfo, const CLocalTy
...
     return S_OK;
 }
 
-
 HRESULT CInstrumentationMethod::ExceptionCatcherEnter(
     _In_ IMethodInfo* pMethodInfo,
     _In_ UINT_PTR   objectId
     )
 {
-    CComPtr<IProfilerManagerLogging> spLogger;
```
jander-msft: Consider reverting these type of changes. I think it's more important to have it test the IProfilerManager::GetLoggingInstance method and make sure that your new logging changes still work over time.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -20,15 +20,24 @@ HRESULT CInstrumentationMethod::Initialize(_In_ IProfilerManager* pProfilerManag
...
     {
         CComPtr<ICorProfilerInfo> pCorProfilerInfo;
         pProfilerManager->GetCorProfilerInfo((IUnknown**)&pCorProfilerInfo);
         pCorProfilerInfo->SetEventMask(COR_PRF_MONITOR_EXCEPTIONS | COR_PRF_ENABLE_STACK_SNAPSHOT);
 
-        CComPtr<IProfilerManagerLogging> spLogger;
-        pProfilerManager->GetLoggingInstance(&spLogger);
-        spLogger->LogDumpMessage(_T("<ExceptionTrace>\r\n"));
+        CComQIPtr<IProfilerManager4> pProfilerManager4 = pProfilerManager;
+        pProfilerManager4->GetGlobalLoggingInstance(&m_pProfilerManagerLogger);
```
jander-msft: nit: If you end up keeping this change, please rename to something like m_pGlobalLogger

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    return m_pProfilerManager->GetLoggingFlags(pLoggingFlags);
+}
+
+HRESULT CProfilerManagerWrapper::SetLoggingFlags(_In_ LoggingFlags loggingFlags)
+{
+    return m_pProfilerManager->SetLoggingFlags(loggingFlags);
+}
+
+#undef PREPEND_FORMATTABLE_PREFIX
```
jander-msft: This wasn't defined, so this is doing nothing. Did you mean "#undef FORMATTABLE_PREFIX"?

src/InstrumentationEngine/ProfilerManagerWrapper.h
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "ProfilerManager.h"
+
+namespace MicrosoftInstrumentationEngine
+{
+    class CProfilerManagerWrapper :
```
jander-msft: I know that we've been calling it CProfilerManagerWrapper, but maybe we can come up with a better name, e.g. CProfilerManagerForInstrumentationMethod, that better describes for what it is used.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
WilliamXieMSFT: Sure

---
In reply to: [341218858](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r341218858) [](ancestors = 341218858)

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
WilliamXieMSFT: Reverting my change back to the original code with spLogger instead.

---
In reply to: [341220900](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r341220900) [](ancestors = 341220900)

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
WilliamXieMSFT: The problem is that now that method prepends the instrumentation method guid and breaks the test verification which parses XML. I am QI for ProfilerManager4 and invoke GetGlobalLogger to work around this.

---
In reply to: [341220369](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r341220369) [](ancestors = 341220369)

src/InstrumentationEngine/ProfilerManagerWrapper.cpp
WilliamXieMSFT: Comment is in refactored method "EscapeFormatSpecifiers"

---
In reply to: [341216601](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r341216601) [](ancestors = 341216601)

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
WilliamXieMSFT: I'll try to update the tests in the next PR to ignore these fields when parsing XML

---
In reply to: [341282058](https://github.com/microsoft/CLRInstrumentationEngine/pull/146#discussion_r341282058) [](ancestors = 341282058,341220369)

PR #156, opened 2019-11-01
Update YAML builds to run on public project

WilliamXieMSFT:
These changes allow exposing the PR and CI yaml build results to the public

PR Build:
https://dev.azure.com/ms/CLRInstrumentationEngine/_build?definitionId=230&_a=summary

CI Build:
https://dev.azure.com/ms/CLRInstrumentationEngine/_build?definitionId=275&_a=summary

Other changes:
* Update build triggers to point to public builds
* Update build badges on docs
* Update docs to explain this process


build/yaml/jobs/linux/binaries.yaml
```diff
@@ -35,4 +32,5 @@ jobs:
       PathToPublish: '$(Build.SourcesDirectory)/out/Linux/bin/$(Platform).$(Configuration)'
       ArtifactName: 'binaries-linux-$(Distro)-$(Configuration)'
 
-  - template: ../../steps/microbuild/cleanup.yaml
\ No newline at end of file
+  - ${{ if notIn(parameters.SignType, '', 'None') }}:
```
delmyers: >notIn [](start = 11, length = 5)

consider doing the inverse of this... check to see if the sign type is "Test" or "Real"

build/yaml/jobs/linux/packages.yaml
```diff
@@ -44,6 +41,7 @@ jobs:
       PathToPublish: '$(Build.ArtifactStagingDirectory)\bin\$(Configuration)'
       ArtifactName: 'packages-all'
 
-  - template: ../../steps/microbuild/cleanup.yaml
+  - ${{ if notIn(parameters.SignType, '', 'None') }}:
```
delmyers: >notIn [](start = 11, length = 5)

similar to binaries.yaml

build/yaml/jobs/linux/binaries.yaml
delmyers: Also, is this just a way to check to see if this an official build? Why not just add a build variable to Azure Devops instead? It seams like an imperfect heuristic that could just be solved by a variable.

---
In reply to: [342160850](https://github.com/microsoft/CLRInstrumentationEngine/pull/156#discussion_r342160850) [](ancestors = 342160850)

WilliamXieMSFT: Since MicroBuild tasks are internal, they will fail with "task not found" on public Azure DevOps projects. The only MicroBuild task we depend on is for Signing, so we can expose our PR and CI builds. MicroBuild cleanup just sends telemetry for builds that use MicroBuild tasks.

I'll refactor to use a separate variable `IsInternalProject` which would allow us to skip any task that is considered internal, not just MicroBuild.

---
In reply to: [342162437](https://github.com/microsoft/CLRInstrumentationEngine/pull/156#discussion_r342162437) [](ancestors = 342162437,342160850)

build/yaml/pipelines/signed.yaml
```diff
@@ -20,11 +20,13 @@ stages:
       IndexSources: true
       SignType: $(SignType)
       Configuration: Release
+      UseInternalTasks: true
```
delmyers: >UseInternalTasks [](start = 6, length = 16)

Nit: consider something like "IsOfficialBuild" (though I think we have a variable for that already), or "IsMicrobuildInternal".

PR #165, opened 2019-11-11
Add support for InstrumentationMethod LogLevels

WilliamXieMSFT:
## Overview 
This change introduces the ability for InstrumentationMethods to declare their own LogLevels via setting "MicrosoftInstrumentationEngine_LogLevel_GUID" (where GUID is the 36 character ClassID of the instrumentation method including dashes, ie. "00000000-0000-0000-0000-000000000000")

## Background
In the previous PR, I created the class `CProfilerManagerForInstrumentationMethod` (`PMforIM` for short), which is a wrapper of ProfilerManager. Every Method is given a new instance of `PMforIM` that keeps track of the InstrumentationMethod's ClassID Guid. If an InstrumentationMethod calls `LogMessage()|LogDumpMessage()|LogError()` through the `IProfilerManagerLogging` interface returned by `IProfilerManager::GetLoggingInstance()`, `PMforIM` will prepend the ClassID Guid to that message.

`IProfilerManager::GetGlobalLoggingInstance()` will return the global static logging instance that is not tied to any specific InstrumentationMethod and therefore will not have an InstrumentationMethod Guid to prepend to messages.

### Loggers
Each instance of `PMforIM` corresponds to an InstrumentationMethod and keeps track of the LoggingFlag from the "MicrosoftInstrumentationEngine_LogLevel_GUID" environment variable corresponding to that Method (if exists). `PMforIM` will use this LogLevel to filter `Log*()` calls.

After creating the `PMforIM`, the ProfilerManager updates `CLogging::SetInstruMethodLoggingFlags` so that the proper Logging Sinks are initialized. See next section for details.

### Logging Sinks 
CLoggerService relies on `defaultFlags` (set by "MicrosoftInstrumentationEngine_LogLevel" environment variable) and `effectiveFlags` (the effective set of flags supported by all logger sinks in response to the `defaultFlags`) in order to handle log messages. The `effectiveFlags` is what filters the GlobalInstanceLogger.

In order to keep the GlobalInstanceLogger and InstrumentationMethodInstanceLoggers from contaminating each other's log level states, I've created an `allInstruMethodFlags` variable that is used alongside the `defaultFlags` to initialize the sinks but does not update the `effectiveFlags`.

This is achieved by updating `CLoggerService::RecalculateLoggingFlags()` to reset each `ILoggerSink`, once with the `defaultFlags` and once with `allInstruMethodFlags` and only setting the `effectiveFlags` to the resultant from the reset done with the `defaultFlags`.


WilliamXieMSFT:
I will add documentation of the details around the Logging Infrastructure in future iterations once everyone is satisfied with the implementation.

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
```diff
@@ -66,7 +77,7 @@ public static void LaunchAppUnderProfiler(string testApp, string testScript, str
             psi.EnvironmentVariables.Add("COR_ENABLE_PROFILING", "1");
             psi.EnvironmentVariables.Add("COR_PROFILER", ProfilerGuid.ToString("B"));
             psi.EnvironmentVariables.Add("COR_PROFILER_PATH", Path.Combine(PathUtils.GetAssetsPath(), string.Format("MicrosoftInstrumentationEngine_{0}.dll", bitnessSuffix)));
-            psi.EnvironmentVariables.Add("MicrosoftInstrumentationEngine_FileLog", "Dumps");
+            psi.EnvironmentVariables.Add("MicrosoftInstrumentationEngine_LogLevel", "Dumps");
```
WilliamXieMSFT: >MicrosoftInstrumentationEngine_LogLevel [](start = 42, length = 39)

Add another test that disables LogLevel altogether

tests/InstrEngineTests/InstrEngineTests/TestConstants.cs
```diff
@@ -11,6 +11,6 @@ namespace InstrEngineTests
 {
     internal class TestConstants
     {
-        public const int TestTimeout = 60000; //Int32.MaxValue;
+        public const int TestTimeout = Int32.MaxValue; // 60000;
```
WilliamXieMSFT: >Int32.MaxValue; // 60000; [](start = 39, length = 25)

reverting

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -514,8 +514,21 @@ HRESULT CProfilerManager::AddInstrumentationMethod(
                 return E_OUTOFMEMORY;
             }
 
+            LoggingFlags flag;
+            pProfilerManagerWrapper->GetInstruMethodLoggingFlags(&flag);
```
delmyers: Does this return an hr?

src/InstrumentationEngine.Lib/Logging.cpp
```diff
@@ -117,6 +137,14 @@ HRESULT CLogging::SetLoggingFlags(_In_ LoggingFlags loggingFlags)
     return s_loggerService.Get()->SetLoggingFlags(loggingFlags);
 }
 
+// static
+HRESULT CLogging::SetInstruMethodLoggingFlags(_In_ LoggingFlags loggingFlags)
```
delmyers: >InstruMethod [](start = 21, length = 12)

Is "InstruMethod" used elsewhere in the code? Seems a little awkward.

src/InstrumentationEngine/ProfilerManager.cpp
delmyers: It does return an hr. Seems like you shouldn't try to accumulate logging flags if there is a failure on this call. You could potentially be setting things to garbage.

I don't know if you can assert here, though, since logging may not be initialized yet, and I think asserts log by default.

---
In reply to: [345310744](https://github.com/microsoft/CLRInstrumentationEngine/pull/165#discussion_r345310744) [](ancestors = 345310744)

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -304,18 +333,23 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
...
         // Reset the sink; get its desired logging level
         LoggingFlags sinkFlags;
+        LoggingFlags instruMethodSinkFlags;
         IfFailRetNoLog(pSink->Reset(m_defaultFlags, &sinkFlags));
+        IfFailRetNoLog(pSink->Reset(m_allInstruMethodFlags, &instruMethodSinkFlags));
 
         // Add the sink to each logging level vector
-        if (IsFlagSet(sinkFlags, LoggingFlags_Errors))
+        if (IsFlagSet(sinkFlags, LoggingFlags_Errors) ||
+            IsFlagSet(instruMethodSinkFlags, LoggingFlags_Errors))
```
delmyers: can this be done similar to above:

IsFlagSet(sinkFlags, | instruMethodSinkFlags, LoggingFlags_Errors)

delmyers:
> CInstrumentationMethod :

There doesn't seem to be any tests for accumulating logging flags. Should there be?

---
Refers to: tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.h:9 in 7c5e8e0. [](commit_id = 7c5e8e05b8d80a966e423c2be66eebc8bd81e2b4, deletion_comment = False)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Thanks, I can guard this with an if (SUCCEEDED) check

---
In reply to: [345313778](https://github.com/microsoft/CLRInstrumentationEngine/pull/165#discussion_r345313778) [](ancestors = 345313778,345310744)

WilliamXieMSFT:
> CInstrumentationMethod :

I can look into adding another instrumentation method so we can test combinations of settings

---
In reply to: [552991258](https://github.com/microsoft/CLRInstrumentationEngine/pull/165#issuecomment-552991258) [](ancestors = 552991258)

---
Refers to: tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.h:9 in 7c5e8e0. [](commit_id = 7c5e8e05b8d80a966e423c2be66eebc8bd81e2b4, deletion_comment = False)

src/InstrumentationEngine.Lib/Logging.cpp
WilliamXieMSFT: It's used mostly in tests. I'll expand this.

---
In reply to: [345312339](https://github.com/microsoft/CLRInstrumentationEngine/pull/165#discussion_r345312339) [](ancestors = 345312339)

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
WilliamXieMSFT: This requires adding a regex matching baseline (since I rely on the global logger for XML nodes)

---
In reply to: [344955335](https://github.com/microsoft/CLRInstrumentationEngine/pull/165#discussion_r344955335) [](ancestors = 344955335)

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -73,6 +74,23 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
     return S_OK;
 }
 
+HRESULT CLoggerService::GetInstruMethodLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
```
WilliamXieMSFT: >GetInstruMethodLoggingFlags [](start = 24, length = 27)

It's a bit strange for ProfilerManager to query this each time before setting it again (and we lose track of which method is logging at which level). I'm revamping this API so instead of Get/Set, there will only be one UpdateInstrumentationMethodLoggingFlag that updates a dictionary in CLoggerService

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -73,6 +99,31 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
...
+
+    m_instrumentationMethodFlags = LoggingFlags_None;
+
+    IfFailRetNoLog(UpdateFlags(classId, loggingFlags));
+
+    for (unordered_map<LoggingFlags, set<GUID>>::iterator it = m_loggingFlagsToInstrumentationMethodsMap.begin();
+        it != m_loggingFlagsToInstrumentationMethodsMap.end();
+        it++)
+    {
+        if (!(it->second.empty()))
```
WilliamXieMSFT: >if (!(it->second.empty())) [](start = 8, length = 26)

I'll add a comment here, effectively we're iterating through all of the logging flags and if the set of guids is not empty (ie. at least one instrumentation method requires that logging level), we will update the m_instrumentationMethodFlags.

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -362,3 +415,53 @@ HRESULT CLoggerService::CreateSinks(vector<shared_ptr<ILoggerSink>>& sinks)
...
+
+    IfFailRetNoLog(UpdateFlagsInternal(
+        classId,
+        loggingFlags,
+        LoggingFlags_InstrumentationResults));
+
+    return S_OK;
+}
+
+HRESULT CLoggerService::UpdateFlagsInternal(_In_ GUID classId, _In_ LoggingFlags loggingFlags, _In_ LoggingFlags loggingLevel)
```
WilliamXieMSFT: >UpdateFlagsInternal [](start = 24, length = 19)

Adding a comment, for the specific LogLevel, we check if the InstrumentationMethod ClassId exists and whether it should Exist (ie. the loggingFlags includes the loggingLevel). If it exists and should not exist, we remove the classId from the set. If it doesn't exist and should exist, we add to the set.

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -362,3 +415,53 @@ HRESULT CLoggerService::CreateSinks(vector<shared_ptr<ILoggerSink>>& sinks)
...
+
+    bool exists = pSet->find(classId) != pSet->end();
+    bool shouldExist = (loggingLevel & loggingFlags) != 0;
+    if (exists && !shouldExist)
+    {
+        // Remove
+        size_t result = pSet->erase(classId);
+        if (result < 1)
+        {
+            // Unable to remove
```
WilliamXieMSFT: >// Unable to remove [](start = 12, length = 19)

Handle this

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -362,3 +415,53 @@ HRESULT CLoggerService::CreateSinks(vector<shared_ptr<ILoggerSink>>& sinks)
...
+        }
+    }
+    else if (!exists && shouldExist)
+    {
+        // Add
+        std::pair<set<GUID>::iterator, bool> ret = pSet->emplace(classId);
+        if (!ret.second)
+        {
+            // Unable to insert
```
WilliamXieMSFT: >// Unable to insert [](start = 12, length = 19)

Handle this

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
```diff
@@ -126,66 +186,99 @@ public static string[] SplitXmlDocuments(string content)
...
+                    }
                 }
             }
+            catch (AssertFailedException)
+            {
+                Console.WriteLine($"Baseline FilePath: {baselinePath}");
+                Console.WriteLine();
+                Console.WriteLine($"Output FilePath: {outputPath}");
+                throw;
+            }
```
WilliamXieMSFT: I've found it annoying to figure out which folder in TestResult\ corresponded to the failure. Luckily VSTest outputs stdout so I'm dumping these filepaths on failures

WilliamXieMSFT:
> CInstrumentationMethod :

Looks like I can reuse Justin's LoggingTests in InstrumentationEngine.Lib.Tests project

---
In reply to: [552999147](https://github.com/microsoft/CLRInstrumentationEngine/pull/165#issuecomment-552999147) [](ancestors = 552999147,552991258)

---
Refers to: tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.h:9 in 7c5e8e0. [](commit_id = 7c5e8e05b8d80a966e423c2be66eebc8bd81e2b4, deletion_comment = False)

PR #183, opened 2019-12-13
Fix target folder for CodeSignVerify and add exclusion list

WilliamXieMSFT:
The most important exclusion is embedded Cab files (which are generated when the installer consumes a merge module).  

According to https://docs.microsoft.com/windows/win32/msi/digital-signatures-and-windows-installer:

_With Windows Installer, digital signatures can be used with Windows Installer packages, transforms, patches, merge modules, and external cabinet files._

And https://www.advancedinstaller.com/user-guide/msm-digital-signature.html

_Advanced Installer can digitally sign all of the following files that it creates: EXE, MSI, MSP (patches) and CAB files. The EXE, MSI and MSP files are always signed while the CAB files are signed only if they are not embedded in the MSI._

Considering both the outer MSI and the internal files are signed, we felt excluding the embedded cab from being sign-verified was reasonable.

### DIFF
--- a/build/yaml/steps/microbuild/codesignverify.yaml
+++ b/build/yaml/steps/microbuild/codesignverify.yaml
@@ -15,6 +15,6 @@ steps:
   displayName: Verify Signed Files
   condition: and(succeeded(), in(variables['SignType'], 'Real', 'Test'))
   inputs:
-    TargetFolder: $(Build.BinariesDirectory)
+    TargetFolder: $(Build.StagingDirectory)
     WhiteListPathForCerts: 'build\CodeSign\certificate.ignore'
     WhiteListPathForSigs: 'build\CodeSign\strongname.ignore'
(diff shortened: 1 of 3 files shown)


build/yaml/steps/microbuild/codesignverify.yaml
```diff
@@ -15,6 +15,6 @@ steps:
   displayName: Verify Signed Files
   condition: and(succeeded(), in(variables['SignType'], 'Real', 'Test'))
   inputs:
-    TargetFolder: $(Build.BinariesDirectory)
+    TargetFolder: $(Build.StagingDirectory)
```
wiktork: Long term we should consider building into the Binaries folder. Do our static analysis builds work properly here?

WilliamXieMSFT: Currently they're pointing to the BuildArtifactStagingDirectory as well (note that BuildArtifactStagingDirectory and StagingDirectoy are interchangeable).

Issue #186, opened 2020-01-03
GetILFunctionBody supports NULL for ULONG* pcbMethodSize


WilliamXieMSFT:
https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo-getilfunctionbody-method

Despite the annotation, pcbMethodSize can be NULL when querying from the CLR. However, CLRIE assumes this is not null (via SAL) but does not null-check and `*pcbMethodSize = 0;` in `CCorProfilerInforWrapper::GetILFunctionBody()` encounters an Access Violation (AV) with nullptr.

WilliamXieMSFT:
For Example: https://github.com/microsoft/clr-samples/blob/master/ProfilingAPI/ReJITEnterLeaveHooks/ILRewriter.cpp#L208 passes in NULL to GetILFunctionBody which works for .NET Framework and .NET Core but fails with CLRIE

WilliamXieMSFT:
https://github.com/dotnet/runtime/blob/5ef367a888fa8a25ce157cf340bd271efd36c72f/src/coreclr/src/vm/proftoeeinterfaceimpl.cpp#L4412

Despite its own SAL annotation, the CLR conducts a NULL check.

delmyers:
What is the bug here? It seems that at most, we should be doing an IfNullRet on the pcbMethodSize, and not depend on a side-effect of a bug in the clr profiler apis. This seems like a bug in the CLR. It is either:

1. A bug in the SAL or
2. A bug in the implementation and 
    1. A bug in the sample provided by the CLR

WilliamXieMSFT:
I agree, it looks like a bug in the CLR that allows behavior despite its own contract via the annotation. I'll open an issue on the CLR and close this issue.

As an aside, I'm worried that profiler authors will depend on these accidental behaviors and encounter issues when connecting to the CLRIE RawProfilerHook.

delmyers:
I would be OK with putting an IfNullRet in our code, so that we are explicit about the error. That way we return a null argument error, rather than throwing an exception.

WilliamXieMSFT:
https://github.com/dotnet/runtime/issues/1281

I'll submit a PR to address this for CLRIE.

WilliamXieMSFT:
[...]

PR #187, opened 2020-01-03
Add null check for SAL _Out_ pointer

WilliamXieMSFT:
Address #186 

To summarize, the SAL annotations produced by MIDL are not enforced by the CLR. This is fine for Instrumentation Methods as static/compile-time failures are easily fixable, and CLRIE will require a more stringent API. However, for the RawProfilerHook, we'll need to mirror the behavior provided by CLR so that existing profilers do not AV when migrating to CLRIE.

### DIFF
--- a/src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
+++ b/src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
@@ -346,6 +346,10 @@ HRESULT MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::GetILFunctionBo
     )
 {
     HRESULT hr = S_OK;
+
+    IfNullRetPointer(ppMethodHeader);
+    IfNullRetPointer(pcbMethodSize);
+
     *ppMethodHeader = nullptr;
     *pcbMethodSize = 0;
 
--- a/src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
+++ b/src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
@@ -349,20 +349,15 @@ HRESULT MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::GetILFunctionBo
 
     // IDL [out] parameter annotation is optional unless the function only has one.
     // The corresponding [in] parameter for the size to fill the [out] parameter must be 0.
-    BYTE* pIgnoreMethodHeader = nullptr;
-    if (ppMethodHeader == NULL)
+    if (ppMethodHeader != nullptr)
     {
-        ppMethodHeader = &pIgnoreMethodHeader;
+        *ppMethodHeader = nullptr;
     }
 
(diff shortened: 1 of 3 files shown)


WilliamXieMSFT:
This fix should not throw on NULL due to the CLR's non-enforcement of the IDL to SAL annotation. 

>All [out] parameters on the API are optional, unless a function has only one [out] parameter. A profiler simply passes NULL for any [out] parameters it is not interested in. The profiler must also pass consistent values for any associated [in] parameters—e.g., if the NULL [out] parameter is a buffer to be filled with data, the [in] parameter specifying its size must be 0.

I'll reuse this PR to address these types of NULL args

src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
```diff
@@ -346,14 +346,29 @@ HRESULT MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::GetILFunctionBo
...
+    if (ppMethodHeader == NULL)
+    {
+        ppMethodHeader = &pIgnoreMethodHeader;
+    }
+
+    ULONG cbIgnoreMethodSize;
+    if (pcbMethodSize == NULL)
+    {
+        pcbMethodSize = &cbIgnoreMethodSize;
+    }
```
delmyers: Rather than doing this, you could just pass ppMethodHeader and pcbMethodSize into pModuleInfo-GetMethodIl(...). Then you don't have to do any of your own checks. They aren't used in this function except for as out parameters from other functions.

WilliamXieMSFT: The issue is that the CLR allows these out parameters to be optional so I am handling them here by creating dummy local variables that do nothing.

I want these checks to be at the CCorProfilerInfoWrapper level since this function is what is called by the raw profiler.

delmyers: I understand that CLR allows the out parameters to be optional, but that doesn't mean that you have to do any additional work to "ignore them". If they aren't used in the function, then you don't need to have local variables that represent them. Just pass the pointers directly through. You can check them for not-null, and initialize them to avoid having the IfFailRets leaving them as garbage.

The hiccup that you may have with that is that the contract for IMethodInfo and IModuleInfo may be different (do they require the out parameters to be non-null? But, if that is the case, you are going to get inconsistent results from this function regardless.

---
In reply to: [363900865](https://github.com/microsoft/CLRInstrumentationEngine/pull/187#discussion_r363900865) [](ancestors = 363900865)

WilliamXieMSFT: I see what you mean now. I will add null checks to conditionally initialize/set the deref ptr values. 

---
In reply to: [365929212](https://github.com/microsoft/CLRInstrumentationEngine/pull/187#discussion_r365929212) [](ancestors = 365929212,363900865)

PR #200, opened 2020-02-04
Fix handling nullptr when cleaning up during ReJIT

WilliamXieMSFT:
It turns out on GetReJITParameters() method end, ~CRemoveMethodInfoAfterRejit invokes CModuleInfo::SetRejitMethodInfo with pMethodInfo as nullptr. The SetRejitMethodInfo does not remove the MethodToken from the map when setting the nullptr.

This addresses #197

### DIFF
--- a/src/InstrumentationEngine/ModuleInfo.cpp
+++ b/src/InstrumentationEngine/ModuleInfo.cpp
@@ -907,14 +907,22 @@ HRESULT MicrosoftInstrumentationEngine::CModuleInfo::SetRejitMethodInfo(_In_ mdT
     if (pMethodInfo != nullptr)
     {
         std::unordered_map<mdMethodDef, CComPtr<CMethodInfo>>::iterator iter = m_methodInfosByToken.find(methodToken);
-
         if (iter != m_methodInfosByToken.end())
         {
             CLogging::LogMessage(_T("CModuleInfo::SetRejitMethodInfo - method token already exists %04x. Overwriting"), methodToken);
         }
-    }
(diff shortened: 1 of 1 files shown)


src/InstrumentationEngine/ModuleInfo.cpp
```diff
@@ -907,14 +907,22 @@ HRESULT MicrosoftInstrumentationEngine::CModuleInfo::SetRejitMethodInfo(_In_ mdT
...
+            CLogging::LogMessage(_T("CModuleInfo::SetRejitMethodInfo - adding new method token %04x."), methodToken);
+        }
 
-    m_methodInfosByToken[methodToken] = pMethodInfo;
+        m_methodInfosByToken[methodToken] = pMethodInfo;
+    }
+    else
+    {
+        CLogging::LogMessage(_T("CModuleInfo::SetRejitMethodInfo - erasing existing method token %04x."), methodToken);
+        m_methodInfosByToken.erase(methodToken);
```
WilliamXieMSFT: >erase [](start = 29, length = 5)

I'll add another check to ensure the MethodToken exists before erasing

WilliamXieMSFT: erase will return 0 if the element does not exist

---
In reply to: [375417655](https://github.com/microsoft/CLRInstrumentationEngine/pull/200#discussion_r375417655) [](ancestors = 375417655)

sanikolov:
confirmed fixed, thanks

Issue #201, opened 2020-02-14
RawProfilerHook bug with DynaTrace - SetEventMask


WilliamXieMSFT:
Hooking up DynaTrace on AzureAppService with InstrumentationEngine RawProfilerHook. The following error occurs:

LogError[21:48:11]:SetEventMask can only be called during initialize of Host or InstrumentationMethod

WilliamXieMSFT:
ICorProfiler calls CProfilerManager::Initialize() method which enables m_bIsInIntialize, calls Initialize() on the RawProfiler, and then disables m_bIsInInitialize. 

It appears DynaTrace profiler calls SetEventMask outside of CProfilerManager::Initialize() and errors out because m_bIsInInitialize is false.

WilliamXieMSFT:
If I skip the SetEventMask section, the process throws an exception: 
Exception during profiler callback MicrosoftInstrumentationEngine::CProfilerManager::ModuleLoadStarted
* ExceptionCode 0xc0000005 (AV)
* ModuleId 0x00007ffaf97c4020

WilliamXieMSFT:
See https://github.com/dotnet/runtime/blob/fcd862e06413a000f9cafa9d2f359226c60b9b42/src/coreclr/src/vm/eetoprofinterfaceimpl.cpp#L2263

PR #202, opened 2020-02-18
Fix SetEventMask to allow mutable flag updates

WilliamXieMSFT:


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -743,18 +743,18 @@ HRESULT CProfilerManager::DetermineClrVersion()
         // TODO-leonidd-080923: Use GetRuntimeInfo when available to set the CLR version
         if (CComQIPtr<ICorProfilerInfo4> pi4 = m_pRealProfilerInfo.p)
         {
-            m_attachedClrVersion = ClrVersion_4_5; // v4.5
-            return S_OK;
+        m_attachedClrVersion = ClrVersion_4_5; // v4.5
```
jander-msft: Please fix indentation

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -841,8 +841,17 @@ HRESULT CProfilerManager::SetEventMask(DWORD dwEventMask)
     }
     else
     {
-        CLogging::LogError(_T("SetEventMask can only be called during initialize of Host or InstrumentationMethod"));
-        return E_FAIL;
+        // Prevent immutable flags from being set.
+        if ((dwEventMask & COR_PRF_MONITOR_IMMUTABLE) !=
```
jander-msft: Make this "else if" and flip the condition so that true is valid (and sets the flags) and the else block is the failure.

jander-msft: I think you need to perform a similar check for COR_PRF_HIGH_MONITOR_IMMUTABLE in SetEventMask2, but I don't know if you need to absolutely fix that right now.

WilliamXieMSFT: Yep, I saw that as well. I'll try to address that next

---
In reply to: [380925449](https://github.com/microsoft/CLRInstrumentationEngine/pull/202#discussion_r380925449) [](ancestors = 380925449)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -870,7 +876,22 @@ HRESULT CProfilerManager::SetEventMask2(_In_ DWORD dwEventMaskLow, _In_ DWORD dw
 {
     HRESULT hr = S_OK;
 
-    m_dwEventMaskHigh |= dwEventMaskHigh;
+    if (GetIsInInitialize())
```
jander-msft: I think we also need to handle when GetIsInitializingInstrumentationMethod() is true here, but I would make a separate change for that. The issue of the high event mask being set directly by the instrumentation methods already exists (if it's an issue at all).

WilliamXieMSFT: We will also have to revisit this behavior for attach

---
In reply to: [380962438](https://github.com/microsoft/CLRInstrumentationEngine/pull/202#discussion_r380962438) [](ancestors = 380962438)

WilliamXieMSFT:
#201

PR #204, opened 2020-02-18
Initial attach support for engine and instrumentation methods.

jander-msft:
These changes refactor the extensions host into the instrumentation engine and allow profile-on-launch and profile-on-attach to share the same initialization logic. I've tested that these changes do not break the existing profile-on-launch scenario for both Windows and Linux.

Overview of changes:
- Initial attach support for engine and instrumentation methods.
- Merge extensions host into instrumentation engine.

What's not in these changes:
- Deserialization of the attach data (instrumentation method settings, raw profiler settings, and engine settings).
- Any settings that are loaded via environment variables have not been changed and will be changed at a later time when we determine which ones need to be fed through the attach data rather than always loaded via environment variables.


src/Common.Lib/Macros.h
```diff
@@ -65,8 +65,8 @@
 #endif
 
 #ifndef IfFalseRet
-#define IfFalseRet(EXPR) \
-    do { if (FALSE == (EXPR)) { CLogging::LogError(_T("IfFalseRet(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__); return E_FAIL; } } while (false)
+#define IfFalseRet(EXPR, HR) \
```
WilliamXieMSFT: >HR [](start = 25, length = 2)

super nit, but I think it would improve readability of the HR is the first parameter so that it would read "IfFalseRet(E_Fail, .....)"

WilliamXieMSFT: I know this would change a lot of code, so I don't mind it as-is.

---
In reply to: [381498139](https://github.com/microsoft/CLRInstrumentationEngine/pull/204#discussion_r381498139) [](ancestors = 381498139)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2672,10 +2770,24 @@ HRESULT CProfilerManager::InitializeForAttach(
...
     PROF_CALLBACK_BEGIN
 
+    m_bAttach = true;
+
+    vector<CComPtr<CConfigurationSource>> configSources;
+
+    // TODO: Parse coniguration information from pvClientData.
+
+    // Mark that this is during the initialize call. This enables operations that can only be supported during initialize
+    CInitializeHolder initHolder(this);
```
WilliamXieMSFT: >CInitializeHolder [](start = 4, length = 17)

Should we have a different holder (ie. CInitializeAttachHolder) since there will be restrictions on SetEventMask calls, and possibly other callbacks that act differently between initialize/initializeForAttach?

jander-msft: Very possible. For now, I'm going to use the same holder and add a task to create a new one that knows about what's allowed during attach. At this time, the Production Breakpoints IM works with the code as it is.

src/InstrumentationEngine.Api/InstrumentationEngine.idl
```diff
@@ -1551,4 +1551,83 @@ library MicrosoftInstrumentationEngine
...
+
+    [
+        object,
+        uuid(9B03D87E-72F0-4D8E-A4B1-15BCD8227073),
+        pointer_default(unique)
+    ]
+    interface IEnumInstrumentationMethodSettings : IUnknown
+    {
+        HRESULT Next(
+            [in] ULONG celt,
```
WilliamXieMSFT: >celt [](start = 23, length = 4)

what does "celt" stand for? and "rgelt"?

src/Common.Lib/Macros.h
jander-msft: I think it's more important to read what is being tested first rather than what the return value is for failure. Also, there will be cases where "HRESULT_FROM_WIN32(GetLastError())" is the return value, which would push back the tested value much further out of view.

src/InstrumentationEngine.Lib/ConfigurationLocator.cpp
```diff
@@ -0,0 +1,211 @@
...
+                wstring wsConfigPath = PathUtils::Combine(strModuleFolder, findFileData.cFileName);
+
+                IfFailRet(AddSource(sources, wsConfigPath.c_str()));
+            } while (FindNextFile(hSearchHandle, &findFileData));
+        }
+
+        return S_OK;
+    }
+
+    HRESULT CConfigurationLocator::AddExpandedPath(
```
WilliamXieMSFT: >AddExpandedPath [](start = 35, length = 15)

// static

src/InstrumentationEngine.Lib/EnumeratorImpl.h
```diff
@@ -0,0 +1,176 @@
...
+
+            if (dwLength != 0)
+            {
+                m_members.Attach(new CComPtr<ENUM_ITEM_TYPE>[dwLength]);
+                if (!m_members)
+                {
+                    return E_OUTOFMEMORY;
+                }
+
+                for (DWORD i = 0; i < m_dwLength; i++)
```
WilliamXieMSFT: >m_dwLength [](start = 38, length = 10)

m_dewLength = dwLength?

src/InstrumentationEngine.Api/InstrumentationEngine.idl
WilliamXieMSFT: Oh, elt = element

---
In reply to: [381505728](https://github.com/microsoft/CLRInstrumentationEngine/pull/204#discussion_r381505728) [](ancestors = 381505728)

src/Common.Lib/ModuleUtils.cpp
```diff
@@ -6,33 +6,29 @@
...
-namespace Agent
-{
-namespace Io
-{
-namespace ModuleUtils
+namespace CommonLib
 {
 
-	HMODULE g_hModuleMarker = nullptr;
+	HMODULE ModuleUtils::s_hModuleMarker = nullptr;
```
WilliamXieMSFT: >	 [](start = 0, length = 1)

Mind converting these tabs to spaces?

src/Common.Lib/PathUtils.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "PathUtils.h"
+
+#pragma warning(push)
+#pragma warning(disable: 4995) // disable so that memcpy, wmemcpy can be used
+#include <sstream>
+#pragma warning(pop)
+
+namespace CommonLib
+{
+	bool PathUtils::Exists(const std::wstring& strPath)
```
WilliamXieMSFT: >	 [](start = 0, length = 1)

Mind converting these tabs to spaces?

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -587,7 +612,17 @@ HRESULT CProfilerManager::AddInstrumentationMethod(
         return E_OUTOFMEMORY;
     }
 
-    return this->AddInstrumentationMethod(pInstrumentationMethod.release(), ppInstrumentationMethod);
+    CComPtr<CConfigurationSource> pSource;
```
WilliamXieMSFT: >pSource [](start = 34, length = 7)

Currently, this CConfigurationSource will always return an empty IEnumInstrumentationMethodSettings. Would it make more sense to just pass NULL to the following AddInstrumentationMethod()?

src/InstrumentationEngine.Lib/ConfigurationSource.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace MicrosoftInstrumentationEngine
+{
+    class CConfigurationSource :
+        public CModuleRefCount,
+        public IUnknown
+    {
+    private:
+        CComBSTR m_bstrPath;
+        std::vector<ATL::CComPtr<IInstrumentationMethodSetting>> m_settings;
```
WilliamXieMSFT: Will we merge the two concepts together in the future? ie. parse the m_bstrPath into IInstrumentationMethodSettings? It feels disjoint that the constructor takes in the xml path but requires AddSetting calls to populate the vector.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -113,14 +120,15 @@ void CProfilerManager::FinalRelease()
 
 
 HRESULT CProfilerManager::SetupProfilingEnvironment(_In_reads_(numConfigPaths) BSTR rgConfigPaths[], _In_ UINT numConfigPaths)
```
wiktork: >SetupProfilingEnvironment [](start = 26, length = 25)

I don't know if this has any InstrumentationMethod callers, but since it's part of the api surface this could be considered a breaking change.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -269,13 +277,11 @@ HRESULT CProfilerManager::GetCorProfilerInfo(
 // return the profiler host instance
 HRESULT CProfilerManager::GetProfilerHost(_Out_ IProfilerManagerHost** ppProfilerManagerHost)
```
wiktork: >GetProfilerHost [](start = 26, length = 15)

Similiar concern here regarding breaking api changes.

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Perhaps update the comment in IDL to say this function is obsolete?

---
In reply to: [381589507](https://github.com/microsoft/CLRInstrumentationEngine/pull/204#discussion_r381589507) [](ancestors = 381589507)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -819,6 +847,78 @@ DWORD CProfilerManager::CalculateEventMask(DWORD dwAdditionalFlags)
...
+    IfNullRetPointer(pCorProfilerInfoUnk);
+
+    HRESULT hr = S_OK;
+
+    IfFailRet(pCorProfilerInfoUnk->QueryInterface(__uuidof(ICorProfilerInfo), (LPVOID*)&m_pRealProfilerInfo));
+
+    IfFailRet(DetermineClrVersion());
+
+    //set event mask to be default one. Later host and Instrumentation methods may call it again
+    IfFailRet(this->SetEventMask(m_dwEventMask));
```
wiktork: >SetEventMask [](start = 20, length = 12)

Are there limitations on SetEventMask for attach vs. startup?

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2687,10 +2799,35 @@ HRESULT CProfilerManager::ProfilerAttachComplete(void)
...
+        ULONG ulFetched = 0;
+        ModuleID moduleIds[1];
+        IfFailRet(pModuleEnum->Next(1, moduleIds, &ulFetched));
+
+        if (fContinueModuleEnum = (S_OK == hr))
+        {
+            // Constructing the module info will also construct the assembly
+            // and app domain info associated with the module.
+            CComPtr<IModuleInfo> pModuleInfo;
+            IfFailRet(ConstructModuleInfo(moduleIds[0], &pModuleInfo));
```
wiktork: >ConstructModuleInfo [](start = 22, length = 19)

I can't remember what we discussed here, but we don't want to "replay" ModuleLoaded here correct, but rather each IM would decide how to instrument inside AttachComplete?

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Yep, [https://github.com/dotnet/runtime/blob/master/docs/design/coreclr/profiling/davbr-blog-archive/Attach.md#specific-callback-limitations](https://github.com/dotnet/runtime/blob/master/docs/design/coreclr/profiling/davbr-blog-archive/Attach.md#specific-callback-limitations)

This PR won't address these issues, but future ones would.

---
In reply to: [381591393](https://github.com/microsoft/CLRInstrumentationEngine/pull/204#discussion_r381591393) [](ancestors = 381591393)

src/InstrumentationEngine.Lib/EnumeratorImpl.h
```diff
@@ -0,0 +1,176 @@
...
+#pragma once
+
+#include "DataContainer.h"
+
+using namespace ATL;
+
+namespace MicrosoftInstrumentationEngine
+{
+    template<class ENUM_INTERFACE, class ENUM_ITEM_TYPE>
+    class CEnumerator : public ENUM_INTERFACE, public CDataContainer
```
wiktork: >CEnumerator [](start = 10, length = 11)

Are there some missing deletes from the PR? (such as this file)

src/InstrumentationEngine.Api/InstrumentationEngine_api.cpp
```diff
@@ -9,8 +9,8 @@
 /* at Mon Jan 18 19:14:07 2038
  */
 /* Compiler settings for InstrumentationEngine.idl:
-    Oicf, W1, Zp8, env=Win64 (32b run), target_arch=AMD64 8.01.0622 
-    protocol : all , ms_ext, c_ext, robust
+    Oicf, W1, Zp8, env=Win32 (32b run), target_arch=X86 8.01.0622
```
jander-msft: I don't think it's necessary, but I'll revert these comments.

src/InstrumentationEngine.Lib/ConfigurationSource.h
jander-msft: m_bstrPath is the path to the configuration file that contains the IM information. m_settings is effectively a key-value pair of settings that will be passed to the IM when it is initialized; such settings are the ones that each IM is currently setting via environment variables (e.g. MicrosoftProductionDiagnostics_RuntimeFolder in the case of the Production Breakpoints IM). This is needed because if you are attaching, you likely didn't have an opportunity to set environment variables on the process before it started.

src/InstrumentationEngine.Lib/EnumeratorImpl.h
jander-msft: I didn't change this code, but just moved the file (actually, I copied the file and forgot to delete the original). I'm going to leave this as-is an log an issue for it.

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: I can easily implement this to forward to the new SetupProfilingEnvironment (which I'll probably rename to something more descriptive, like SetupInstrumentationMethods).

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: I think it's better as-is so that downstream doesn't have to account for nullptr. All of the SAL annotations say that it will not be nullptr ( _In_ ).

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: That's what I'm thinking for now. Makes it easier for each IM to decide what it wants to do.

In the future, we could add a new IInstrumentationMethodAttach* interface with a method that allows the engine to ask the method if it would like the engine to replay app domain, assembly, and module loads for it. The caveat is that the IM would have to know to call ApplyMetadata during those replays on attach whereas for profiling at startup does not require the ApplyMetadata call, but that could just easily be documented (or if we want to get crafty, we could monitor the metadata emit interfaces and automatically apply the changes if changes were made).

src/InstrumentationEngine.Lib/EnumeratorImpl.h
jander-msft: Good catch. Looks like I copied instead of moved them.

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: I think the API became defunct when the profiler manager host extensibility was removed. In all cases, it was made to just return CExtensionsHost, which only implemented IProfilerManagerHost. So calling the Initialize method on it after getting it from this API would either (a) do nothing, or (b) reek havoc on the engine (it probably does the former since it would try to load everything again, but everything was already loaded, so it should no-op).

I got rid of CExtensionsHost because it would be silly to have the engine already know what the instrumentation methods are during attach and pass them (along with their settings) to the CExtensionsHost and have it just pass it back. So I folded CExtensionsHost into CProfilerManager to avoid the middle man.

Any suggestion what should be done for this API?

src/InstrumentationEngine.Api/InstrumentationEngine_api.cpp
WilliamXieMSFT: These get changed based on change in compilation. I'm not sure if there's a way to prevent it but it's benign

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Sure, not a nullptr, but why call EnumSettings from a newly created ConfigurationSource (which won't have any settings)?

src/InstrumentationEngine.Lib/ConfigurationSource.h
WilliamXieMSFT: I am confused by the code in ProfilerManager that creates a new ConfigurationSource but calls EnumSettings, which would always return an empty vector.

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: #205

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: #205

src/InstrumentationEngine.Api/InstrumentationEngine.idl
```diff
@@ -1551,4 +1551,83 @@ library MicrosoftInstrumentationEngine
...
+
+        HRESULT Skip(
+            [in] ULONG celt
+            );
+
+        HRESULT Reset(void);
+
+        HRESULT Clone(
+            [out]IEnumInstrumentationMethodSettings** ppenum
+            );
```
delmyers: What is the purpose of the Clone function? I understand that it would be to copy the enum to another enum. But would the clone be a deep copy, or a shallow copy? Would it just be the same enum with the index reset? I'm a little confused on why you would want to clone an existing enum rather than just getting another enum from the same object.

src/InstrumentationEngine.Api/InstrumentationEngine_api.cpp
```diff
@@ -227,6 +227,18 @@ MIDL_DEFINE_GUID(IID, IID_IMethodJitInfo2,0x8311A7CF,0x30EC,0x42C9,0x85,0xA4,0xF
...
+MIDL_DEFINE_GUID(IID, IID_IInstrumentationMethodSetting,0xEF0B0C79,0x08E7,0x4C3A,0xA4,0xC5,0x02,0xA9,0xC9,0xCE,0x88,0x09);
+
+
+MIDL_DEFINE_GUID(IID, IID_IEnumInstrumentationMethodSettings,0x9B03D87E,0x72F0,0x4D8E,0xA4,0xB1,0x15,0xBC,0xD8,0x22,0x70,0x73);
+
+
+MIDL_DEFINE_GUID(IID, IID_IInstrumentationMethodAttachContext,0x2C37B76C,0xB350,0x4738,0x8B,0x29,0xB9,0x2C,0x7E,0xD6,0xC5,0x22);
+
+
+MIDL_DEFINE_GUID(IID, IID_IInstrumentationMethodAttach,0x3BD6C171,0x4F3C,0x45C3,0x8C,0xB9,0xBC,0x8C,0x33,0x7D,0x16,0x83);
```
delmyers: Update InstrumentationEngine_i.cpp

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: Because it will return an object that has nothing to enumerate so that I can pass in something that is not nullptr. And I don't want to just new up a concrete instance of IEnumInstrumentationMethodSettings and pass that in because that would require me to create a vector<CComPtr<IInstrumentationMethodSetting>>, make an enumerator from that, and then pass that in. It's just as much code as what I have already, but introduces implementation detail to the ProfilerManager that it doesn't really care to know about. It is primarily concerned about creating sources and passing them through for use elsewhere.

src/InstrumentationEngine.Lib/InstrumentationMethodAttachContext.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "InstrumentationMethodAttachContext.h"
+
+namespace MicrosoftInstrumentationEngine
+{
+    CInstrumentationMethodAttachContext::CInstrumentationMethodAttachContext(_In_ IEnumInstrumentationMethodSettings* pSettingsEnum) noexcept :
+        m_pSettingsEnum(pSettingsEnum)
+    {
+    }
+
+    STDMETHODIMP CInstrumentationMethodAttachContext::QueryInterface(_In_ REFIID riid, _Out_ void** ppvObject)
```
delmyers: >(_In_ REFIID riid, _Out_ void** ppvObject) [](start = 68, length = 42)

do you need to add "noexcept" to these as well, to avoid warnings?

src/InstrumentationEngine.Api/InstrumentationEngine.idl
jander-msft: It creates a new CEnumerator instance with a new CComPtr vector of the same pointers elements from the existing enumerator. See CEnumerator::Clone in src/InstrumentationEngine.Lib/EnumeratorImpl.h.

Every other IEnum* interface in the IDL file uses this pattern, so I'm continuing it here.

jander-msft:
azp /run

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -587,7 +624,17 @@ HRESULT CProfilerManager::AddInstrumentationMethod(
         return E_OUTOFMEMORY;
     }
 
-    return this->AddInstrumentationMethod(pInstrumentationMethod.release(), ppInstrumentationMethod);
+    CComPtr<CConfigurationSource> pSource;
+    pSource.Attach(new CConfigurationSource(bstrModulePath));
+    if (nullptr == pSource)
+    {
+        return E_OUTOFMEMORY;
+    }
```
delmyers: Have we set the flag to not throw on "new" for this codebase? If not, there is no point in testing this. Othewise, if you want to do this, then you should use "new nothrow" above.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -819,6 +859,78 @@ DWORD CProfilerManager::CalculateEventMask(DWORD dwAdditionalFlags)
...
+    {
+        IfFailRet(SetupInstrumentationMethods(configSources));
+    }
+    else
+    {
+        CLogging::LogError(_T("No instrumentation method configs found to load in process %ul"), GetCurrentProcessId());
+    }
+
+#ifndef PLATFORM_UNIX
+    IfFailRet(SetupRawProfiler());
```
delmyers: >SetupRawProfiler() [](start = 14, length = 18)

Do we just not support the Raw profiler on Linux?

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2672,10 +2782,24 @@ HRESULT CProfilerManager::InitializeForAttach(
...
 
-    // TODO: How to handle this? Does this need to support profiler attach? How does that impact initlization etc...
-
     PROF_CALLBACK_BEGIN
 
+    m_bAttach = true;
+
+    vector<CComPtr<CConfigurationSource>> configSources;
+
+    // TODO: Parse coniguration information from pvClientData.
```
delmyers: >coniguration [](start = 19, length = 12)

configuration

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: Previous to this change, we did not, so I'd like to not change that at this time.

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Thanks for the explanation!

src/InstrumentationEngine.Lib/InstrumentationMethodAttachContext.cpp
jander-msft: I don't think so. I don't see it on IUnknown::QueryInterface and I don't see build warnings for this in the PR build nor on my local machine. What warning do you expect?

---
In reply to: [382173711](https://github.com/microsoft/CLRInstrumentationEngine/pull/204#discussion_r382173711) [](ancestors = 382173711)

delmyers: Sometimes you get warnings about declarations mismatching definitions with this kind of thing. If you don't get a warning, that's fine.

---
In reply to: [382213914](https://github.com/microsoft/CLRInstrumentationEngine/pull/204#discussion_r382213914) [](ancestors = 382213914,382173711)

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: I added a comment in the change log that this method is deprecated and added helpstring to the method in the IDL stating the same.

PR #212, opened 2020-02-26
Initial attach executable

jander-msft:
Create a .NET Core 3.1 executable that simplifies the attach process for the Instrumentation Engine. It currently only takes the process ID to which it should attach the engine. It reports errors to standard error and returns a non-zero exit code on error. Success will return zero.

### DIFF
--- a/src/InstrumentationEngine.Attach/InstrumentationEngine.Attach.csproj
+++ b/src/InstrumentationEngine.Attach/InstrumentationEngine.Attach.csproj
@@ -7,7 +7,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-    <AssemblyName>attach</AssemblyName>
+    <AssemblyName>Microsoft.InstrumentationEngine.Attach</AssemblyName>
     <RootNamespace>Microsoft.InstrumentationEngine</RootNamespace>
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
--- a/src/InstrumentationEngine.Attach/InstrumentationEngine.Attach.csproj
+++ b/src/InstrumentationEngine.Attach/InstrumentationEngine.Attach.csproj
@@ -1,0 +1,22 @@
+﻿<Project>
+  <PropertyGroup>
+    <SubComponent>Tools\Attach</SubComponent>
+  </PropertyGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'EnlistmentRoot.marker'))\build\Managed.props" />
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AssemblyName>attach</AssemblyName>
+    <RootNamespace>Microsoft.InstrumentationEngine</RootNamespace>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <Nullable>Enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Diagnostics.NETCore.Client" Version="0.2.61701" />
+    <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" Version="$(MicroBuildCoreVersion)" />
+  </ItemGroup>
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
+  <Import Project="$(EnlistmentRoot)\build\Managed.targets" />
+</Project>
--- a/src/InstrumentationEngine.Attach/Program.cs
+++ b/src/InstrumentationEngine.Attach/Program.cs
@@ -19,6 +19,13 @@ namespace Microsoft.InstrumentationEngine
         private const int ExitCodeSuccess = 0;
         private const int ExitCodeFailure = -1;
 
+        private const string Instrumentation32FolderName = "Instrumentation32";
+        private const string Instrumentation64FolderName = "Instrumentation64";
+
+        private const string InstrumentationEngineX64LinuxName = "libInstrumentationEngine.so";
+        private const string InstrumentationEngineX64WindowsName = "MicrosoftInstrumentationEngine_x64.dll";
+        private const string InstrumentationEngineX86WindowsName = "MicrosoftInstrumentationEngine_x86.dll";
+
         internal static int Main(string[] args)
         {
             // TODO: Parse command line for additional arguments such as configuration files
@@ -110,13 +117,17 @@ namespace Microsoft.InstrumentationEngine
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 enginePath = isTargetProcess32Bit ?
-                    Path.Combine(rootDirectory, "Instrumentation32", "MicrosoftInstrumentationEngine_x86.dll") :
-                    Path.Combine(rootDirectory, "Instrumentation64", "MicrosoftInstrumentationEngine_x64.dll");
+                    Path.Combine(rootDirectory, Instrumentation32FolderName, InstrumentationEngineX86WindowsName) :
+                    Path.Combine(rootDirectory, Instrumentation64FolderName, InstrumentationEngineX64WindowsName);
+            }
+            else if (!isTargetProcess32Bit)
+            {
+                enginePath = Path.Combine(rootDirectory, Instrumentation64FolderName, InstrumentationEngineX64LinuxName);
             }
             else
             {
-                Debug.Assert(!isTargetProcess32Bit, "Only 64 bit is supported on non-Windows platforms.");
-                enginePath = Path.Combine(rootDirectory, "Instrumentation64", "libInstrumentationEngine.so");
+                WriteError("Only the 64 bit engine is supported on non-Windows platforms.");
+                return ExitCodeFailure;
             }
 
             if (!File.Exists(enginePath))
--- a/src/InstrumentationEngine.Attach/Program.cs
+++ b/src/InstrumentationEngine.Attach/Program.cs
@@ -7,8 +7,6 @@ using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Text;
-using System.Text.Json;
 using Microsoft.Diagnostics.NETCore.Client;
 using static System.FormattableString;
 
--- a/src/InstrumentationEngine.Attach/Program.cs
+++ b/src/InstrumentationEngine.Attach/Program.cs
@@ -18,21 +18,24 @@ namespace Microsoft.InstrumentationEngine
     {
         private static readonly Guid InstrumentationEngineClsid = new Guid("324F817A-7420-4E6D-B3C1-143FBED6D855");
 
-        internal static void Main(string[] args)
+        private const int ExitCodeSuccess = 0;
+        private const int ExitCodeFailure = -1;
+
+        internal static int Main(string[] args)
         {
             // TODO: Parse command line for additional arguments such as configuration files
             if (args.Length != 1)
             {
-                WriteErrorJson("Expected single argument that represents the process ID to which the engine shall be attached.");
-                return;
+                WriteError("Expected single argument that represents the process ID to which the engine shall be attached.");
+                return ExitCodeFailure;
             }
 
             string processIdString = args[0];
 
             if (!Int32.TryParse(processIdString, out int processId))
             {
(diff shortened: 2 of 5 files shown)


src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,170 @@
...
+namespace Microsoft.InstrumentationEngine
+{
+    internal sealed class Program
+    {
+        private static readonly Guid InstrumentationEngineClsid = new Guid("324F817A-7420-4E6D-B3C1-143FBED6D855");
+
+        internal static void Main(string[] args)
+        {
+            // TODO: Parse command line for additional arguments such as configuration files
+            if (args.Length != 1)
```
WilliamXieMSFT: >!= [](start = 28, length = 2)

< 1 ?

src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,170 @@
...
+                        }
+                    }
+                }
+                catch (Win32Exception ex)
+                {
+                    WriteErrorJson(Invariant($"Failed to check process bitness (code: {ex.NativeErrorCode}): {ex.Message}"));
+                    return;
+                }
+
+                enginePath = isWow64 ?
```
WilliamXieMSFT: >isWow64 [](start = 29, length = 7)

IsWow64 returns false on 32-bit Windows. Do we care about that?

WilliamXieMSFT: The Proxy uses this to determine whether to use ProgramFiles or ProgramFiles(x86) which is nice since ProgramFiles corresponds to the bitness of the machine.

Could we also use agent's bitness to determine whether we're running on 32-bit or 64-bit process if IsWow64 is false?

---
In reply to: [384696104](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384696104) [](ancestors = 384696104)

src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                enginePath = Path.Combine(rootDirectory, "Instrumentation64", "libInstrumentationEngine.so");
+            }
+            else
+            {
+                bool isWow64;
+                try
+                {
+                    // isWow64 is true for 32-bit applications running on a 64-bit OS
+                    if (!NativeMethods.IsWow64Process(process.Handle, out isWow64))
```
delmyers: >NativeMethods.IsWow64Process [](start = 25, length = 28)

Amy solution for Linux? Or do we only support 64bit on Linux?

src/InstrumentationEngine.Attach/Program.cs
delmyers: The correct way to do this is:

bool is64Bit = (sizeof(IntPtr) == 8) && !isWow64
enginePath = (is64Bit) ? ....


---
In reply to: [384696104](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384696104) [](ancestors = 384696104)

delmyers: Oh, but you also have to take into account whether *this process* is 64 bit:
bool thisIs64 = (sizeof(IntPtr) == 8) && !thisIsWow;
bool is64BitProc = (thisIs64 && !isWow) || ((!thisIs64 && thisIsWow) && !isWow);

Or something of the like :-)


---
In reply to: [384715740](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384715740) [](ancestors = 384715740,384696104)

src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,170 @@
...
+                WriteErrorJson(Invariant($"Could not attach engine to process: '{ex.Message}'."));
+                return;
+            }
+
+            WriteSuccessJson();
+        }
+
+        private static void WriteErrorJson(string message)
+        {
+            Console.WriteLine(CreateResultJson("Failed", message));
```
delmyers: >CreateResultJson [](start = 30, length = 16)

Why in Json?

src/InstrumentationEngine.Attach/Program.cs
WilliamXieMSFT: We currently only support 64bit Linux. I don't think we have plans to support 32bit.

---
In reply to: [384714445](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384714445) [](ancestors = 384714445)

src/InstrumentationEngine.Attach/Program.cs
jander-msft: It's also an error if there is more than one argument as it is currently parsed.

---
In reply to: [384688540](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384688540) [](ancestors = 384688540)

src/InstrumentationEngine.Attach/Program.cs
jander-msft: This code is specifically checking if the target process (not the current process) is 32 bit or not. Anything the current process indicates should not be considered.

---
In reply to: [384718155](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384718155) [](ancestors = 384718155,384715740,384696104)

src/InstrumentationEngine.Attach/Program.cs
jander-msft: On Linux, there is no 32 bit support for .NET Core (except for ARM, but CLRIE does not support that at all).

---
In reply to: [384714445](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384714445) [](ancestors = 384714445)

src/InstrumentationEngine.Attach/Program.cs
jander-msft: Currently, CLRIE is only running on 64 bit OSes in production diagnostics scenarios. But I can add some code to check the OS bitness in the case that isWow64 is false.

---
In reply to: [384713327](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384713327) [](ancestors = 384713327,384696104)

src/InstrumentationEngine.Attach/Program.cs
jander-msft: The executable is going to be executed by an http handler. I felt it was best to use a standard data format that it could easily consume.

---
In reply to: [384719780](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384719780) [](ancestors = 384719780)

src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,170 @@
...
+using Microsoft.Diagnostics.NETCore.Client;
+using static System.FormattableString;
+
+namespace Microsoft.InstrumentationEngine
+{
+    internal sealed class Program
+    {
+        private static readonly Guid InstrumentationEngineClsid = new Guid("324F817A-7420-4E6D-B3C1-143FBED6D855");
+
+        internal static void Main(string[] args)
```
jander-msft: >void [](start = 24, length = 4)

Should change this to int and return a non-zero exit code if any failure occurs.

src/InstrumentationEngine.Attach/Program.cs
jander-msft: Maybe I should have it write messages to standard error and return a non-zero exit code. Then let the http handler convert it to whatever it needs.

---
In reply to: [384732346](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384732346) [](ancestors = 384732346,384719780)

src/InstrumentationEngine.Attach/InstrumentationEngine.Attach.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project>
+  <PropertyGroup>
+    <SubComponent>Tools\Attach</SubComponent>
+  </PropertyGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'EnlistmentRoot.marker'))\build\Managed.props" />
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AssemblyName>attach</AssemblyName>
```
wiktork: Did you want to make this name a little more descriptive? i.e. Microsoft.InstrumentationEngine.Attach.exe or something simliar?

src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,150 @@
...
+            string enginePath;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                enginePath = isTargetProcess32Bit ?
+                    Path.Combine(rootDirectory, "Instrumentation32", "MicrosoftInstrumentationEngine_x86.dll") :
+                    Path.Combine(rootDirectory, "Instrumentation64", "MicrosoftInstrumentationEngine_x64.dll");
+            }
+            else
+            {
+                Debug.Assert(!isTargetProcess32Bit, "Only 64 bit is supported on non-Windows platforms.");
```
wiktork: Consider WriteError

src/InstrumentationEngine.Preinstall/InstrumentationEngine.Preinstall.csproj
```diff
@@ -45,30 +45,35 @@
       <ZipItem Include="scmApplicationHost.xdt">
         <Destination>$(PackageVersion)</Destination>
       </ZipItem>
+      <!-- Tools -->
+      <ZipItem Include="$(InputBinCfgRoot)\AnyCPU\Tools\Attach\*"
+               Exclude="$(InputBinCfgRoot)\AnyCPU\Tools\Attach\*.pdb">
+        <Destination>$(PackageVersion)\Tools\Attach</Destination>
+      </ZipItem>
       <!-- Instrumentationx86Files -->
```
wiktork: Did this need to get added to the msi as well?

src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -0,0 +1,150 @@
...
+                        return ExitCodeFailure;
+                    }
+                }
+            }
+
+            string enginePath;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                enginePath = isTargetProcess32Bit ?
+                    Path.Combine(rootDirectory, "Instrumentation32", "MicrosoftInstrumentationEngine_x86.dll") :
```
wiktork: Can we push these to constants?

src/InstrumentationEngine.Preinstall/InstrumentationEngine.Preinstall.csproj
WilliamXieMSFT: Justin and I talked about this offline yesterday. 

We're foregoing that for now since it's currently not used and requires some more refactoring since attach.exe will need to interact with the proxy.

Our general design would have an "attachforwarder.exe" that calls to the specific CLRIE version of attach.exe so we can keep the API between CLRIE and attach.exe consistent.

---
In reply to: [384754757](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384754757) [](ancestors = 384754757)

jander-msft: Additionally, this executable is .NET Core 3.1, which will not work in the MSI case unless we do a self-contained publish or install .NET Core 3.1 on the machine as a dependency. Or we compile a .NET Framework version of this as well. Either way, we are holding off on making any decisions on that at this time.

---
In reply to: [384754757](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384754757) [](ancestors = 384754757)

src/InstrumentationEngine.Attach/Program.cs
delmyers: Doesn't matter. You can't tell if the target process is 32bit without knowing if *this* process is a WOW process or not. IsWOW64 tells you "the given process is a 32bit process on a 64bit machine". If you are running on a 32 bit machine IsWOW64 will always return false. So, if IsWOW64 returns false in this case, that means one of two things:

The target process is 64bit OR the target process is 32bit running on a 32bit OS.

To determine which of those is the case, you have to check what the OS architecture is. There is no way to check that. You have to check what the bitness of your current process is. The way to do that is to check the size of a pointer.

But wait, there's more!

The size of the pointer is not good enough. If it is 4bytes, then there are two options:

1) The OS architecture is 32bit
2) The OS architecture is 64bit, and this process is running under WOW64.

To figure out that, you need to check whether *this* process is WOW64, hence the complicated Boolean logic above.

---
In reply to: [384728717](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384728717) [](ancestors = 384728717,384718155,384715740,384696104)

delmyers: I had a larger comment about this, but then I saw that you are using OsArchitecture now. I didn't know that was available in .NET.

---
In reply to: [384728717](https://github.com/microsoft/CLRInstrumentationEngine/pull/212#discussion_r384728717) [](ancestors = 384728717,384718155,384715740,384696104)

Issue #176, opened 2019-12-02
Attach: Refactor environment variable usage into initialization pattern


jander-msft:
Components that use environment variables (e.g. logging) will likely need to be refactored to take settings during initialization so that attach can properly initialize them.

PR #216, opened 2020-02-28
Support configuration parsing for Profiler Attach

WilliamXieMSFT:
These changes are specifically the Profiler changes, which include:
* Adding XmlNode and XmlDocWrapper implementations
* Parsing the configuration xml from attach.exe for only adding instrumentation methods and configuration the engine.

The attach.exe changes will be in another PR

List of TODOs outside this PR:
* Support other portions of the configuration schema like "removing Instrumentation Method" and "attach RawProfiler"
* Test Linux changes

### DIFF
--- a/src/Common.Lib/Macros.h
+++ b/src/Common.Lib/Macros.h
@@ -22,6 +22,12 @@
 #define __FUNCTIONT__
 #endif
 
+    // Log an error without returning.
+#ifndef IfFailLogNoRet
+#define IfFailLogNoRet(EXPR) \
+    do { if (FAILED(hr = (EXPR))) { CLogging::LogError(_T("IfFailLogNoRet(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__ _T(": %d"), hr); } } while (false)
+#endif
+
 #ifndef IfFailRet
 #define IfFailRet(EXPR) \
     do { if (FAILED(hr = (EXPR))) { CLogging::LogError(_T("IfFailRet(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__); return hr; } } while (false)
@@ -41,6 +47,11 @@
 #ifndef IfFailRetNoLog
 #define IfFailRetNoLog(EXPR) \
     do { if (FAILED(hr = (EXPR))) { return hr; } } while (false)
+#endif
+
+#ifndef IfFalseFail
+#define IfFalseFail(EXPR) \
+    do { if (!(EXPR)) { CLogging::LogError(_T("IfFalseFail(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__); return E_FAIL; } } while (false)
 #endif
 
 // Treats errno 0 as severity=SUCCESS others to severity=ERROR, sets facility to null, and code to 16 least-significant bits of errno.
--- a/src/Common.Lib/Macros.h
+++ b/src/Common.Lib/Macros.h
@@ -23,19 +23,14 @@
 #endif
 
     // Log an error without returning.
-#ifndef IfFailLogNoRet
-#define IfFailLogNoRet(EXPR) \
-    do { if (FAILED(hr = (EXPR))) { CLogging::LogError(_T("IfFailLogNoRet(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__ _T(": %d"), hr); } } while (false)
+#ifndef IfFailLog
+#define IfFailLog(EXPR) \
+    do { if (FAILED(hr = (EXPR))) { CLogging::LogError(_T("IfFailLog(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__ _T(": %d"), hr); } } while (false)
 #endif
 
 #ifndef IfFailRet
 #define IfFailRet(EXPR) \
     do { if (FAILED(hr = (EXPR))) { CLogging::LogError(_T("IfFailRet(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__); return hr; } } while (false)
-#endif
-
-#ifndef IfFailRetNoLog
-#define IfFailRetNoLog(EXPR) \
-    do { if (FAILED(hr = (EXPR))) { return hr; } } while (false)
 #endif
 
 // Wrap errno_t result of EXPR in HRESULT and then IfFailRet.
@@ -47,11 +42,6 @@
 #ifndef IfFailRetNoLog
 #define IfFailRetNoLog(EXPR) \
     do { if (FAILED(hr = (EXPR))) { return hr; } } while (false)
-#endif
-
-#ifndef IfFalseFail
-#define IfFalseFail(EXPR) \
-    do { if (!(EXPR)) { CLogging::LogError(_T("IfFalseFail(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__); return E_FAIL; } } while (false)
 #endif
 
 // Treats errno 0 as severity=SUCCESS others to severity=ERROR, sets facility to null, and code to 16 least-significant bits of errno.
--- a/src/InstrumentationEngine.Lib/StringUtils.h
+++ b/src/InstrumentationEngine.Lib/StringUtils.h
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include <string>
+
+#ifdef PLATFORM_UNIX
+using tstring = std::basic_string<WCHAR, std::char_traits<WCHAR>, std::allocator<WCHAR>>;
+using tstringstream = std::basic_stringstream<WCHAR>;
+using tistringstream = std::basic_istringstream<WCHAR>;
+#else
+using tstring = std::wstring;
+using tstringstream = std::wstringstream;
+using tistringstream = std::wistringstream;
+#endif
+using tofstream = std::basic_ofstream<WCHAR, std::char_traits<WCHAR>>;
+
+class StringUtils
+{
+public:
+
+    // The maximum string length. Set to 10000;
+    static const size_t MAX_STRING_LEN;
+
+    // Gets the length of the given string using the the standard wcsnlen_s.
+    // Returns E_BOUNDS if the length of wszString is greater than MAX_STRING_LEN.
+    // For use with untrusted data which may not have a null terminator.
+    static HRESULT WStringLen(_In_ const WCHAR* wszString, _Out_ size_t& len);
+
+    // Gets the length of the given string using the standard strnlen.
+    // Returns E_BOUNDS if the length of wszString is greater than MAX_STRING_LEN.
+    // For use with untrusted data which may not have a null terminator.
+    static HRESULT StringLen(_In_ const char* szString, _Out_ size_t& len);
+
+    // Gets the length of the given string using the standard wcsnlen_s.
+    // Yields the same result as calling wcsnlen_s(wszString, MAX_STRING_LEN);
+    // For use with untrusted data which may not have a null terminator.
+    static size_t WStringLen(_In_ const WCHAR* wszString);
+
+    // Gets the length of the given string using the standard strnlen.
+    // Yields the same result as calling strnlen(szString, MAX_STRING_LEN);
+    // For use with untrusted data which may not have a null terminator.
+    static size_t StringLen(_In_ const char* szString);
+};
--- a/src/InstrumentationEngine.Lib/StringUtils.h
+++ b/src/InstrumentationEngine.Lib/StringUtils.h
@@ -5,17 +5,6 @@
 
 #include <string>
 
-#ifdef PLATFORM_UNIX
-using tstring = std::basic_string<WCHAR, std::char_traits<WCHAR>, std::allocator<WCHAR>>;
-using tstringstream = std::basic_stringstream<WCHAR>;
-using tistringstream = std::basic_istringstream<WCHAR>;
-#else
-using tstring = std::wstring;
-using tstringstream = std::wstringstream;
-using tistringstream = std::wistringstream;
-#endif
-using tofstream = std::basic_ofstream<WCHAR, std::char_traits<WCHAR>>;
-
 class StringUtils
 {
 public:
--- a/src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
+++ b/src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
@@ -15,7 +15,7 @@ CXmlDocWrapper::CXmlDocWrapper()
 
 }
 
-HRESULT CXmlDocWrapper::LoadContent(_In_ LPCWSTR wszValue, _In_ UINT cbValue)
+HRESULT CXmlDocWrapper::LoadContent(_In_ LPCWSTR wszValue)
 {
     HRESULT hr = S_OK;
 
@@ -35,9 +35,9 @@ HRESULT CXmlDocWrapper::LoadContent(_In_ LPCWSTR wszValue, _In_ UINT cbValue)
         return E_FAIL;
     }
 
-    IfFailLogNoRet(pDocument->put_async(VARIANT_FALSE));
-    IfFailLogNoRet(pDocument->put_validateOnParse(VARIANT_FALSE));
-    IfFailLogNoRet(pDocument->put_resolveExternals(VARIANT_TRUE));
+    IfFailLog(pDocument->put_async(VARIANT_FALSE));
+    IfFailLog(pDocument->put_validateOnParse(VARIANT_FALSE));
+    IfFailLog(pDocument->put_resolveExternals(VARIANT_TRUE));
 
     VARIANT_BOOL vbResult = VARIANT_TRUE;
     hr = pDocument->loadXML(CComBSTR(wszValue), &vbResult);
@@ -102,10 +102,10 @@ HRESULT CXmlDocWrapper::LoadContent(_In_ LPCWSTR wszValue, _In_ UINT cbValue)
     // TODO: willxie Test Linux change
     xmlDoc* pDocument = xmlReadMemory(
         CW2A(wszValue, CP_UTF8),
-        cbValue + 1,    // size of the buffer
-        "",             // the base URL to use for the document
-        NULL,           // document encoding
-        0               // xmlParserOption
+        wcslen_s(wszValue), // size of the buffer
+        "",                 // the base URL to use for the document
+        NULL,               // document encoding
+        0                   // xmlParserOption
     );
 
     IfNullRet(pDocument);
--- a/src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
+++ b/src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
@@ -3,6 +3,8 @@
 
 #include "stdafx.h"
 #include "XmlDocWrapper.h"
+#include "Encoding.h"
+#include "StringUtils.h"
 
 using namespace ATL;
 
@@ -99,13 +101,18 @@ HRESULT CXmlDocWrapper::LoadContent(_In_ LPCWSTR wszValue)
 #else
     LIBXML_TEST_VERSION
 
-    // TODO: willxie Test Linux change
+    CAutoVectorPtr<char> utf8Value;
+    CEncoding::ConvertUtf16ToUtf8(wszValue, utf8Value);
+
+    size_t utf8BufLen = 0;
+    IfFailRet(StringUtils::StringLen(utf8Value.m_p, utf8BufLen));
+
     xmlDoc* pDocument = xmlReadMemory(
-        CW2A(wszValue, CP_UTF8),
-        wcslen_s(wszValue), // size of the buffer
-        "",                 // the base URL to use for the document
-        NULL,               // document encoding
-        0                   // xmlParserOption
+        utf8Value.m_p,  // buffer
+        utf8BufLen,     // size of the buffer
+        "",             // the base URL to use for the document
+        NULL,           // document encoding
+        0               // xmlParserOption
     );
 
     IfNullRet(pDocument);
--- a/src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
+++ b/src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
@@ -1,0 +1,139 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "stdafx.h"
+#include "XmlDocWrapper.h"
+
+using namespace ATL;
+
+CXmlDocWrapper::CXmlDocWrapper()
+#ifdef PLATFORM_UNIX
+    : m_cleanupDoc(),
+    m_pDocument(nullptr)
+#endif
+{
+
+}
+
+HRESULT CXmlDocWrapper::LoadContent(_In_ LPCWSTR wszValue, _In_ UINT cbValue)
+{
+    HRESULT hr = S_OK;
+
+    if (m_pDocument != nullptr)
+    {
+        CLogging::LogError(_T("Attempted to load multiple xml files into a single document."));
+        return E_FAIL;
+    }
+
+#ifndef PLATFORM_UNIX
+    // IMPORTANT: Default configuration here is safe from XXE attacks by default in MSXML 6.0 (which we are using), but not MSXML 3.0!
+    CComPtr<IXMLDOMDocument2> pDocument;
+    hr = CoCreateInstance(CLSID_DOMDocument60, NULL, CLSCTX_INPROC_SERVER, IID_IXMLDOMDocument2, (void**)&pDocument);
+    if (FAILED(hr))
+    {
+        CLogging::LogError(_T("Failed to create xml document"));
+        return E_FAIL;
+    }
+
+    IfFailLogNoRet(pDocument->put_async(VARIANT_FALSE));
+    IfFailLogNoRet(pDocument->put_validateOnParse(VARIANT_FALSE));
+    IfFailLogNoRet(pDocument->put_resolveExternals(VARIANT_TRUE));
+
+    VARIANT_BOOL vbResult = VARIANT_TRUE;
+    hr = pDocument->loadXML(CComBSTR(wszValue), &vbResult);
+    if (FAILED(hr) || vbResult == VARIANT_FALSE)
+    {
+        CComPtr<IXMLDOMParseError> pError;
+        pDocument->get_parseError(&pError);
+
+        if (pError != nullptr)
+        {
+            CComBSTR bstrReason;
+            CComBSTR bstrSrcText;
+            long lineNumber = 0;
+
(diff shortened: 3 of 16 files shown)


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +420,178 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+
+                        CXmlNode* nextInstrumentationMethod = pInstrumentationMethodChildNode->Next();
+                        pInstrumentationMethodChildNode.Release();
+                        pInstrumentationMethodChildNode.Attach(nextInstrumentationMethod);
+                    }
+
+                    // TODO: validate
+                    pProfilerManager->m_configSources.push_back(pSource.p);
+                }
+                else if (wcscmp(bstrInstrumentationMethodNodeName, _T("RemoveInstrumentationMethod")) == 0)
```
jander-msft: Remove this for now

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -398,8 +410,9 @@ namespace MicrosoftInstrumentationEngine
         HRESULT AssemblyUnloadStartedImpl(_In_ AssemblyID assemblyId);
         HRESULT AssemblyUnloadFinishedImpl(_In_ AssemblyID assemblyId, _In_ HRESULT hrStatus);
 
-        HRESULT SetupInstrumentationMethods(
-            _In_ const std::vector<CComPtr<CConfigurationSource>>& configSources
+        HRESULT InvokeThreadRoutine(
+            _In_ const std::vector<CComPtr<CConfigurationSource>>& configSources,
```
jander-msft: Remove this parameter. If the caller is actually providing configuration sources, let it set m_configSources before invoking the thread routine.

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -252,12 +258,18 @@ namespace MicrosoftInstrumentationEngine
...
-        // lpParameter points to the this pointer so m_configFilePaths can be accessed. This is safe because the thread proc
+        // lpParameter points to the this pointer so m_configSources can be accessed. This is safe because the thread proc
         // blocks until this is complete.
         static DWORD WINAPI InstrumentationMethodThreadProc(
             _In_  LPVOID lpParameter
             );
 
+        // lpParameter points to the this pointer so m_wszConfigXml can be accessed. This is safe because the thread proc
+        // blocks until this is complete.
+        static DWORD WINAPI InitializeForAttachThreadProc(
```
jander-msft: I think ParseAttachConfigurationThreadProc better describes what this is method is doing.

src/Common.Lib/Macros.h
```diff
@@ -43,6 +49,11 @@
     do { if (FAILED(hr = (EXPR))) { return hr; } } while (false)
 #endif
 
+#ifndef IfFalseFail
```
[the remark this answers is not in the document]
WilliamXieMSFT: This was copied from ProductionBreakpoints. I'll use IfFalseRet

---
In reply to: [385965901](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385965901) [](ancestors = 385965901)

src/InstrumentationEngine.Lib/StringUtils.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include <string>
+
+#ifdef PLATFORM_UNIX
```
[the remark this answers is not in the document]
WilliamXieMSFT: Copy-pasted, I'll remove these.

---
In reply to: [385967831](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385967831) [](ancestors = 385967831)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2954,40 @@ HRESULT CProfilerManager::InitializeForAttach(
...
+    IfFailRet(InitializeCore(pCorProfilerInfoUnk, m_configSources));
 
     if (m_bProfilingDisabled)
     {
         return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
     }
 
-    IfFailRet(SendEventToRawProfilerCallback(&ICorProfilerCallback3::InitializeForAttach, pCorProfilerInfoUnk, pvClientData, cbClientData));
+    // TODO: willxie The ClientData for RawProfilerHook should be a subset of the ClientData parameter (which includes instrumentationmethods).
+    IfFailRet(SendEventToRawProfilerCallback(&ICorProfilerCallback3::InitializeForAttach, pCorProfilerInfoUnk, (void*)nullptr, 0U));
```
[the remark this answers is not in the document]
WilliamXieMSFT: Yep, otherwise it thinks nullptr is of type `std::nullptr_t`

---
In reply to: [385970264](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385970264) [](ancestors = 385970264)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +420,178 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+                LPWSTR wszLogLevel = nullptr;
+                while (pSettingNode != nullptr)
+                {
+                    CComBSTR bstrSettingName;
+                    pSettingNode->GetAttribute(_T("Name"), &bstrSettingName);
+
+                    CComBSTR bstrSettingValue;
+                    pSettingNode->GetAttribute(_T("Value"), &bstrSettingValue);
+
+                    ::SetEnvironmentVariable(bstrSettingName, bstrSettingValue);
```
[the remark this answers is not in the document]
WilliamXieMSFT: That's correct

---
In reply to: [385973615](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385973615) [](ancestors = 385973615)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +420,178 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+
+            CComBSTR bstrSettingsNodeName;
+            IfFailRet(pSettingsNode->GetName(&bstrSettingsNodeName));
+
+            if (wcscmp(bstrSettingsNodeName, _T("Settings")) == 0)
+            {
+                CComPtr<CXmlNode> pSettingNode;
+                pSettingsNode->GetChildNode(&pSettingNode);
+                LPWSTR wszLogLevel = nullptr;
+                while (pSettingNode != nullptr)
```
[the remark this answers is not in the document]
WilliamXieMSFT: Thanks for the catch

---
In reply to: [385973190](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385973190) [](ancestors = 385973190)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +420,178 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+                {
+                    CComBSTR bstrSettingName;
+                    pSettingNode->GetAttribute(_T("Name"), &bstrSettingName);
```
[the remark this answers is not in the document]
WilliamXieMSFT: IfFalseRet(bstrName.Length() != 0, E_FAIL)

---
In reply to: [385973370](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385973370) [](ancestors = 385973370)

WilliamXieMSFT:
I forgot GitHub shows each commit as an iteration in CodeFlow; I've squashed my 5 commits into 1 so it shows up correctly.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2954,40 @@ HRESULT CProfilerManager::InitializeForAttach(
...
+        cbClientData == 0)
+    {
+        // No configuration provided
+        return S_FALSE;
+    }
+
     vector<CComPtr<CConfigurationSource>> configSources;
 
-    // TODO: Parse configuration information from pvClientData.
+    UINT bufferSize = cbClientData + 1; // Hold data + null terminator
```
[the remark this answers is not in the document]
WilliamXieMSFT: Good catch, I do see the wsz length is half of cbClientData

---
In reply to: [385969792](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385969792) [](ancestors = 385969792)

src/InstrumentationEngine/ProfilerManager.cpp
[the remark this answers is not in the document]
WilliamXieMSFT: I will only allow the three values we've discussed (LogLevel, FileLogLevel, FileLogPath). For now, the FileLoggerSink reads the environment so I will need to set the latter 2 as an environment variable. I can create a task to address the FileLoggerSink dependency on environment variables.

---
In reply to: [385973711](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385973711) [](ancestors = 385973711)

src/InstrumentationEngine/ProfilerManager.cpp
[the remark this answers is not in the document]
WilliamXieMSFT: I'll use an unordered_map to ensure unique keys for settings. Handling duplicates by the last element set wins.

---
In reply to: [385974026](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r385974026) [](ancestors = 385974026)

src/InstrumentationEngine/ProfilerManager.cpp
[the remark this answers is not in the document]
jander-msft: That task already exists: #176

src/InstrumentationEngine/ProfilerManager.cpp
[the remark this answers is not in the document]
WilliamXieMSFT: Handled in ParseSettingsConfigurationNode

---
In reply to: [386678893](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386678893) [](ancestors = 386678893,385974026)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,203 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+                        ::SetEnvironmentVariable(_T("MicrosoftInstrumentationEngine_FileLogPath"), it->second.c_str());
+                    }
+                    
+                }
+            }
+        }
+        else if (wcscmp(bstrCurrChildNodeName, _T("InstrumentationMethods")) == 0)
+        {
+            CComPtr<CXmlNode> pInstrumentationMethodNode;
+            IfFailRet(pCurrChildNode->GetChildNode(&pInstrumentationMethodNode));
```
jander-msft: >GetChildNode [](start = 38, length = 12)

Can you check if GetChildNode returns an error if there are no child nodes? If so, we don't want to fail the parsing, but just don't try to process on the nonexistent node.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,203 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+
+            while (pInstrumentationMethodNode != nullptr)
+            {
+                CComBSTR bstrInstrumentationMethodNodeName;
+                IfFailRet(pInstrumentationMethodNode->GetName(&bstrInstrumentationMethodNodeName));
+
+                if (wcscmp(bstrInstrumentationMethodNodeName, _T("AddInstrumentationMethod")) == 0)
+                {
+                    CComBSTR bstrConfigPath;
+                    IfFailRet(pInstrumentationMethodNode->GetAttribute(_T("ConfigPath"), &bstrConfigPath));
```
jander-msft: >GetAttribute [](start = 58, length = 12)

Can you check if GetAttribute returns an error if the attribute doesn't exist? Even if it doesn't, it's probably good to verify that bstrConfigPath is not null/empty (check it's length).

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,203 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+
+                CXmlNode* next = pInstrumentationMethodNode->Next();
+                pInstrumentationMethodNode.Release();
+                pInstrumentationMethodNode.Attach(next);
+            }
+        }
+        else if (wcscmp(bstrCurrChildNodeName, _T("Profilers")) == 0)
+        {
+            // TODO: handle raw profiler
+        }
```
jander-msft: I would suggest just removing this. We might not get to this at this time. We probably do want to support it, but don't know when we'll get to it.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,203 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+{
+    HRESULT hr = S_OK;
+    CComPtr<CXmlNode> pSettingNode;
+    IfFailRet(parentNode->GetChildNode(&pSettingNode));
+    while (pSettingNode != nullptr)
+    {
+        CComBSTR bstrSettingNodeName;
+        IfFailRet(pSettingNode->GetName(&bstrSettingNodeName));
+
+        if (wcscmp(bstrSettingNodeName, _T("Setting")) != 0)
```
jander-msft: >if (wcscmp(bstrSettingNodeName, _T("Setting")) != 0) [](start = 8, length = 52)

I think we should just fail if it's not a Setting node. You can also check that it is not a comment node (but that's secondary).

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,203 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+
+        if (settings.find(bstrSettingName.m_str) == settings.end())
+        {
+            settings.insert(std::pair<tstring, tstring>(bstrSettingName, bstrSettingValue));
+        }
+        else
+        {
+            // Override value.
+            settings[bstrSettingName.m_str] = bstrSettingValue;
+        }
```
jander-msft: Is there any value in allowing a setting to be declared more than one time? I think we should start out by failing if a setting appears more than once. We can relax that later if it's somehow useful to allow for this to occur.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2978,47 @@ HRESULT CProfilerManager::InitializeForAttach(
...
+    {
+        // No configuration provided
+        return S_FALSE;
+    }
+
     vector<CComPtr<CConfigurationSource>> configSources;
 
-    // TODO: Parse configuration information from pvClientData.
+    // WCHAR = 2 bytes, hence we halve the byte count.
+    UINT charCount = cbClientData / 2;
```
jander-msft: >2 [](start = 36, length = 1)

sizeof(WCHAR) or (sizeof(WCHAR) / sizeof(byte)) if you really want to get technical.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2978,47 @@ HRESULT CProfilerManager::InitializeForAttach(
...
+    UINT bufferSize = charCount + 1; // add room for null terminator.
+    unique_ptr<WCHAR[]> wszConfigXml = make_unique<WCHAR[]>(bufferSize);
+
+    LPCWSTR pData = reinterpret_cast<LPCWSTR>(pvClientData);
+    IfFailRetErrno(
+        memcpy_s(
+            wszConfigXml.get(), // Destination buffer
+            bufferSize * 2,     // Destination buffer size in Bytes
```
jander-msft: >2 [](start = 25, length = 1)

Same thing about using sizeof(WCHAR)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: I was mirroring the InstrumentationEngine logic which does allow duplicates due to parsing in a loop, but I agree with your point.

---
In reply to: [386707618](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386707618) [](ancestors = 386707618)

src/InstrumentationEngine/ProfilerManager.cpp
jander-msft: Probably could just make this a constant, something like: const size_t WCharSizeInBytes = sizeof(WCHAR) / sizeof(byte);

---
In reply to: [386708337](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386708337) [](ancestors = 386708337)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2978,47 @@ HRESULT CProfilerManager::InitializeForAttach(
...
+            pData,              // Source buffer
+            cbClientData        // Source buffer size in Bytes
+        ));
+
+    m_wszConfigXml = wszConfigXml.release();
+    m_configSources = configSources;
```
jander-msft: >m_configSources = configSources; [](start = 4, length = 32)

This shouldn't be necessary m_configSources should already be an empty vector. Copying from another empty vector should not be necessary.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2978,47 @@ HRESULT CProfilerManager::InitializeForAttach(
...
     m_bAttach = true;
 
+    if (pvClientData == nullptr ||
+        cbClientData == 0)
+    {
+        // No configuration provided
+        return S_FALSE;
+    }
+
     vector<CComPtr<CConfigurationSource>> configSources;
```
jander-msft: >vector<CComPtr<CConfigurationSource>> configSources; [](start = 4, length = 52)

I think this can be removed.

jander-msft:
>     _In_ const vector<CComPtr<CConfigurationSource>>& configSources

Consider removing configSources parameter and have callers just set m_configSources. In the attach scenario, the InitializeForAttach method is calling this method by passing in m_configSources. I don't know if this behavior is defined or if it is implementation dependent; just better to avoid it.

---
Refers to: src/InstrumentationEngine/ProfilerManager.cpp:1055 in 5ddfdf7. [](commit_id = 5ddfdf744a8ddb5f76919c0d64b3ad120ca5222f, deletion_comment = False)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: GetChildNode will return a nullptr, good point on catching this

---
In reply to: [386705329](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386705329) [](ancestors = 386705329)

WilliamXieMSFT:
>     _In_ const vector<CComPtr<CConfigurationSource>>& configSources

Sure, I'll move things to use the private member instead.

---
In reply to: [593674176](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#issuecomment-593674176) [](ancestors = 593674176)

---
Refers to: src/InstrumentationEngine/ProfilerManager.cpp:1055 in 5ddfdf7. [](commit_id = 5ddfdf744a8ddb5f76919c0d64b3ad120ca5222f, deletion_comment = False)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Removed the else statement

---
In reply to: [386708529](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386708529) [](ancestors = 386708529,386707618)

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -244,20 +251,29 @@ namespace MicrosoftInstrumentationEngine
...
             _In_  LPVOID lpParameter
             );
 
+        // lpParameter points to the this pointer so m_wszConfigXml can be accessed. This is safe because the thread proc
+        // blocks until this is complete.
+        static DWORD WINAPI ParseAttachConfigurationThreadProc(
+            _In_ LPVOID lpParameter
+            );
+
+        static HRESULT ParseSettingsConfigurationNode(
```
jander-msft: >ParseSettingsConfigurationNode [](start = 23, length = 30)

Maybe add a brief comment with a small example of what this method expects the XML node to look like.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2781,21 +2963,43 @@ HRESULT CProfilerManager::InitializeForAttach(
 
     m_bAttach = true;
 
-    vector<CComPtr<CConfigurationSource>> configSources;
+    if (pvClientData == nullptr ||
+        cbClientData == 0)
+    {
+        // No configuration provided
+        return S_FALSE;
```
jander-msft: >S_FALSE [](start = 15, length = 7)

Should this be considered a failure? I think if you return S_FALSE, the CLR is still going to initialize CLRIE and keep it attached. Might cause issues down the line if we think that the engine is working properly when we inspect a process that already has a profiler attached but no instrumentation methods.

WilliamXieMSFT: Makes sense. We don't want CLRIE to be attached when it can't support adding additional IM's yet. I'll return E_FAIL.

---
In reply to: [386720873](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386720873) [](ancestors = 386720873)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,192 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+                        IfFailRet(pInstrumentationMethodChildNode->GetName(&bstrInstrumentationMethodChildNodeName));
+
+                        if (wcscmp(bstrInstrumentationMethodChildNodeName, _T("Settings")) == 0)
+                        {
+                            unordered_map<tstring, tstring> settingsMap;
+                            IfFailRet(ParseSettingsConfigurationNode(pInstrumentationMethodChildNode, settingsMap));
+
+                            for (unordered_map<tstring, tstring>::iterator it = settingsMap.begin();
+                                 it != settingsMap.end();
+                                 it++)
```
delmyers: >it++ [](start = 33, length = 4)

nit, typically ++it is used with iterators, so that a temporary variable doesn't have to be made.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -419,6 +418,192 @@ DWORD WINAPI CProfilerManager::InstrumentationMethodThreadProc(
...
+                            {
+                                IfFailRet(pSource->AddSetting(it->first.c_str(), it->second.c_str()));
+                            }
+                        }
+
+                        CXmlNode* nextInstrumentationMethod = pInstrumentationMethodChildNode->Next();
+                        pInstrumentationMethodChildNode.Release();
+                        pInstrumentationMethodChildNode.Attach(nextInstrumentationMethod);
```
delmyers: >Attach [](start = 56, length = 6)

Attach doesn't increment the ref count. Is that what you want?

WilliamXieMSFT: I believe so. This is the pattern used in ProductionBreakpoints for iterating child nodes.

---
In reply to: [386727155](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386727155) [](ancestors = 386727155)

delmyers: It's weird that you release them after their use, then. It seems that this means the document will point to a bunch of invalid nodes.

---
In reply to: [386727621](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386727621) [](ancestors = 386727621,386727155)

delmyers: Would you mind double-checking the reference counting on these via the debugger? It seems like this could cause mysterious AVs

---
In reply to: [386728314](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386728314) [](ancestors = 386728314,386727621,386727155)

WilliamXieMSFT: That's because Next() news up a new node each time. CXmlNode is an abstraction over the underlying implementation (libxml2 or msxml6) which does the actual navigation to get the next node in the document. This is just throwing away the wrapper once done.

---
In reply to: [386728314](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386728314) [](ancestors = 386728314,386727621,386727155)

WilliamXieMSFT: no problem, I'll double-check

---
In reply to: [386729381](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386729381) [](ancestors = 386729381,386728314,386727621,386727155)

delmyers: I see. That makes sense, then. The normal root for this would be an HRESULT Node::Next(_Out_ Node** pNext), so as to not cause confusion. 

---
In reply to: [386729418](https://github.com/microsoft/CLRInstrumentationEngine/pull/216#discussion_r386729418) [](ancestors = 386729418,386728314,386727621,386727155)

src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        return E_FAIL;
+    }
+#else
+    LIBXML_TEST_VERSION
+
+    // TODO: willxie Test Linux change
+    xmlDoc* pDocument = xmlReadMemory(
+        CW2A(wszValue, CP_UTF8),
```
jander-msft: Try using CEncoding::ConvertUtf16ToUtf8 here instead.

src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
```diff
@@ -0,0 +1,139 @@
...
+        wcslen_s(wszValue), // size of the buffer
```
jander-msft: Try using StringUtils::StringLen to get the length and then pass that in. I think the Linux build issue is that wcslen_s is not defined.

src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
WilliamXieMSFT: Thanks, that fixed it!

src/InstrumentationEngine.Lib/XmlDocWrapper.cpp
WilliamXieMSFT: Thanks!

PR #217, opened 2020-03-03
Migrate static analysis builds to yaml

WilliamXieMSFT:
Note: this doesn't include PREFast yet, but I aim to make it similar to our Aerobee repo


build/yaml/jobs/codeanalysis/apiscan.yaml
```diff
@@ -0,0 +1,57 @@
...
+      downloadPath: $(Build.ArtifactStagingDirectory)
+
+  # Run API Scan
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-apiscan.APIScan@1
+    displayName: 'Run APIScan'
+    inputs:
+      softwareFolder: '$(Build.ArtifactStagingDirectory)/binaries-windows-Release'
+      softwareName: Aerobee
+      softwareVersionNum: '$(Build.BuildNumber)'
+      symbolsFolder: 'SRV*http://symweb;$(Build.ArtifactStagingDirectory)/binaries-windows-Release'
```
delmyers: >http://symweb [](start = 26, length = 13)

This might itself fail scans because it doesn't use https. It is an internal Microsoft website, only available in our network. Maybe consider making this a build variable?

WilliamXieMSFT: Static analysis builds will only be run on internal VSEng-MicroBuildVS2019 pool since they rely on internal tasks. 

---
In reply to: [387305287](https://github.com/microsoft/CLRInstrumentationEngine/pull/217#discussion_r387305287) [](ancestors = 387305287)

WilliamXieMSFT: I can move the URL to https

---
In reply to: [387309275](https://github.com/microsoft/CLRInstrumentationEngine/pull/217#discussion_r387309275) [](ancestors = 387309275,387305287)

WilliamXieMSFT: Https results in ERROR: System.ArgumentException: One or more symbol path parts are incorrect. I'll leave it as Http for now

PR #219, opened 2020-03-03
Add environment variable for path to extension root.

jander-msft:
Add environment variable to allow discovering of root of the site extension. This is useful for if the site extension is already enabled (either via app setting for the preinstall site extension or via upload as a private site extension) to allow other processes to discover which engine version should be used.

### DIFF
--- a/src/InstrumentationEngine.Preinstall/scmApplicationHost.xdt
+++ b/src/InstrumentationEngine.Preinstall/scmApplicationHost.xdt
@@ -21,7 +21,7 @@
   <system.webServer>
     <runtime xdt:Transform="InsertIfMissing">
       <environmentVariables xdt:Transform="InsertIfMissing">
-        <add name="MicrosoftInstrumentationEngine_AttachPath" value="%XDT_EXTENSIONPATH%\Tools\Attach\Microsoft.InstrumentationEngine.Attach.exe" xdt:Locator="Match(name)" xdt:Transform="InsertIfMissing" />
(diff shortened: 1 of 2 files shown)


src/InstrumentationEngine.Preinstall/scmApplicationHost.xdt
```diff
@@ -21,6 +21,7 @@
   <system.webServer>
     <runtime xdt:Transform="InsertIfMissing">
       <environmentVariables xdt:Transform="InsertIfMissing">
+        <add name="MicrosoftInstrumentationEngine_AttachPath" value="%XDT_EXTENSIONPATH%\Tools\Attach\Microsoft.InstrumentationEngine.Attach.exe" xdt:Locator="Match(name)" xdt:Transform="InsertIfMissing" />
```
wiktork: >MicrosoftInstrumentationEngine_AttachPath [](start = 19, length = 41)

So if preinstalled site extension is already enabled, the http handler won't have to work out the folder location but rather use the env variable correct?

jander-msft: Correct. I might change this to \be the root of the site extension instead of the full path to the executable depending on how this information is passed to the various executables so they can pull out the various paths under the site extension (the attach exe, the engine assemblies, etc).

jander-msft: I changed the variable to MicrosoftInstrumentationEngine_InstallationRoot and it has a value of %XDT_EXTENSIONPATH%

PR #224, opened 2020-03-06
Convert attach data to UTF16 before consuming it.

jander-msft:
- Convert attach data to UTF16 before consuming it.
- Fix configuration XML memory leak.

### DIFF
--- a/src/InstrumentationEngine/ProfilerManager.h
+++ b/src/InstrumentationEngine/ProfilerManager.h
@@ -82,7 +82,7 @@ namespace MicrosoftInstrumentationEngine
         vector<CComPtr<CConfigurationSource>> m_configSources;
 
         // The configuration xml for profiler attach.
-        LPCWSTR m_wszConfigXml;
+        tstring m_tstrConfigXml;
 
         // list of loaded instrumentation method guids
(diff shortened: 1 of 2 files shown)


WilliamXieMSFT:
>     const size_t WCharSizeInBytes = sizeof(WCHAR) / sizeof(BYTE);

Not used anymore

---
Refers to: src/InstrumentationEngine/ProfilerManager.h:25 in dc539b5. [](commit_id = dc539b53f245250964243fa7cf983b3648e9209f, deletion_comment = False)

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -82,7 +82,7 @@ namespace MicrosoftInstrumentationEngine
         vector<CComPtr<CConfigurationSource>> m_configSources;
 
         // The configuration xml for profiler attach.
-        LPCWSTR m_wszConfigXml;
```
WilliamXieMSFT: >LPCWSTR [](start = 8, length = 7)

Was this what caused the memory leak?

jander-msft: Yes

PR #225, opened 2020-03-07
Generate private site extension XDT for attach

jander-msft:
The private site extension XDT generation creates an XDT file that applies the engine and profiler environment variables automatically when the site extension is applied. For testing the attach scenario, we need to have an XDT that removes the environment variables in order to mimic the production scenario where the InstrumentationEngine preinstalled site extension is not enabled.

### DIFF
--- a/build/Scripts/PrivateXdt/GeneratePrivateXdt.ps1
+++ b/build/Scripts/PrivateXdt/GeneratePrivateXdt.ps1
@@ -27,7 +27,11 @@ param(
 
     [Parameter(Mandatory=$false)]
     [Switch]
-    $DisableSignatureValidation
+    $DisableSignatureValidation,
+
+    [Parameter(Mandatory=$false)]
+    [Switch]
+    $ForAttach
 )
 
 Import-Module "$PSScriptRoot\XdtHelper.psm1" -Force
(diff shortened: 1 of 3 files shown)


WilliamXieMSFT:
I'm fine with this change but wouldn't we just ignore enabling the preinstalled site extension all-together when testing? Or is this so we can avoid updating the deploytestapps release pipeline we have?

jander-msft:
We're going to need "attach" versions of the XDT files to enable using of the bits in the private site extension without setting the profiler environment variables. What I'm thinking is that we'll need to create another private site extension (call it InstrumentationEngineAttach) that is applied when we want to test new bits, along with the InstrumentationEngine private site extension. It will override everything from the InstrumentationEngine private site extension to remove all of its variables and then set the ones it needs for attach (namely MicrosoftInstrumentationEngine_DisableCodeSignatureValidation in the applicationHost.xdt and MicrosoftInstrumentationEngine_InstallationRoot in scmApplicationHost.xdt). When VS starts the session, it will be working against the bits in the InstrumentationEngineAttach private site extension. If attach fails, and it needs to set the application settings to use the profiler environment variables, VS needs to delete the InstrumentationEngineAttach private site extension (or somehow disable it), apply the application settings, and restart the site.

This is my current thinking, which isn't fully reflected in this PR. Let's talk about this more tomorrow.

---
In reply to: [596266862](https://github.com/microsoft/clrinstrumentationengine/pull/225#issuecomment-596266862) [](ancestors = 596266862)

WilliamXieMSFT:
Thanks for the write-up, right we'll need the private extension to have all of the "debug" environment variables and avoid profiling on startup.

---
In reply to: [596290316](https://github.com/microsoft/CLRInstrumentationEngine/pull/225#issuecomment-596290316) [](ancestors = 596290316,596266862)

PR #230, opened 2020-03-11
Resolve path Settings in ConfigurationSources

WilliamXieMSFT:


src/InstrumentationEngine.Attach/Program.cs
```diff
@@ -316,11 +316,25 @@ private static void WriteError(string message)
                                 {
                                     foreach (var setting in source.Settings)
                                     {
-                                        methodSettings.Add(new SettingsTypeSetting()
+                                        if (setting.IsPathSpecified &&
```
jander-msft: nit: Maybe have the if-else assign to a "string settingValue" local and then only write the "methodsSettings.Add(...)" code once.

PR #231, opened 2020-03-12
Add parsing Instrumentation Method loglevel from ConfigurationSource settings

WilliamXieMSFT:


src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -16,6 +16,19 @@
 
 using namespace ATL;
 
+// The following fixes Error C2280 'std::_Uhash_compare<_Kty,_Hasher,_Keyeq>::_Uhash_compare(const std::_Uhash_compare<_Kty,_Hasher,_Keyeq> &)': attempting to reference a deleted function
+// which occurs when compiling using GUID key for std::unordered_map
+namespace std {
+    template<> struct hash<GUID>
+    {
```
wiktork: Do we have a common we could put this into? It seems useful in general.

src/InstrumentationEngine/ProfilerManagerForInstrumentationMethod.cpp
```diff
@@ -29,7 +29,13 @@ CProfilerManagerForInstrumentationMethod::CProfilerManagerForInstrumentationMeth
...
-        if (GetEnvironmentVariable(wszInstrumentationMethodLogLevelEnvVar.c_str(), wszEnvVar, MAX_PATH) > 0)
+        LoggingFlags flags;
+
+        if (SUCCEEDED(pProfilerManager->GetInstrumentationMethodLogLevel(instrumentationMethodGuid, &flags)))
+        {
+            m_flags = flags;
+        }
+        else if (GetEnvironmentVariable(wszInstrumentationMethodLogLevelEnvVar.c_str(), wszEnvVar, MAX_PATH) > 0)
         {
             m_flags = CLoggerService::ExtractLoggingFlags(wszEnvVar);
```
wiktork: Is Error our ultimately default, even if nothing is specified?

src/InstrumentationEngine/ProfilerManagerForInstrumentationMethod.cpp
```diff
@@ -29,7 +29,13 @@ CProfilerManagerForInstrumentationMethod::CProfilerManagerForInstrumentationMeth
         // Get InstrumentationMethod-specific LogLevel Environment Variable
         tstring wszInstrumentationMethodLogLevelEnvVar = _T("MicrosoftInstrumentationEngine_LogLevel_") + m_wszInstrumentationMethodGuid;
         WCHAR wszEnvVar[MAX_PATH];
```
jander-msft: Could you refactor this so that we only look up the environment variable if the engine doesn't know the specific value? e.g.

if (get from profiler manager) {
}
else
{
  lookup from env var
  if (env exists)
  {
    get from env var
  }
  else
  {
    get from global
  }
}

src/InstrumentationEngine/ProfilerManagerForInstrumentationMethod.cpp
jander-msft: I believe that None is the default, which I think is the appropriate default value.

---
In reply to: [399476438](https://github.com/microsoft/CLRInstrumentationEngine/pull/231#discussion_r399476438) [](ancestors = 399476438)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -537,6 +537,32 @@ DWORD WINAPI CProfilerManager::ParseAttachConfigurationThreadProc(
                                  it != settingsMap.end();
                                  ++it)
                             {
+                                // Assume format is:
+                                // MicrosoftInstrumentationEngine_LogLevel_12341234-1234-1234-1234-123412341234
+                                if (wcsncmp(it->first.c_str(), _T("MicrosoftInstrumentationEngine_LogLevel_"), 40) == 0)
```
jander-msft: >if (wcsncmp(it->first.c_str(), _T("MicrosoftInstrumentationEngine_LogLevel_"), 40) == 0) [](start = 32, length = 88)

I'm thinking we might want to provide this via a different element/attribute in the XML description. It could easily be just an attribute:

<AddInstrumentationMethod ConfigPath="..." LogLevel="Errors">

or if we want to leave it open for further expansion, add another element child of AddInstrumentationMethod to specify values meant to be consumed by the engine (rather than the method) but describe method configuration:

<AddInstrumentationMethod ConfigPath="...">
  <Configuration>
    <LogLevel>Errors</LogLevel>
    ...
  </Configuration>
  <Settings>
    ...
  </Settings>
</AddInstrumentationMethod>

jander-msft: In either of these cases, we don't have to search for patterns of elements, just for exact names, and we don't force IM producers to set additional settings that they'll see during initialization that they don't really need to consume.

---
In reply to: [400363077](https://github.com/microsoft/CLRInstrumentationEngine/pull/231#discussion_r400363077) [](ancestors = 400363077)

src/InstrumentationEngine/ProfilerManager.h
WilliamXieMSFT: I can move this to common.lib's stdafx.h

---
In reply to: [399475650](https://github.com/microsoft/CLRInstrumentationEngine/pull/231#discussion_r399475650) [](ancestors = 399475650)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: I'll add the following to the ConfigurationSources schema:

<InstrumentationConfigurationSource>
 <Platforms>
  ...
  </Platforms>
  <Configuration>
    <LogLevel>Errors Dumps</LogLevel>
  </Configuration>
  <Settings>
  ...
  </Settings>
</InstrumentationConfigurationSource>

This would make it an easy mapping to either an attribute on or <Configuration /> section of <AddInstrumentationMethod />

---
In reply to: [400363975](https://github.com/microsoft/CLRInstrumentationEngine/pull/231#discussion_r400363975) [](ancestors = 400363975,400363077)

src/InstrumentationEngine/InstrumentationEngine.vcxproj
```diff
@@ -148,7 +148,7 @@
     <ClCompile Include="TokenType.cpp" />
     <ClCompile Include="Type.cpp" />
     <ClCompile Include="TypeCreator.cpp" />
-    <ClInclude Include="ProfilerManagerForInstrumentationMethod.h" />
+    <ClInclude Include="*.h" />
```
wiktork, line 151: Intended?

WilliamXieMSFT: Yep, it originally was *.h, and when I added my ProfilerManagerForInstrumentationMethod.h file in VS it overwrote this so VS stopped displaying all of the other .h files. 

---
In reply to: [405691979](https://github.com/microsoft/CLRInstrumentationEngine/pull/231#discussion_r405691979) [](ancestors = 405691979)

WilliamXieMSFT: MSBuild is now generating a warning about this, I'll revert *.h to explicit files.

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: I'll need to get the MethodGuid somehow in order to track the correct loglevel to method. I'm thinking about adding this logic to the attach handler when parsing the configurationsource.

---
In reply to: [405687799](https://github.com/microsoft/CLRInstrumentationEngine/pull/231#discussion_r405687799) [](ancestors = 405687799,400363975,400363077)

WilliamXieMSFT:
Closing this PR since it's low priority and has not moved for years.

PR #236, opened 2020-03-25
Test helper site extension for determining location of private site extension.

jander-msft:
- Test helper site extension for determining location of private site extension.
- Remove specialized attach XDT file generation.


WilliamXieMSFT:
If we check this in now it will disable CSI's ability to test since the changes to enable extension.xml doesn't yet exist.

src/InstrumentationEngine.SiteExtensionTestHelper/extension.xml
```diff
@@ -0,0 +1,3 @@
+<extension>
```
jander-msft: Remove this file as it is not needed.

WilliamXieMSFT:
Nvm, I had assumed the extension.xml file would sit in the site extension folder.

---
In reply to: [603964400](https://github.com/microsoft/CLRInstrumentationEngine/pull/236#issuecomment-603964400) [](ancestors = 603964400)

src/InstrumentationEngine.SiteExtensionTestHelper/scmApplicationHost.xdt
```diff
@@ -0,0 +1,32 @@
...
+                provides system.webServer/runtime/environmentVaraibles. This will cause the overriding behavior for kudu environment variables.
+  -->
+  <system.webServer>
+    <runtime xdt:Transform="InsertIfMissing">
+      <environmentVariables xdt:Transform="InsertIfMissing">
+        <!-- Set the location of the InstrumentationEngine site extension just in case the private site extension is not enabled. This will allow testing to more easily
+             locate the InstrumentationEngine site extension files without having to resort to complicated logic of determining if it is preinstalled vs private, which
+             version to use, etc
+        -->
+        <add name="MicrosoftInstrumentationEngine_InstallationRoot" value="%HOME%\SiteExtensions\InstrumentationEngine" xdt:Locator="Match(name)" xdt:Transform="InsertIfMissing" />
```
wiktork: >MicrosoftInstrumentationEngine_InstallationRoot [](start = 19, length = 47)

I'm not opposed to do this, but wouldn't it be simpler to just set an AppSetting with this value instead of deploying an entire site extension in test scenarios?

jander-msft: Good call; I was thinking we could use this site extension for anything else we needed to add just for testing purposes, but this would simplify it. I'll remove this site extension and use the app setting instead and we can revisit using the test site extension later if we need to add additional functionality.

PR #239, opened 2020-03-31
Fix signing files used by Profiler Attach

WilliamXieMSFT:
Add the following files to be signed:
* Microsoft.InstrumentationEngine.Attach.exe

### DIFF
--- a/build/CodeSign/certificate.ignore
+++ b/build/CodeSign/certificate.ignore
@@ -16,5 +16,3 @@ bin\*\*.wixpdb, Ignore wix artifact
 bin\*\*.msidir\*.cab, Ignore embedded cab files
 
 binaries-windows-release\*, Ignore downloaded binary artifacts
-
-bin\*\xdtextensions.dll, Ignore test assembly
--- a/build/CodeSign/certificate.ignore
+++ b/build/CodeSign/certificate.ignore
@@ -16,3 +16,5 @@ bin\*\*.wixpdb, Ignore wix artifact
 bin\*\*.msidir\*.cab, Ignore embedded cab files
 
 binaries-windows-release\*, Ignore downloaded binary artifacts
+
+bin\*\xdtextensions.dll, Ignore test assembly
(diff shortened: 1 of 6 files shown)


src/InstrumentationEngine.Attach/InstrumentationEngine.Attach.csproj
```diff
@@ -46,4 +47,8 @@
     <Error Condition="'%(XsdSchemaFile.Namespace)' == ''" Text="Missing Namespace attribute for XsdSchemaFile %(XsdSchemaFile.FullPath)" />
     <Exec Command="&quot;$(SDK40ToolsPath)\xsd.exe&quot; &quot;%(XsdSchemaFile.FullPath)&quot; /classes /namespace:%(XsdSchemaFile.Namespace) /language:%(XsdSchemaFile.Language) /out:$(MSBuildThisFileDirectory)" />
   </Target>
+  <ItemGroup>
+    <!-- dotnet core 3.1 produces both a dll and exe; project references only pick up the dll so this explicitly adds the exe for signing -->
+    <FilesToSign Include="$(OutputPath)\Microsoft.InstrumentationEngine.Attach.exe"/>
```
wiktork, line 52: This comment is confusing. I would expect the Microbuild targets file to sign the exe output as well. Whether other projects have references to this one seems orthogonal.

WilliamXieMSFT: Ah, the comment was copied from Aerobee which uses a signproj with a projectreference. I had assumed this was what caused the dll to get signed and not the exe, hence the comment.

build/CodeSign/certificate.ignore
```diff
@@ -15,4 +15,6 @@ bin\*\*.wixlib, Ignore wix artifact
 bin\*\*.wixpdb, Ignore wix artifact
 bin\*\*.msidir\*.cab, Ignore embedded cab files
 
-binaries-windows-release\*, Ignore downloaded binary artifacts
\ No newline at end of file
+binaries-windows-release\*, Ignore downloaded binary artifacts
```
jander-msft, line 18: I think you can revert all of the exclusions if my XDT extensions signing PR is completed.

PR #241, opened 2020-04-01
Fix xdt

WilliamXieMSFT:


build/Scripts/PrivateXdt/XdtHelper.psm1
```diff
@@ -32,19 +32,38 @@ function Get-EnvironmentVariableAddInfo {
...
     # <add name="$EnvironmentVariable" xdt:Locator="Match(name)" xdt:Transform="$Transform" />
-    Write-Output @{
-        destination = '/configuration/system.webServer/runtime/environmentVariables'
-        operation = $Operation
-        element = 'add'
-        attributes = [ordered] @{
-            name = $EnvironmentVariable
-            'xdt:Locator' = 'Match(name)'
-            'xdt:Transform' = $Transform
+    if ($Value) {
```
jander-msft, line 45: Rather than duplicating the object creation, why not create the object and then conditionally set the value property on the attributes object?

PR #243, opened 2020-04-02
Make nested types report fully qualified names.

### LINKED ISSUE #141: IMethodInfo::GetFullName does not return the full name of async state machine methods
I found this while debugging my instrumentation method. I was trying to match the name of an async state machine method, and it seems that ::GetFullName() is just returning the typename of the inner type. It is not encapsulating it in the outer type and namespace. The way that ILDASM does this is:

```<namespace>.<TypeName>+<InnerType>.MethodName```

### PR DESCRIPTION
**Pull request by delmyers**
Updates CTokenType to give fully qualified names of typedefs, including nested types.

Before this change, nested types only reported the immediate type in which they were created. This change adds the nested types separated by '/' characters in the same way that ILDASM does.

This change is required by the fakes team because they need to be able to match full names of methods, especially for async code.

fixes #141

### DIFF
--- a/src/InstrumentationEngine/TokenType.cpp
+++ b/src/InstrumentationEngine/TokenType.cpp
@@ -40,14 +40,44 @@ HRESULT MicrosoftInstrumentationEngine::CTokenType::GetName(_Out_ BSTR* pbstrNam
 
         if (TypeFromToken(m_token) == mdtTypeDef)
         {
+            mdTypeDef tkEnclosing = mdTokenNil;
+            mdTypeDef tkCurrent = m_token;
+            tstring fullName;
+            vector<WCHAR> nameBuffer(100);
             ULONG cchLength = 0;
+            while ((pImport->GetNestedClassProps(tkCurrent, &tkEnclosing) == S_OK) && (tkEnclosing != mdTokenNil))
+            {
+                IfFailRet(pImport->GetTypeDefProps(tkEnclosing, nullptr, 0, &cchLength, nullptr, nullptr));
+                if (nameBuffer.size() < cchLength)
+                {
+                    nameBuffer.resize(cchLength);
+                }
+
+                // Add nested class names separated by a "/".
+                IfFailRet(pImport->GetTypeDefProps(tkEnclosing, nameBuffer.data(), cchLength, &cchLength, nullptr, nullptr));
+                fullName.insert(0, _T("/"));
+                fullName.insert(0, nameBuffer.data());
+
+                tkCurrent = tkEnclosing;
+            }
+
             IfFailRet(pImport->GetTypeDefProps(m_token, nullptr, 0, &cchLength, nullptr, nullptr));
-            std::vector<WCHAR> nameBuffer(cchLength);
+            if (nameBuffer.size() < cchLength)
+            {
+                nameBuffer.resize(cchLength);
+            }
+
             IfFailRet(pImport->GetTypeDefProps(m_token, nameBuffer.data(), cchLength, &cchLength, nullptr, nullptr));
-            m_name = CComBSTR(cchLength, nameBuffer.data());
+            fullName += nameBuffer.data();
+
+            m_name = CComBSTR(static_cast<ULONG>(fullName.length()+1), fullName.c_str());
         }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/InstrumentationEngine/TokenType.cpp [hunk 1]
```diff
@@ -1,69 +1,99 @@
...
+            {
+                IfFailRet(pImport->GetTypeDefProps(tkEnclosing, nullptr, 0, &cchLength, nullptr, nullptr));
+                if (nameBuffer.size() < cchLength)
+                {
+                    nameBuffer.resize(cchLength);
+                }
+
+                // Add nested class names separated by a "/".
+                IfFailRet(pImport->GetTypeDefProps(tkEnclosing, nameBuffer.data(), cchLength, &cchLength, nullptr, nullptr));
+                fullName.insert(0, _T("/"));
```
clguimanMSFT, line 58: why do you use '/' instead of '+'?

delmyers: That's just the way that ILDASM does it now. I'm fine with either one.

AbhitejJohn: It'd be great to have `+` here so this aligns with the output that reflection and metadata based tools including roslyn return.

src/InstrumentationEngine/TokenType.cpp
```diff
@@ -1,69 +1,99 @@
...
+    {
+        CComPtr<IMetaDataImport> pImport;
+        IfFailRet(m_pOwningModule->GetMetaDataImport((IUnknown **)&pImport));
+
+        if (TypeFromToken(m_token) == mdtTypeDef)
+        {
+            mdTypeDef tkEnclosing = mdTokenNil;
+            mdTypeDef tkCurrent = m_token;
+            tstring fullName;
+            vector<WCHAR> nameBuffer(100);
```
jander-msft, line 46: >nameBuffer [](start = 26, length = 10)

Should you zero this on each iteration of the loop? The GetTypeDefProps probably always null terminates the name, so maybe not necessary.

src/InstrumentationEngine/TokenType.cpp
```diff
@@ -1,69 +1,99 @@
...
+            IfFailRet(pImport->GetTypeDefProps(m_token, nullptr, 0, &cchLength, nullptr, nullptr));
+            if (nameBuffer.size() < cchLength)
+            {
+                nameBuffer.resize(cchLength);
+            }
+
+            IfFailRet(pImport->GetTypeDefProps(m_token, nameBuffer.data(), cchLength, &cchLength, nullptr, nullptr));
+            fullName += nameBuffer.data();
+
+            m_name = CComBSTR(static_cast<ULONG>(fullName.length()+1), fullName.c_str());
```
jander-msft, line 73: >static_cast<ULONG>(fullName.length()+1), fullName.c_str() [](start = 30, length = 57)

Why not just pass fullName.c_str()?

If not, why do you add 1 to the length? The CComBSTR documentation seems to suggest that it will copy that many characters from the source string and then add a null terminating character. So you might be copying the null terminating character from the source, and then it adds its own null terminating character, thus having two null terminating characters; might be an unnecessary overallocation.

src/InstrumentationEngine/TokenType.cpp
delmyers: Updated to use '+'

src/InstrumentationEngine/TokenType.cpp
delmyers: I could give it a try. I was just following what was being done previously. The cchNameLength includes the null terminator (or there wouldn't be enough space for the string), and previously it was copying the entire array into the bstr.

src/InstrumentationEngine/TokenType.cpp
wiktork, line 46: >nameBuffer [](start = 26, length = 10)

consider vector.reserve instead of resize

delmyers: Then I just have to check the capacity. Plus resizing will allocate more memory than needed. Reserving allocates _exactly_ what is requested, so it is probably less efficient.

PR #248, opened 2020-04-03
Fix signing XdtExtensions.dll

WilliamXieMSFT:


src/InstrumentationEngine.XdtExtensions/XdtExtensions.csproj
```diff
@@ -11,6 +11,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Web.Xdt" Version="1.4.0" />
+    <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" Version="$(MicroBuildCoreVersion)" />
```
delmyers, line 14: >Microsoft.VisualStudioEng.MicroBuild.Core [](start = 31, length = 41)

Why does this need to reference microbuild? Shouldn't it be installed on the build machines?

WilliamXieMSFT: Referencing the microbuild packages allows the project to consume the nuget's build props and targets which redirect to the installed microbuild signing plugin.

PR #249, opened 2020-04-03
Fix patching xdtextensions.dll in test extension

WilliamXieMSFT:
$(InputBinCfgRoot) is set in the pipeline to point to the binary artifacts produced during the build binaries stage, however the xdtextensions.dll are produced in the bin\configuration path.

### DIFF
--- a/build/SiteExtensions.targets
+++ b/build/SiteExtensions.targets
@@ -46,7 +46,7 @@
       <ZipTestItem Include="@(SiteExtensionContents)">
         <Destination>$(Title)\%(RootPath)\%(RecursiveDir)</Destination>
       </ZipTestItem>
-      <ZipTestItem Include="$(InputBinCfgRoot)\AnyCPU\XdtExtensions.*"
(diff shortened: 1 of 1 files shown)


build/SiteExtensions.targets
```diff
@@ -46,7 +46,7 @@
       <ZipTestItem Include="@(SiteExtensionContents)">
         <Destination>$(Title)\%(RootPath)\%(RecursiveDir)</Destination>
       </ZipTestItem>
-      <ZipTestItem Include="$(InputBinCfgRoot)\AnyCPU\XdtExtensions.*"
+      <ZipTestItem Include="$(BinRoot)\$(Configuration)\AnyCPU\XdtExtensions.*"
```
jander-msft, line 49: Isn't "$(BinRoot)\$(Configuration)\AnyCPU" just $(OutputPath)?

WilliamXieMSFT: Yep

PR #256, opened 2020-05-04
Remove duplicate Initialize callback

WilliamXieMSFT:


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -244,22 +244,12 @@ HRESULT CProfilerManager::AddRawProfilerHook(
...
 
-    if (m_attachedClrVersion != ClrVersion_2)
-    {
-        hr = pCorProfilerCallback->Initialize(m_pWrappedProfilerInfo);
-    }
-    else
-    {
-        hr = pCorProfilerCallback->Initialize(m_pRealProfilerInfo);
-    }
-
```
wiktork, line 255: You may want to maintain the error semantics that we have now. The other Initialize call doesn't care about the HRESULT of the raw profiler hook. This one does.

WilliamXieMSFT: Good catch!

---
In reply to: [419803414](https://github.com/microsoft/CLRInstrumentationEngine/pull/256#discussion_r419803414) [](ancestors = 419803414)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -932,6 +922,12 @@ HRESULT CProfilerManager::Initialize(
         }
     }
 
+    if (FAILED(hr))
```
jander-msft, line 925: This should go inside the "if (m_profilerCallbackHolder != nullptr)" so that it's only considered when the raw profiler is available. Also, consider just using IfFailRet on the Intialize calls instead of manually checking of failure, logging a custom error, and returning.

WilliamXieMSFT: Thanks

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -244,22 +244,12 @@ HRESULT CProfilerManager::AddRawProfilerHook(
...
-    else
-    {
-        hr = pCorProfilerCallback->Initialize(m_pRealProfilerInfo);
-    }
-
     if (FAILED(hr))
     {
         CLogging::LogError(_T("Raw profiler hook returned failure"));
         return E_FAIL;
     }
```
delmyers, line 251: I don't think that this is valid anymore. The last hr will be set to the result of the query interface, which could validly be E_NOINTERFACE. What you are doing here is making the profiler incompatible raw profilers that implement callbacks >= 7. I could see valid reasons that the raw profilers would not implement this interface (they might just not need the callbacks).

WilliamXieMSFT: Thanks for catching this

---
In reply to: [420238513](https://github.com/microsoft/CLRInstrumentationEngine/pull/256#discussion_r420238513) [](ancestors = 420238513)

WilliamXieMSFT:
This also addressed #273

PR #258, opened 2020-05-07
Replace LoggingFlags with atomic_int

WilliamXieMSFT:
To reduce the load of our logging infrastructure for InstrumentationMethods, this change allows us to remove the use of critical sections when getting the effective logging flags (which should not change very often).

### DIFF
--- a/src/InstrumentationEngine.Lib/LoggerService.cpp
+++ b/src/InstrumentationEngine.Lib/LoggerService.cpp
@@ -67,14 +67,12 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
 
     IfNotInitRetUnexpected(m_initialize);
 
-    CCriticalSectionHolder holder(&m_cs);
-
     // Intentionally returning the effective flags instead of the default flags
     // (ones set by default or by SetLoggingFlags). The GetLoggingFlags method
     // is primarily meant for consumers external of the Instrumentation Engine
     // assembly to determine which flags are supported rather than literally
     // reporting what the default was or what was set by SetLoggingFlags.
-    *pLoggingFlags = m_effectiveFlags;
+    *pLoggingFlags = (LoggingFlags)m_effectiveFlags.load(std::memory_order_acquire);
 
     return S_OK;
 }
@@ -359,7 +357,8 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
         effectiveFlags = static_cast<LoggingFlags>(effectiveFlags | sinkFlags);
     }
 
-    m_effectiveFlags = effectiveFlags;
+    //m_effectiveFlags = effectiveFlags;
+    m_effectiveFlags.store((int)effectiveFlags, std::memory_order_release);
 
     return S_OK;
 }
--- a/src/InstrumentationEngine.Lib/LoggerService.cpp
+++ b/src/InstrumentationEngine.Lib/LoggerService.cpp
@@ -72,7 +72,7 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
     // is primarily meant for consumers external of the Instrumentation Engine
     // assembly to determine which flags are supported rather than literally
     // reporting what the default was or what was set by SetLoggingFlags.
-    *pLoggingFlags = m_effectiveFlags.load();
+    *pLoggingFlags = m_effectiveFlags;
 
     return S_OK;
 }
--- a/src/InstrumentationEngine.Lib/LoggerService.cpp
+++ b/src/InstrumentationEngine.Lib/LoggerService.cpp
@@ -72,7 +72,7 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
     // is primarily meant for consumers external of the Instrumentation Engine
     // assembly to determine which flags are supported rather than literally
     // reporting what the default was or what was set by SetLoggingFlags.
-    *pLoggingFlags = (LoggingFlags)m_effectiveFlags.load(std::memory_order_acquire);
+    *pLoggingFlags = m_effectiveFlags.load();
 
     return S_OK;
 }
@@ -357,8 +357,7 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
         effectiveFlags = static_cast<LoggingFlags>(effectiveFlags | sinkFlags);
     }
 
-    //m_effectiveFlags = effectiveFlags;
-    m_effectiveFlags.store((int)effectiveFlags, std::memory_order_release);
+    m_effectiveFlags = effectiveFlags;
 
     return S_OK;
 }
--- a/src/InstrumentationEngine.Lib/LoggerService.h
+++ b/src/InstrumentationEngine.Lib/LoggerService.h
@@ -53,7 +53,7 @@ namespace MicrosoftInstrumentationEngine
         // each of the logger sinks. This is updated each time a logger sink dependency is changed
         // e.g. calling SetLoggingFlags, SetLoggingHost, and SetLogToDebugPort.
         //
-        // We're using atomic_int here to avoid a critical section in GetLoggingFlags.
+        // We're using atomic here to avoid a critical section in GetLoggingFlags.
         std::atomic<LoggingFlags> m_effectiveFlags;
         // This is the cumulative LoggingFlags for all InstrumentationMethods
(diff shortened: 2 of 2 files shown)


src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -67,14 +67,12 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
...
 
-    CCriticalSectionHolder holder(&m_cs);
-
     // Intentionally returning the effective flags instead of the default flags
     // (ones set by default or by SetLoggingFlags). The GetLoggingFlags method
     // is primarily meant for consumers external of the Instrumentation Engine
     // assembly to determine which flags are supported rather than literally
     // reporting what the default was or what was set by SetLoggingFlags.
-    *pLoggingFlags = m_effectiveFlags;
+    *pLoggingFlags = (LoggingFlags)m_effectiveFlags.load(std::memory_order_acquire);
```
wiktork, line 75: You'll also need to change AllowLogEntry

https://github.com/microsoft/CLRInstrumentationEngine/blob/d034e98ada89a49a08dfc94bbf850c0f301714d1/src/InstrumentationEngine.Lib/LoggerService.cpp#L277

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -359,7 +357,8 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
         effectiveFlags = static_cast<LoggingFlags>(effectiveFlags | sinkFlags);
     }
 
-    m_effectiveFlags = effectiveFlags;
+    //m_effectiveFlags = effectiveFlags;
```
wiktork, line 360: remove comment

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -67,14 +67,12 @@ HRESULT CLoggerService::GetLoggingFlags(_Out_ LoggingFlags* pLoggingFlags)
 
     IfNotInitRetUnexpected(m_initialize);
 
-    CCriticalSectionHolder holder(&m_cs);
```
wiktork, line 70: While this is a good first step, don't you also need to move the lock acquisition in functions like LogMessage and others in order to take advantage of them?

WilliamXieMSFT: I think we still need the lock acquisition in those functions because they guard the logging sinks. This change ensures we  don't lock in LogMessage and quickly return if "Errors" are enabled.

src/InstrumentationEngine.Lib/LoggerService.cpp
WilliamXieMSFT: Thanks

---
In reply to: [421832135](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#discussion_r421832135) [](ancestors = 421832135)

src/InstrumentationEngine.Lib/LoggerService.h
```diff
@@ -52,7 +52,9 @@ namespace MicrosoftInstrumentationEngine
         // The effective set of flags (flags that are in actual use) as determined by querying
         // each of the logger sinks. This is updated each time a logger sink dependency is changed
         // e.g. calling SetLoggingFlags, SetLoggingHost, and SetLogToDebugPort.
-        LoggingFlags m_effectiveFlags;
+        //
+        // We're using atomic_int here to avoid a critical section in GetLoggingFlags.
```
WilliamXieMSFT, line 56: >atomic_int [](start = 23, length = 10)

comment

src/InstrumentationEngine.Lib/LoggerService.cpp
WilliamXieMSFT: Do you mean I should change m_instrumentationMethodFlags to type atomic<LoggingFlags> ?

---
In reply to: [421832010](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#discussion_r421832010) [](ancestors = 421832010)

src/InstrumentationEngine.Lib/LoggerService.h
delmyers, line 56: >atomic_int [](start = 23, length = 10)

This isn't an atomic_int

src/InstrumentationEngine.Lib/LoggerService.cpp
delmyers: AllowLogEntry is always called within a critical section, there should be no race there.

---
In reply to: [421837264](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#discussion_r421837264) [](ancestors = 421837264,421832010)

delmyers:
Did you already compile this and send bits to Jakub? I would like to see the results before checking this in. More work might need to be done.

src/InstrumentationEngine.Lib/LoggerService.cpp
wiktork: Hmm, that's not what https://github.com/microsoft/CLRInstrumentationEngine/blob/0ae92f3bd8249aa5eff8615a6f3d2b1602c5fbf1/src/InstrumentationEngine.Lib/LoggerService.cpp#L170-L180 seems to be doing

delmyers:
@jakubch1 Would you mind signing off on this PR when you have verified that the changes have improved performance to a satisfactory level?

WilliamXieMSFT:
Yep, sent a link in the chat

---
In reply to: [625537348](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#issuecomment-625537348) [](ancestors = 625537348)

src/InstrumentationEngine.Lib/LoggerService.cpp
delmyers: Though, we might be able to save additional time if AllowLogEntry can be pulled out of the critical sections.

---
In reply to: [421839504](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#discussion_r421839504) [](ancestors = 421839504,421837264,421832010)

src/InstrumentationEngine.Lib/LoggerService.cpp
WilliamXieMSFT: CLogging checks GetLoggingFlags before calling LoggerService::LogMessage, but you're right, a call from some other place to LoggerService would hit this critical section.

jakubch1:
I've compared this branch (dev/willxie/atomic) with my branch without logging. Results:

Dynamic CLR IE ( dev/willxie/atomic): 00:00:51.0227404 00:00:50.7287865 00:00:51.0248403 00:00:51.9103193
Dynamic CLR IE No logging: 00:00:49.9031805 00:00:50.4418515 00:00:54.9378793 00:00:49.6434928

It looks good. I can provide perfview results from run from this branch if you need.

I have also some data about the slowest instrumentation:

Method name: Microsoft.CodeAnalysis.CSharp.CSharpCommandLineParser.Parse instrumented in: 6910ms  Number of instr: 4917  Number of branches: 1669

in this case we invoke LogMessage at least 4917*3*1669*4*2 + 4917*1669*2*2 = 229.781.244 times

Thanks for help.

src/InstrumentationEngine.Lib/LoggerService.cpp
delmyers: We might be able to speed things up some more by updating AllowLogEntry so that it can be called before the critical section. However, most callers don't go through this directly. They first get the flags from the logging service, check to see if the flag matches what they expect, and then call log. AllowLogEntry does a double-check. We have verification from Jakub that, under normal usage, this fixes his issue.

---
In reply to: [421841204](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#discussion_r421841204) [](ancestors = 421841204)

delmyers:
Awesome. Seems like 1 second is within a margin of error. It would be great if we could get another perf trace, though. Your scenario is definitely the one that stresses ClrInstrumentationEngine the most with regular usage. It would be nice to know if there are more things that we could do to speed it up. I would put those fixes as separate issues, though.

---
In reply to: [625724679](https://github.com/microsoft/CLRInstrumentationEngine/pull/258#issuecomment-625724679) [](ancestors = 625724679)

src/InstrumentationEngine.Lib/LoggerService.cpp
WilliamXieMSFT: I'll add a task for addressing AllowLogEntry

PR #260, opened 2020-05-08
Disable Wpragma-pack on unix

hughbe:
E.g.

```
/home/hughbe/CLRInstrumentationEngine/out/Linux/CoreCLRPAL/inc/rt/pshpack1.h:36:9: note: previous '#pragma pack' directive that modifies alignment is here [Semantic Issue]
#pragma pack(1)
        ^
In file included from /home/hughbe/CLRInstrumentationEngine/src/InstrumentationEngine.Lib/ConfigurationLoader.cpp:4:
In file included from /home/hughbe/CLRInstrumentationEngine/src/InstrumentationEngine.Lib/./stdafx.h:45:
/home/hughbe/CLRInstrumentationEngine/out/Linux/CoreCLRPAL/inc/cor.h:1951:10: error: the current #pragma pack aligment value is modified in the included file [-Werror,-Wpragma-pack,Semantic Issue]
#include <poppack.h>
         ^
/home/hughbe/CLRInstrumentationEngine/out/Linux/CoreCLRPAL/inc/rt/poppack.h:37:9: note: previous '#pragma pack' directive that modifies alignment is here [Semantic Issue]
#pragma pack()
        ^
18 errors generated.
18 errors generated
```

See https://github.com/dotnet/coreclr/pull/16855/files

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -149,6 +149,7 @@ macro (build_init build_language build_type)
     add_compile_options(-Wno-new-returns-null)
     add_compile_options(-Wno-deprecated-declarations)
     add_compile_options(-Wno-delete-incomplete)
+    add_compile_options(-Wno-pragma-pack)
 
     # Turn all enabled warnings into errors
     add_compile_options(-Werror)


weltkante:
If pragma pack is not available doesn't that mean the current headers will have the wrong memory layout? Might need more work than just disabling the warning.

hughbe:
I'm fairly sure the error isn't that pragma pack isn't defined, but rather that it is being modified in the file?

weltkante:
Yeah but thats the point of these headers, to modify the default packing for a section of declarations.

hughbe:
> Yeah but thats the point of these headers, to modify the default packing for a section of declarations.

Correct. Basically the warning is tell us not to do precisely what we want to do. Here's the output of a clean build below
[output.txt](https://github.com/microsoft/CLRInstrumentationEngine/files/4598825/output.txt)

For this reason we disabled this warning in coreclr: https://github.com/dotnet/coreclr/pull/16855/files#diff-3b3e127c2ec93e1085823b22265a52f9R44

hughbe:
> Looks like this warning only comes up from the #include of palrt.h inside unix.h. Could we surgically disable it just for the duration of that include using #pragmas inside unix.h instead of adding it to the cmake?

I did this change in `CMakeLists.txt` because I wanted to be consistent with the CLR and also because I didn't want to have to add a warning supression everywhere its used.

That said I gave your suggestion a punt by having the following code in `palrt.h` instead of in `CMakeLists.txt`
```cpp
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wpragma-pack"
#include "palrt.h"
#pragma clang diagnostic pop
```

This disabled some warnings but a whole load of others flared up in code like `DbgHelp.h`. See output.txt below

[output.txt](https://github.com/microsoft/CLRInstrumentationEngine/files/4604101/output.txt)

I would suggest that we stick with this current approach

delmyers:
> This disabled some warnings but a whole load of others flared up in code like `DbgHelp.h`. See output.txt below
> 
> [output.txt](https://github.com/microsoft/CLRInstrumentationEngine/files/4604101/output.txt)
> 
> I would suggest that we stick with this current approach

OK, sounds good.

delmyers:
Unfortunately, we can't take this change right now.

PR #261, opened 2020-05-08
Don't define builtins on newer clang versions in palrt2.h

hughbe:
Avoids errors such as `error: definition of builtin function '_InterlockedIncrement`

### DIFF
--- a/src/unix/inc/palrt2.h
+++ b/src/unix/inc/palrt2.h
@@ -44,30 +44,40 @@ inline void * _recalloc(void *memblock, size_t num, size_t size)
 #define ULLONG_MAX    0xffffffffffffffffu       /* maximum unsigned long long int value */
 EXTERN_C PALAPI errno_t _ultow_s( unsigned long inValue, WCHAR* outBuffer, size_t inDestBufferSize, int inRadix );
 
+#if !__has_builtin(_InterlockedIncrement)
 inline LONG _InterlockedIncrement(LONG volatile * _Addend)
 {
     return InterlockedIncrement(_Addend);
 }
+#endif
 
+#if !__has_builtin(_InterlockedDecrement)
 inline LONG _InterlockedDecrement(LONG volatile * _Addend)
 {
     return InterlockedDecrement(_Addend);
 }
+#endif
 
+#if !__has_builtin(_InterlockedOr)
 inline LONG _InterlockedOr(LONG volatile * Destination, LONG Value)
 {
     return InterlockedOr(Destination, Value);
 }
+#endif
 
+#if !__has_builtin(_InterlockedAnd)
 inline LONG _InterlockedAnd(LONG volatile * Destination, LONG Value)
 {
     return InterlockedAnd(Destination, Value);
 }
+#endif
 
+#if !__has_builtin(_InterlockedExchange)
 inline LONG _InterlockedExchange(LONG volatile * Target, LONG Value)
 {
     return InterlockedExchange(Target, Value);
 }
+#endif
 
 // --- guid ---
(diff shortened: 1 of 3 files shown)


delmyers:
Hi, @hugbe, it looks like you are trying to get Mac support for ClrInstrumentationEngine. If that's the case, we should probably have a larger discussion about that as a community since we don't have any validation for mac scenarios right now. The ClrInstrumentationEngine depends on an abstraction layer called the PAL, which we don't link to for Mac machines currently.

Also, it seems that this change might not be the correct one. @jander-msft @wiktork you might have more context, but it seems like we would probably need to do a little more refactoring to get this to work? I'm not sure.

hughbe:
I made some cmake modifications to get it building on mac as a test.

However, I created a WSL Ubuntu 18.04 with latest version of clang and got these errors too. So this isn’t necessarily a fix for Mac

hughbe:
> Also, it seems that this change might not be the correct one. @jander-msft @wiktork you might have more context, but it seems like we would probably need to do a little more refactoring to get this to work? I'm not sure.
 
I’ll try to detail what I’ve done here. Basically, at some point between clang 3.5 (which is the one we use to build this repo) and the latest version, clang added support for _Interlocked* intrinsics on Windows, mac, Linux etc.

Basically clang throws an error if you try to redefine a builtin. By guarding for this using #if !__has_builtin we keep the existing behaviour on older compilers and fix the build on newer ones. We can delete the code for _Interlocked* if we start supporting clang with minimum version that already has the built ins

WilliamXieMSFT:
Spoke with @jander-msft offline, Clang 3.5 was used by the dotnet core libraries at the time we adopted their PAL implementation, so likely not able to upgrade any time soon. I'm fine with this change as a stop-gap, @hughbe, but you may encounter other issues down the line.

src/unix/inc/palrt2.h
```diff
@@ -82,7 +92,7 @@ __inline int InlineIsEqualGUID(REFGUID rguid1, REFGUID rguid2)
 
 // --- winerror ---
 #define ERROR_INVALID_THREAD_ID          1444L
-#define ERROR_ELEVATION_REQUIRED         740L
+#define ERROR_ELEVATIONO_REQUIRED         740L
```
delmyers, line 95: Looks like a typo?

hughbe:
> Spoke with @jander-msft offline, Clang 3.5 was used by the dotnet core libraries at the time we adopted their PAL implementation, so likely not able to upgrade any time soon

I will note that Ubuntu 18.04 does not have clang 3.5 available (`sudo apt-get clang-3.5` Does not work)

src/unix/inc/palrt2.h
hughbe: Yup not sure how this snuck in

hughbe:
Could you try triggering the build again? I suspect there may have been a sign issue where we casted ULONG to LONG. Not sure though. If that doesn't work, will see about `volatile`

hughbe:
Looks like this worked :)

WilliamXieMSFT:
@hughbe Looks like the builds fail with:

d:\a\1\s\src\InstrumentationEngine.Lib\refcount.cpp(245,69): error C2220: the following warning is treated as an error [d:\a\1\s\src\InstrumentationEngine.Lib\InstrumentationEngine.Lib.vcxproj]
##[warning]src\InstrumentationEngine.Lib\refcount.cpp(245,69): Warning C4197: 'volatile PLONG': top-level volatile in cast is ignored

hughbe:
Fingers crossed... sorry about this

delmyers:
> Fingers crossed... sorry about this

Trying the build again

PR #262, opened 2020-05-08
Include iconv.h in ConfigurationLoader.cpp

hughbe:
Seems like this is implicitly included on my ubuntu WSL distro, but it isn't on mac etc.

This can get some erorrs e.g. for `iconv_t icnv = iconv_open("UTF-16LE", "UTF-8");` we get 'iconv_t' was not declared in this scope

### DIFF
--- a/src/InstrumentationEngine.Lib/ConfigurationLoader.cpp
+++ b/src/InstrumentationEngine.Lib/ConfigurationLoader.cpp
@@ -219,6 +219,7 @@ HRESULT CConfigurationLoaderHelper::ProcessInstrumentationMethodNode(_In_ BSTR b
 #else
 
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/InstrumentationEngine.Lib/ConfigurationLoader.cpp
```diff
@@ -219,6 +219,7 @@ HRESULT CConfigurationLoaderHelper::ProcessInstrumentationMethodNode(_In_ BSTR b
 #else
 
 #include <string.h>
+#include <iconv.h>
```
delmyers, line 222: Would you mind adding an issue for this, and attaching the full error that you get from your compiler, including the compiler version?

Note, we don't currently have any validation running for mac machines.

PR #264, opened 2020-05-11
Add cached logging flag

WilliamXieMSFT:
This allows CLogging to bypass querying the CLoggerService and use a static flag instead. On CLoggerService::RecalculateLoggingFlags(), it calls back into CLogging to update the static flag.

### DIFF
--- a/src/InstrumentationEngine.Lib/LoggerService.cpp
+++ b/src/InstrumentationEngine.Lib/LoggerService.cpp
@@ -133,13 +133,9 @@ void CLoggerService::SetLogToDebugPort(_In_ bool enable)
     RecalculateLoggingFlags();
 }
 
-HRESULT CLoggerService::Initialize(_In_ std::function<void(const LoggingFlags&)> callback)
+HRESULT CLoggerService::Initialize(_In_ std::function<void(const LoggingFlags&)> loggingFlagsCallback)
 {
-    if (callback != nullptr)
-    {
-        m_LoggingFlagsCallback = callback;
-    }
-
+    m_LoggingFlagsCallback = loggingFlagsCallback;
     return m_initialize.Get();
 }
 
--- a/src/InstrumentationEngine.Lib/LoggerService.cpp
+++ b/src/InstrumentationEngine.Lib/LoggerService.cpp
@@ -133,8 +133,13 @@ void CLoggerService::SetLogToDebugPort(_In_ bool enable)
     RecalculateLoggingFlags();
 }
 
-HRESULT CLoggerService::Initialize()
+HRESULT CLoggerService::Initialize(_In_ std::function<void(const LoggingFlags&)> callback)
 {
+    if (callback != nullptr)
+    {
+        m_LoggingFlagsCallback = callback;
+    }
+
     return m_initialize.Get();
 }
 
@@ -358,7 +363,10 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
     }
 
     m_effectiveFlags = effectiveFlags;
-    CLogging::SetCachedLoggingFlag(effectiveFlags);
+    if (m_LoggingFlagsCallback != nullptr)
+    {
+        m_LoggingFlagsCallback(effectiveFlags);
+    }
 
     return S_OK;
 }
--- a/src/InstrumentationEngine.Lib/LoggerService.cpp
+++ b/src/InstrumentationEngine.Lib/LoggerService.cpp
@@ -358,6 +358,7 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
     }
 
     m_effectiveFlags = effectiveFlags;
+    CLogging::SetCachedLoggingFlag(effectiveFlags);
 
     return S_OK;
 }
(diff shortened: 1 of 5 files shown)


src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -358,6 +358,7 @@ HRESULT CLoggerService::RecalculateLoggingFlags()
     }
 
     m_effectiveFlags = effectiveFlags;
+    CLogging::SetCachedLoggingFlag(effectiveFlags);
```
delmyers, line 361: >CLogging::SetCachedLoggingFlag(effectiveFlags); [](start = 4, length = 47)

This looks like a strange circular dependency. I didn't know that the logger service had access to CLogging. Seems like it shouldn't.

I was expecting that you add a way to register a callback on the CLoggerService for when the flags changed.

delmyers: @jander-msft what do you think of this? Is it good enough? You wrote this code initially.

src/InstrumentationEngine.Lib/Logging.h
```diff
@@ -52,6 +52,9 @@ namespace MicrosoftInstrumentationEngine
         static CSingleton<CLoggerService> s_loggerService;
         static std::atomic_size_t s_initCount;
 
+        // This caches the effective loggingflag of s_loggerService for performance
+        static std::atomic<LoggingFlags> s_cachedFlags;
```
wiktork, line 56: Is this really that much more performant? What makes GetLoggingFlags so slow given that it just reads m_effectiveFlags?

src/InstrumentationEngine.Lib/Logging.cpp
```diff
@@ -174,6 +177,15 @@ HRESULT CLogging::SetLogFileLevel(_In_ LoggingFlags fileLogFlags)
     return s_loggerService.Get()->SetLogFileLevel(fileLogFlags);
 }
 
+// static
+HRESULT CLogging::SetCachedLoggingFlag(_In_ LoggingFlags flags)
+{
+    // Do not check initialization as that can cause an infinite loop.
```
delmyers, line 183: nit: If you want to keep things as-is and not have to check initialization, you could tighten this up by making this function private, and declaring CLoggerService as a friend class. That way no one else accidentally calls it.

Some people don't like using friend classes, though, so, I'm not tied to the idea.

src/InstrumentationEngine.Lib/Logging.h
delmyers: Currently, in Code coverage scenarios (which are very heavy), calls to CLogging::LogMessage take up about 8% of cpu time. If we remove the calls to initialize the logger service, and to get the logging flags, we may be able to get that down to about 4%.

src/InstrumentationEngine.Lib/LoggerService.cpp
jander-msft: I was expecting callback registration as well. Though, this is probably good enough for now.

I think it would be nice to allow for IMs to do callback registration for logging flag changes as well so that they don't have to call IProfilerManagerLogging::GetLoggingFlags for each log operation, but that is probably a separate issue we'll want to address.

WilliamXieMSFT: I can implement the callback registration for this.

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -133,8 +133,13 @@ void CLoggerService::SetLogToDebugPort(_In_ bool enable)
     RecalculateLoggingFlags();
 }
 
-HRESULT CLoggerService::Initialize()
+HRESULT CLoggerService::Initialize(_In_ std::function<void(const LoggingFlags&)> callback)
```
delmyers, line 136: >In_ std::function<void(const LoggingFlags&)> callback [](start = 36, length = 53)

 I think Justin was looking for something that would be possible for instrumentation methods to register to as well. That would imply adding something to the idl, and making it late registration. I'm fin with this for now, though as well. I wouldn't want to block this check in.

delmyers, line 136: >callback [](start = 81, length = 8)

loggingFlagsCallback?

WilliamXieMSFT: True, I didn't want to modify the API surface yet, we can definitely address these callbacks holistically in the future.

src/InstrumentationEngine.Lib/LoggerService.cpp
```diff
@@ -133,8 +133,13 @@ void CLoggerService::SetLogToDebugPort(_In_ bool enable)
...
 {
+    if (callback != nullptr)
```
delmyers, line 138: >if (callback != nullptr) [](start = 4, length = 24)

Don't do this check. It will currently cause m_LoggingFlagsCallback to be random memory if callback == nullptr.

delmyers: Actually, std::function might initialize to null. I'm not sure how that works, to be honest.


---
In reply to: [423354284](https://github.com/microsoft/CLRInstrumentationEngine/pull/264#discussion_r423354284) [](ancestors = 423354284)

delmyers: Still, get rid of the check :-). It isn't needed.

---
In reply to: [423355501](https://github.com/microsoft/CLRInstrumentationEngine/pull/264#discussion_r423355501) [](ancestors = 423355501,423354284)

WilliamXieMSFT: Thanks

PR #265, opened 2020-05-11
Update Ubuntu build image to 18.04 and use Clang 3.9 on Ubuntu

jander-msft:


src/unix/docker/dockerfiles/build/ubuntu/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN dotnet help
 # CLR Instrumentation Engine Build Prerequisites
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-        clang-3.5 \
+        clang-3.9 \
```
delmyers, line 28: >3.9 [](start = 14, length = 3)

The instructions in the build.sh say to use clang 4.0, but this uses clang 3.9. Is that on purpose?

jander-msft: Yes, this is matching the same clang version for each distro on which the CoreCLR PAL was produced in order to match the ABI for each CoreCLR PAL payload. The musl version was built on alpine 3.6 using clang 4.0; the gnu version was built on centos 7 using clang 3.9.

I would have to dynamically determine which os or clib is being used to properly report the exact clang version that the build script would suggest to install.

PR #270, opened 2020-06-08
Support specifying CoreCLR-specific RawProfilerHook.

WilliamXieMSFT:
This change introduces the ability to specify CoreCLR-specific RawProfilerHook. If the CoreCLR RPH variables are not set, CLRIE falls back to the original RPH variables.

The following variables are supported:
* MicrosoftInstrumentationEngine_CoreRawProfilerHook
* MicrosoftInstrumentationEngine_CoreRawProfilerHookPath_32
* MicrosoftInstrumentationEngine_CoreRawProfilerHookPath_64

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -50,8 +50,10 @@ The RawProfilerHook allows one additional profiler that has not yet on-boarded t
 
 | Variable | Value | Description |
 |-|-|-|
-MicrosoftInstrumentationEngine_RawProfilerHook|"{GUID}"|This would be the value set to CORECLR/COR_PROFILER.
-MicrosoftInstrumentationEngine_RawProfilerHookPath_32/64|"[FULL PATH TO raw profiler dll]"|This would be the value set to CORECLR/COR_PROFILER_PATH_32/64.
+MicrosoftInstrumentationEngine_RawProfilerHook|"{GUID}"|This would be the value set to COR_PROFILER.
+MicrosoftInstrumentationEngine_CoreRawProfilerHook|"{GUID}"|This would be the value set to CORECLR_PROFILER. Falls back to MicrosoftInstrumentationEngine_RawProfilerHook if this is not set.
+MicrosoftInstrumentationEngine_RawProfilerHookPath_32/64|"[FULL PATH TO raw profiler dll]"|This would be the value set to COR_PROFILER_PATH_32/64.
+MicrosoftInstrumentationEngine_CoreRawProfilerHookPath_32/64|"[FULL PATH TO raw profiler dll]"|This would be the value set to CORECLR_PROFILER_PATH_32/64. Only checked if MicrosoftInstrumentationEngine_CoreRawProfilerHook is set.
 
 ## Deprecated as of Version 1.0.22
 The following variables allowed custom ExtensionHosts for the InstrumentationEngine. The responsibility of the ExtensionsHost involves setting
--- a/src/InstrumentationEngine.Lib/RawProfilerHookSettingsReader.cpp
+++ b/src/InstrumentationEngine.Lib/RawProfilerHookSettingsReader.cpp
@@ -6,12 +6,17 @@
 
 namespace MicrosoftInstrumentationEngine
 {
-    LPCWSTR cszRawProfilerHookVariableName = L"MicrosoftInstrumentationEngine_RawProfilerHook";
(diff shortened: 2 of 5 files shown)


WilliamXieMSFT:
The current behavior is that CoreCLR RPH will fallback to desktop CLR RPH, however that may not be what we want. I can restrict it so that desktop RPH will not be used for CoreCLR processes.

wiktork:
> The current behavior is that CoreCLR RPH will fallback to desktop CLR RPH, however that may not be what we want. I can restrict it so that desktop RPH will not be used for CoreCLR processes.

While I would agree wit this behavior (since it matches the original COR_PROFILER and CORECLR_PROFILER concept), it is a breaking change.

docs/environment_variables.md
```diff
@@ -50,8 +50,10 @@ The RawProfilerHook allows one additional profiler that has not yet on-boarded t
 
 | Variable | Value | Description |
 |-|-|-|
-MicrosoftInstrumentationEngine_RawProfilerHook|"{GUID}"|This would be the value set to CORECLR/COR_PROFILER.
-MicrosoftInstrumentationEngine_RawProfilerHookPath_32/64|"[FULL PATH TO raw profiler dll]"|This would be the value set to CORECLR/COR_PROFILER_PATH_32/64.
+MicrosoftInstrumentationEngine_RawProfilerHook|"{GUID}"|This would be the value set to COR_PROFILER.
+MicrosoftInstrumentationEngine_CoreRawProfilerHook|"{GUID}"|This would be the value set to CORECLR_PROFILER. Falls back to MicrosoftInstrumentationEngine_RawProfilerHook if this is not set.
```
wiktork, line 54: So this solves the situation where someone wants a raw profiler for coreclr, but not desktop clr. The opposite scenario is not possible (without a breaking change).

WilliamXieMSFT: Correct, we could introduce an additional separate desktop-specific RPH variable as well, that would not break anything.

WilliamXieMSFT:
Closing this PR for now, we can re-open this PR if CLRIE requires supporting this API in the future.

src/InstrumentationEngine.Lib/RawProfilerHookSettingsReader.cpp
```diff
@@ -27,45 +33,74 @@ namespace MicrosoftInstrumentationEngine
...
 
-        CLogging::LogMessage(_T("Examining environment variable: %s"), cszRawProfilerHookVariableName);
+        // If CoreCLR RPH variable is not set, fallback to the desktop one if it exists.
+        // We only want to use the CoreCLR RPH path variable if the CoreCLR classId was set.
+        bool fUseCoreClrPath; 
+        if (runtimeType == COR_PRF_RUNTIME_TYPE::COR_PRF_CORE_CLR)
+        {
+            fUseCoreClrPath = true;
+            CLogging::LogMessage(_T("Examining coreCLR environment variable: %s"), cszCoreRawProfilerHookVarName);
+            IfFailRet(GetEnvironmentVariableWrapper(cszCoreRawProfilerHookVarName, strRawProfilerHookClsid));
```
AbhitejJohn, line 43: Should we be logging if the environment variable isn't found here?

docs/environment_variables.md
AbhitejJohn: Or if a data collector wants to blindly specify both CLR and CoreCLR profiler information and let the runtime decide what profiler to load.

src/InstrumentationEngine.Lib/RawProfilerHookSettingsReader.cpp
WilliamXieMSFT: Sure, that will help with tracking what happened here.

---
In reply to: [436912916](https://github.com/microsoft/CLRInstrumentationEngine/pull/270#discussion_r436912916) [](ancestors = 436912916)

AbhitejJohn:
From our offline discussion, this would become useful (specifically in our workflow) when the Raw Profiler Hook supports IntelliTrace as well. Although the requirement isn't super critical, it would be nice for dogfooding if it isn't very involved. I agree though in the long run that moving Code Coverage to CLRIE would satisfy these workflows better (but that might be some way off).

src/InstrumentationEngine.Lib/RawProfilerHookSettingsReader.cpp
WilliamXieMSFT: Yes, good idea

PR #272, opened 2020-06-26
Address issue where GetILFunctionBody may be called during OnModuleLoadFinished

WilliamXieMSFT:
There are two callbacks associated with module load completion. ICorProfilerCallback::ModuleLoadFinished occurs when the module is successfully loaded and the associated moduleId is valid for use. However the module does not know its associated assembly until ICorProfilerCallback::ModuleAttachedToAssembly which occurs later.

CLRIE constructs a CModuleInfo object for each moduleId and populates a cached mapping of moduleId to moduleInfo during ModuleAttachedToAssembly. CLRIE then reports to InstrumentationMethods of the IInstrumentationMethod::OnModuleLoaded event; no event is reported to InstrumentationMethods during the actual ModuleLoadFinished callback.

The bug here is that GetILFunctionBody is valid for use during ModuleLoadFinished callback and profilers on the RawProfilerHook will still receive this callback and may start instrumentation. CLRIE presents a CorProfilerInfoWrapper to raw profilers which exposes its own GetILFunctionBody that interacts with the moduleId to moduleinfo via the cached map. During ModuleLoadFinished, there is no CModuleInfo constructed yet, so raw profilers that call GetILFunctionBody will encounter a nullptr exception.

The fix is to ignore CLRIE's cache altogether and just pass the raw profiler's request to the actual profiler's GetILFunctionBody.

### DIFF
--- a/src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
+++ b/src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
@@ -364,6 +364,10 @@ HRESULT MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::GetILFunctionBo
 
     CComPtr<CModuleInfo> pModuleInfo;
     CComPtr<CMethodInfo> pMethodInfo;
+
+    IMAGE_COR_ILMETHOD* pMethodHeader = nullptr;
+    ULONG cbMethodSize = 0;
+
     if (SUCCEEDED(pAppDomainCollection->GetModuleInfoById(moduleId, (IModuleInfo**)&pModuleInfo)))
     {
         hr = pModuleInfo->GetMethodInfoByToken(methodToken, &pMethodInfo);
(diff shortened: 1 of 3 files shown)


src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
```diff
@@ -364,21 +364,29 @@ HRESULT MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::GetILFunctionBo
...
+        // This will properly manage the cache of original il bytes.
+        IfFailRet(pModuleInfo->GetMethodIl(m_pRealCorProfilerInfo, methodToken, (IMAGE_COR_ILMETHOD**)&pMethodHeader, &cbMethodSize));
+    }
+    else
+    {
+        // If this CorProfilerInfoWrapper::GetILFunctionBody() is called during OnModuleLoadFinished
+        // instead of ModuleAttachedToAssembly callback (which can happen for raw profilers), the 
+        // assembly information does not yet exist and the ModuleInfo is not yet constructed. We will
+        // not hit our cached map of ModuleId to CModuleInfo and instead call the real profiler.
+        IfFailRet(m_pRealCorProfilerInfo->GetILFunctionBody(moduleId, methodToken, (LPCBYTE*)&pMethodHeader, &cbMethodSize));
```
wiktork, line 388: Are there any other wrapped ICorProfiler methods that rely on the moduleId cache?

WilliamXieMSFT: SetILFunctionBody calls if (SUCCEEDED(pAppDomainCollection->GetModuleInfoById(moduleId, (IModuleInfo**)&pModuleInfo))) and returns S_OK if that fails. No other GetModuleInfo* calls occur in this class.

WilliamXieMSFT: All other calls using moduleId get passed to the real profiler

PR #274, opened 2020-07-16
Check if token is nil before trying to get the nested class properties.

jander-msft:
Debug version of dotnet will fail fast because CLRIE passes in a nil token when calling GetNestedClassProps in order to calculate the name of a token. This fixes the by checking if the token is nil before calling the method.

### DIFF
--- a/src/InstrumentationEngine/TokenType.cpp
+++ b/src/InstrumentationEngine/TokenType.cpp
@@ -45,7 +45,7 @@ HRESULT MicrosoftInstrumentationEngine::CTokenType::GetName(_Out_ BSTR* pbstrNam
             tstring fullName;
             vector<WCHAR> nameBuffer(100);
             ULONG cchLength = 0;
-            while ((pImport->GetNestedClassProps(tkCurrent, &tkEnclosing) == S_OK) && (tkEnclosing != mdTokenNil))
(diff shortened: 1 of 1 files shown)


WilliamXieMSFT:
>         if (TypeFromToken(m_token) == mdtTypeDef)

If I'm reading this correctly, a value of NilToken would succeed this check?

---
Refers to: src/InstrumentationEngine/TokenType.cpp:41 in e7c30b8. [](commit_id = e7c30b808f2f34799aa76fb343aff5f06401d9db, deletion_comment = False)

jander-msft:
> >         if (TypeFromToken(m_token) == mdtTypeDef)
> 
> 
> 
> If I'm reading this correctly, a value of NilToken would succeed this check?
> 
> 
> 
> ---
> 
> Refers to: src/InstrumentationEngine/TokenType.cpp:41 in e7c30b8. [](commit_id = e7c30b808f2f34799aa76fb343aff5f06401d9db, deletion_comment = False)

Yes, the TypeFromToken checks the upper 8 bits for the type of the token. Whether a token is nil or not is determined by the lower 24 bits. In the specific case that I'm fixing this, the token value is 0x02000000; its token type is TypeDef, but is also nil.

We should probably be using IsNilToken everywhere instead of == mdTokenNil. IIRC, mdNilToken is 0x00000000.

PR #280, opened 2020-08-03
Ensure AddNewExceptionClause stores the new clause

mattwarren:
Currently `AddNewExceptionClause` asserts if `ppExceptionClause` is a null pointer, but it doesn't assign anything to it.

### DIFF
--- a/src/InstrumentationEngine/ExceptionSection.cpp
+++ b/src/InstrumentationEngine/ExceptionSection.cpp
@@ -275,7 +275,7 @@ HRESULT MicrosoftInstrumentationEngine::CExceptionSection::AddNewExceptionClause
     IfFailRet(FindExceptionClauseInsertionPoint(pExceptionClause, iter));
     m_exceptionClauses.insert(iter, (IExceptionClause*)pExceptionClause);
 
-    *ppExceptionClause = pExceptionClause;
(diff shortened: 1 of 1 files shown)


src/InstrumentationEngine/ExceptionSection.cpp
```diff
@@ -275,6 +275,8 @@ HRESULT MicrosoftInstrumentationEngine::CExceptionSection::AddNewExceptionClause
     IfFailRet(FindExceptionClauseInsertionPoint(pExceptionClause, iter));
     m_exceptionClauses.insert(iter, (IExceptionClause*)pExceptionClause);
 
+    *ppExceptionClause = pExceptionClause;
```
delmyers, line 278: Use *ppExceptionClause = pExceptionClaues.Detach() instead. Otherwise, this could cause an AV. The CComPtrReferencing the CExceptionClause will decrement the ref count. If the caller also uses a CComPtr, then it will also decrement the ref count when it is destroyed, causing the exception clause stored in m_exceptionClauses to become invalid.

mattwarren: Sorry, I didn't think about the interaction with `CComPtr`, this is fixed now

delmyers:
:shipit:

PR #285, opened 2020-08-23
Fix C6385 - Buffer overrun warning

WilliamXieMSFT:
Our CI (and PR) builds run on public Microsoft-hosted agent pools.

It appears the latest VS 2019 version now reports C6385 warning: https://dev.azure.com/ms/CLRInstrumentationEngine/_build/results?buildId=102411&view=logs&j=f206dd81-de4a-56b3-b3ee-174a7e3cd854&t=ea571755-1369-573a-2879-794004f22d5b

The issue is m_dwLength is not assigned a value during Initialize() and can cause index out of range exception.

### DIFF
--- a/src/InstrumentationEngine.Lib/EnumeratorImpl.h
+++ b/src/InstrumentationEngine.Lib/EnumeratorImpl.h
@@ -46,7 +46,7 @@ namespace MicrosoftInstrumentationEngine
         {
             HRESULT hr = S_OK;
 
-            if (dwLength > 0)
+            if (dwLength >= 0)
             {
                 m_members.Attach(new CComPtr<ENUM_ITEM_TYPE>[dwLength]);
                 if (!m_members)
@@ -74,7 +74,7 @@ namespace MicrosoftInstrumentationEngine
         {
             HRESULT hr = S_OK;
 
-            if (members.size() > 0)
+            if (members.size() >= 0)
             {
                 m_members.Attach(new CComPtr<ENUM_ITEM_TYPE>[members.size()]);
                 if (!m_members)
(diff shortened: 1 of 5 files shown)


src/InstrumentationEngine.Lib/EnumeratorImpl.h
```diff
@@ -46,13 +46,14 @@ namespace MicrosoftInstrumentationEngine
         {
             HRESULT hr = S_OK;
 
-            if (dwLength != 0)
+            if (dwLength > 0)
```
jander-msft, line 49: If there are zero elements, then m_members is not initialized correctly or is it already an empty vector without calling Attach?
If the length is negative for some reason, then this method returns S_OK, which I expect it to return E_INVALIDARG

WilliamXieMSFT: m_members would be a nullptr if not iniitialized, however it is only indexed if m_dwLength > 0. With my change, negative lengths would result in m_dwLength = 0 which is would still be valid.

---
In reply to: [475785866](https://github.com/microsoft/CLRInstrumentationEngine/pull/285#discussion_r475785866) [](ancestors = 475785866)

WilliamXieMSFT: I could return E_INVALIDARG, I'll double check the callers.

---
In reply to: [475805050](https://github.com/microsoft/CLRInstrumentationEngine/pull/285#discussion_r475805050) [](ancestors = 475805050,475785866)

src/InstrumentationEngine.Lib/EnumeratorImpl.h
```diff
@@ -46,19 +46,24 @@ namespace MicrosoftInstrumentationEngine
...
```
WilliamXieMSFT, line 49: >> [](start = 25, length = 1)

`>=` otherwise 0 would result in error

PR #287, opened 2020-09-09
Create log directory if it doesn't exist

### LINKED ISSUE #157: FileLoggerSink.cpp will fail to create a file if the directory path does not exist.
FileLoggerSink.cpp uses _fwopen/_fwsopen to create the file. This [API](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019) will fail if the directory path does not exist. 

Consider adding CreateIfNotExist directory check before file open. This has an added benefit of keeping the behavior consistent with our ProductionBreakpoints InstrumentationMethods.

jander-msft:
closes #157


WilliamXieMSFT (approved):
:shipit:

src/InstrumentationEngine.Lib/FileLoggerSink.cpp
```diff
@@ -75,45 +75,55 @@ HRESULT CFileLoggerSink::Reset(_In_ LoggingFlags defaultFlags, _Out_ LoggingFlag
             wszExtension,
             _MAX_EXT);
 
-        tstring tsFile(wszFile);
-        tstring tsExtension(wszExtension);
+        // Try to create log directory if it doesn't exist
+        BOOL fDirectoryExists = PathFileExists(wszDirectory);
+        if (!fDirectoryExists)
+        {
+            fDirectoryExists = CreateDirectory(wszDirectory, nullptr);
```
wiktork, line 82: It looks like CreateDirectory can return 0, with an ERROR_ALREADY_EXISTS error code. CreateDirectory also doesn't handle intermediate paths. You probably want to check both of those error codes.

wiktork (changes requested):
Need additional handling for CreateDirectory error codes.

delmyers (approved):
:shipit:

jander-msft:
As @wiktork pointed out, there's more work here to get the issue completely fixed. I'll abandon at this time and create a new PR when ready.

PR #292, opened 2020-09-18
Fix RPH QI behavior to match CLR

WilliamXieMSFT:
CLRIE QI’s for each ICorProfilerCallback## interface separately: https://github.com/microsoft/CLRInstrumentationEngine/blob/bafaa00419bf3a9d4681c77909cb67ed438a0fc6/src/InstrumentationEngine/ProfilerManager.cpp#L200

Turns out the CLR static casts higher number ICorProfilerCallback## to lower due to interface inheritance: https://github.com/dotnet/runtime/blob/cf6b06b1d36d545e37b00bf1a6311b7fff33ff4e/src/coreclr/src/vm/eetoprofinterfaceimpl.cpp#L613

### DIFF
--- a/src/InstrumentationEngine/ProfilerManager.cpp
+++ b/src/InstrumentationEngine/ProfilerManager.cpp
@@ -196,50 +196,106 @@ HRESULT CProfilerManager::AddRawProfilerHook(
 
     unique_ptr<CProfilerCallbackHolder> profilerCallbackHolder(new CProfilerCallbackHolder);
 
-    CComPtr<ICorProfilerCallback> pCorProfilerCallback;
-    IfFailRet(pUnkProfilerCallback->QueryInterface(__uuidof(ICorProfilerCallback), (LPVOID*)&pCorProfilerCallback));
-    profilerCallbackHolder->m_CorProfilerCallback = pCorProfilerCallback;
+    // Follow the implementation set by the CLR at
+    // https://github.com/dotnet/runtime/blob/cf6b06b1d36d545e37b00bf1a6311b7fff33ff4e/src/coreclr/src/vm/eetoprofinterfaceimpl.cpp#L613
+    // where due to interface inheritance, higher numbered ICorProfilerCallback## are static-casted to lower numbered ICorProfilerCallback##
 
-    CComPtr<ICorProfilerCallback2> pCorProfilerCallback2;
-    hr = pUnkProfilerCallback->QueryInterface(__uuidof(ICorProfilerCallback2), (LPVOID*)&pCorProfilerCallback2);
-    if (SUCCEEDED(hr))
-    {
-        profilerCallbackHolder->m_CorProfilerCallback2 = pCorProfilerCallback2;
-    }
-
-    CComPtr<ICorProfilerCallback3> pCorProfilerCallback3;
-    hr = pUnkProfilerCallback->QueryInterface(__uuidof(ICorProfilerCallback3), (LPVOID*)&pCorProfilerCallback3);
-    if (SUCCEEDED(hr))
-    {
-        profilerCallbackHolder->m_CorProfilerCallback3 = pCorProfilerCallback3;
-    }
-
-    CComPtr<ICorProfilerCallback4> pCorProfilerCallback4;
-    hr = pUnkProfilerCallback->QueryInterface(__uuidof(ICorProfilerCallback4), (LPVOID*)&pCorProfilerCallback4);
-    if (SUCCEEDED(hr))
-    {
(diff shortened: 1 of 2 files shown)


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -196,50 +196,106 @@ HRESULT CProfilerManager::AddRawProfilerHook(
...
+        profilerCallbackHolder->m_CorProfilerCallback7 = pCorProfilerCallback7;
     }
 
-    CComPtr<ICorProfilerCallback3> pCorProfilerCallback3;
-    hr = pUnkProfilerCallback->QueryInterface(__uuidof(ICorProfilerCallback3), (LPVOID*)&pCorProfilerCallback3);
-    if (SUCCEEDED(hr))
+    // ICorProfilerCallback6
+    if (profilerCallbackHolder->m_CorProfilerCallback7)
+    {
+        profilerCallbackHolder->m_CorProfilerCallback6 = static_cast<CComPtr<ICorProfilerCallback6>>(profilerCallbackHolder->m_CorProfilerCallback7);
```
wiktork, line 214: Can you double check that assignment to a CComPtr from a CComPtr automatically calls AddRef?

wiktork, line 214: does `static_cast<ICorProfilerCallback6>` give you the same semantics as `static_cast<CComPtr<ICorProfilerCallback6>>`?

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -196,50 +196,106 @@ HRESULT CProfilerManager::AddRawProfilerHook(
 
     unique_ptr<CProfilerCallbackHolder> profilerCallbackHolder(new CProfilerCallbackHolder);
 
-    CComPtr<ICorProfilerCallback> pCorProfilerCallback;
-    IfFailRet(pUnkProfilerCallback->QueryInterface(__uuidof(ICorProfilerCallback), (LPVOID*)&pCorProfilerCallback));
-    profilerCallbackHolder->m_CorProfilerCallback = pCorProfilerCallback;
+    // Follow the implementation set by the CLR at
```
wiktork, line 199: Can you add additional comments regarding the fact that ICorProfiler implementations may only implement the highest interface?

wiktork (approved):

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: spoke offline, I'll add more details around this. Due to interface inheritance, raw profilers QI can just return the highest supported version but would still contain implementation for all lower versioned.

---
In reply to: [491181036](https://github.com/microsoft/CLRInstrumentationEngine/pull/292#discussion_r491181036) [](ancestors = 491181036)

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Confirmed in a separate ConsoleApp using refCount and as well, docs around the CComPtr::operator= 

---
In reply to: [491177570](https://github.com/microsoft/CLRInstrumentationEngine/pull/292#discussion_r491177570) [](ancestors = 491177570)

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -196,50 +196,106 @@ HRESULT CProfilerManager::AddRawProfilerHook(
...
-    if (SUCCEEDED(hr))
+    // ICorProfilerCallback5
+    if (profilerCallbackHolder->m_CorProfilerCallback6)
     {
-        profilerCallbackHolder->m_CorProfilerCallback4 = pCorProfilerCallback4;
+        profilerCallbackHolder->m_CorProfilerCallback5 = static_cast<CComPtr<ICorProfilerCallback5>>(profilerCallbackHolder->m_CorProfilerCallback6);
+    }
+    else
+    {
+        CComPtr<ICorProfilerCallback6> pCorProfilerCallback5;
```
WilliamXieMSFT, line 233: >6 [](start = 36, length = 1)

oops

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: `static_cast<ICorProfilerCallback*>` seems to work as well

---
In reply to: [491178672](https://github.com/microsoft/CLRInstrumentationEngine/pull/292#discussion_r491178672) [](ancestors = 491178672)

PR #306, opened 2020-10-09
Get code analysis builds to work.

delmyers:
1) Removed InstrumentationEngine.Api from the release build
   so that the .h/.cpp files will not be deleted when the
   solution is cleaned. This is inline with our documentation
   which states that the project must be built manually.
2) Upgraded all of the current code analysis yaml tasks to the latest
3) Removed fxcop in favor of roslyn analyzers
4) Added analyzers (roslyn csharp, prefast) that must run
    in their own build stage
5) Added build step for Semmle C++, but disabled it because
    it seems to not be working (I'll get the semmle team to look
    at it).
6) Reduced the number of binaries that are scanned by apiscan
7) Stopped calls to TSA. Break the build instead on error.
8) Fixed C++ builds so that they are compatible with ApiScan
   - Enabled /profile link option in all release builds
   - Set /incremental:no for all realease builds
9) Fixed a couple of prefast/apiscan/roslyn code analysis failures.
   - Enabled control flow guard in release builds
     (excluded in debug builds to enable edit and continue)
10) Added CodeAnalysis.Managed and CodeAnalysis.Native
    configurations to the InstrumentationEngine.sln so that
    we have code analysis builds that will build either CS or C++
    code but not both for the purpose of code analysis.
    Previously, the regular x86/x64 builds build some CS
    test code that should not be part of the C++ code
    analysis.

Note: this only gets the builds running. I will work on a separate
task to file bugs when code analysis fails.

These builds are not executed as part of the PR, but there
is a CI trigger that should run them after every check in.


build/yaml/jobs/codeanalysis/buildanalyzer.yaml
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Job Template: Run Roslyn Analyzers
+
+parameters:
+  SignType: None
```
WilliamXieMSFT, line 7: >SignType: None [](start = 2, length = 14)

not used

build/yaml/steps/windows/solutionbuild.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Step Template: Build Windows Binaries
```
WilliamXieMSFT, line 4: >Build Windows Binaries [](start = 17, length = 22)

This looks very similar to binaries.yaml, just without nuget restore. Can we consolidate the two?

build/Unmanaged.props
```diff
@@ -1,234 +1,233 @@
...
+    <UseDebugLibraries>false</UseDebugLibraries>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration" Condition="'$(Configuration)'=='Debug'">
+    <UseDebugLibraries>true</UseDebugLibraries>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)'=='Debug'">
+    <LinkIncremental>true</LinkIncremental>
+  </PropertyGroup>
+    <PropertyGroup Condition="'$(Configuration)'=='Release'">
```
WilliamXieMSFT, line 66: >   [](start = 2, length = 2)

nit indentation

WilliamXieMSFT (approved):
:shipit:

Issue #307, opened 2020-10-14
Stack overflow when method has a lot of instructions


jakubch1:
We found out that methods with a lot of instructions cause stack overflow just after instrumentation. From my investigation issue happens just after method is instrumented (dumps logs contains instrumented method) but before method is invoked (no code coverage for method).

To reproduce just create method with a lot of blocks, for example 1000 and don't do any instrumentation, just set method to be instrumented.

Unhandled exception at 0x00007FF88D1F5CB7 (MicrosoftInstrumentationEngine_x64.dll) in testhost_19552_20201014T133532_crashdump.dmp: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x000000B37C103FD8).

Stack (not full):
[Inline Frame] MicrosoftInstrumentationEngine_x64.dll!MicrosoftInstrumentationEngine::CRefCount::Release() Line 121
	at C:\git\clrie\src\InstrumentationEngine.Lib\refcount.h(121)
MicrosoftInstrumentationEngine_x64.dll!MicrosoftInstrumentationEngine::CEnumerator<IEnumAppDomainInfo,IAppDomainInfo>::Release() Line 32
	at C:\git\clrie\src\InstrumentationEngine.Lib\EnumeratorImpl.h(32)
[External Code]
[Inline Frame] MicrosoftInstrumentationEngine_x64.dll!MicrosoftInstrumentationEngine::CRefCount::Release() Line 121
	at C:\git\clrie\src\InstrumentationEngine.Lib\refcount.h(121)
MicrosoftInstrumentationEngine_x64.dll!MicrosoftInstrumentationEngine::CEnumerator<IEnumAppDomainInfo,IAppDomainInfo>::Release() Line 32
	at C:\git\clrie\src\InstrumentationEngine.Lib\EnumeratorImpl.h(32)
[External Code]
[Inline Frame] MicrosoftInstrumentationEngine_x64.dll!MicrosoftInstrumentationEngine::CRefCount::Release() Line 121
	at C:\git\clrie\src\InstrumentationEngine.Lib\refcount.h(121)
MicrosoftInstrumentationEngine_x64.dll!MicrosoftInstrumentationEngine::CEnumerator<IEnumAppDomainInfo,IAppDomainInfo>::Release() Line 32
	at C:\git\clrie\src\InstrumentationEngine.Lib\EnumeratorImpl.h(32)
[External Code]
[...]

PR #308, opened 2020-10-16
Fix stack overflow when destructing graph

jakubch1:
This commit fixes stack overflow when destructing graph with many instructions #307 

Currently when graph is deconstructed we have:
1) Removing "Previous" poiners from all instructions
2) Release all pointers in graph for example first and last instructions. This causes stack overflow as it deconstruct all instructions:
```
Instr1.Release() [deconstruct];
    -> Instr2.Release() [deconstruct];
        -> Instr3.Release() [deconstruct];
             -> Instr4.Release() [deconstruct];
                      ..............
```

This PR changes logic so we have:
1) Release all pointers in graph to first instruction.
2) Removing "Previous" pointers from all instructions, which causes that instructions are deconstructed not in cascade mode:

```
Instr1.Disconnect();
Instr2.Disconnect();
    -> Instr1.Release() [deconstruct];
Instr3.Disconnect();
    -> Instr2.Release() [deconstruct];
Instr4.Disconnect();
    -> Instr3.Release() [deconstruct];
.......
```

### DIFF
--- a/src/InstrumentationEngine/InstructionGraph.cpp
+++ b/src/InstrumentationEngine/InstructionGraph.cpp
@@ -24,28 +24,28 @@ Initial state:
 |========| <--- |========| <--- |========| <--- |========|
 
 After node 1 disconnect:
-|=node 1=| ---> |=node 2=| ---> |=node 3=| ---> |=node 4=|
+|=node 1=|      |=node 2=| ---> |=node 3=| ---> |=node 4=|
 |        |      |        |      |        |      |        |
 | ref: 2 |      | ref: 1 |      | ref: 2 |      | ref: 2 |
 |========| <--- |========| <--- |========| <--- |========|
 
 After node 2 disconnect:
-|=node 1=| ---> |=node 2=| ---> |=node 3=| ---> |=node 4=|
+|=node 1=|      |=node 2=|      |=node 3=| ---> |=node 4=|
 |        |      |        |      |        |      |        |
 | ref: 1 |      | ref: 1 |      | ref: 1 |      | ref: 2 |
-|========| <--- |========| <--- |========| <--- |========|
+|========|      |========| <--- |========| <--- |========|
 
 After node 3 disconnect:
-|=node 1=| ---> |=node 2=| ---> |=node 3=| ---> |=node 4=|
+|=node 1=|      |=node 2=|      |=node 3=|      |=node 4=|
 |        |      |        |      |        |      |        |
 | ref: 1 |      | ref: 0 |      | ref: 1 |      | ref: 1 |
-|========| <--- |========| <--- |========| <--- |========|
+|========|      |========|      |========| <--- |========|
 
 After node 4 disconnect:
-|=node 1=| ---> |=node 2=| ---> |=node 3=| ---> |=node 4=|
+|=node 1=|      |=node 2=|      |=node 3=|      |=node 4=|
 |        |      |        |      |        |      |        |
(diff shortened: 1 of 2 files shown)


src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -17,6 +17,15 @@ MicrosoftInstrumentationEngine::CInstructionGraph::~CInstructionGraph()
         CCriticalSectionHolder lock(&m_cs);
 
         CComPtr<CInstruction> pInstr = (CInstruction*)m_pFirstInstruction;
+
+        m_pFirstInstruction.Release();
```
wiktork, line 21: Can you add some comments here, describing in more details why the order is important?

wiktork: This might be best described In terms of reference counts:

Initial ref counts:
2<=>2<=>2
After firstInstruction.Release
1<=>2<=>2
After first Disconnect
1<=>2

wiktork:
Thank you for making this fix, I believe this will also address the issue of stack overflows in Debug mode.

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -17,6 +17,15 @@ MicrosoftInstrumentationEngine::CInstructionGraph::~CInstructionGraph()
...
+        m_pOrigFirstInstruction.Release();
+        
+        if(m_pUninstrumentedFirstInstruction != nullptr)
+        {
+            m_pUninstrumentedFirstInstruction.Release();
+        }
+
         while (pInstr != nullptr)
```
wiktork, line 29: You can verify if you have any leaking instruction objects by using Debug mode bits of clrie and having a debugger attached during shutdown. See [refcount.cpp](https://github.com/microsoft/CLRInstrumentationEngine/blob/develop/src/InstrumentationEngine.Lib/refcount.cpp)

wiktork (approved):

delmyers (approved):
:shipit:

src/InstrumentationEngine/InstructionGraph.cpp
delmyers: I actually find this kind of confusing. I think the description of the PR is useful. Maybe we can just put that in?

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -17,6 +17,15 @@ MicrosoftInstrumentationEngine::CInstructionGraph::~CInstructionGraph()
...
 
         CComPtr<CInstruction> pInstr = (CInstruction*)m_pFirstInstruction;
+
+        m_pFirstInstruction.Release();
+        m_pOrigFirstInstruction.Release();
+        
+        if(m_pUninstrumentedFirstInstruction != nullptr)
+        {
+            m_pUninstrumentedFirstInstruction.Release();
+        }
```
delmyers, line 27: You could also just do the following:

m_pFirstInstruction = nullptr;
m_OirigFirstInstruction = nullptr;
m_pUninstrumentedFirst = nullptr;

That way, you don't need to check m_pUninstrumentedFirstInstruction to see if it is null.

jakubch1:
I've added comment but also changed logic so now when disconnecting we remove both forward and backward links. I think this is more safe change as order doesn't matter any more and also it's more secure if pointer to first instruction is kept somewhere else. Debugged the change and it looks good => Disconnect of n-th instruction is destructing (n-1)th instruction

PR #317, opened 2020-10-22
Reenable ref counting on debug.

delmyers:


src/InstrumentationEngine/dllmain.cpp
```diff
@@ -27,11 +27,12 @@ BOOL APIENTRY DllMain(HINSTANCE hInstance, DWORD dwReason, LPVOID lpReserved)
             _CRTDBG_LEAK_CHECK_DF // dump leak report at shutdown
         );
 #endif
-        //INITIALIZE_REF_RECORDER(MicrosoftInstrumentationEngine::CRefCount::EnableRecorder);
+        INITIALIZE_REF_RECORDER(MicrosoftInstrumentationEngine::CRefCount::EnableRecorder);
```
WilliamXieMSFT, line 30: >INITIALIZE_REF_RECORDER [](start = 8, length = 23)

Should this be inside the IfDef?

wiktork, line 30: How bad is the impact in debug mode when you turn this on?

wiktork (approved):

src/InstrumentationEngine/dllmain.cpp
delmyers: It depends. It is pretty well eclipsed by having debug on at all. For example, one of my tests goes from running in 9 seconds in release to running in 2.5 minutes in debug. This adds another few seconds I think. But I could double check.

src/InstrumentationEngine.Lib/refcount.h
```diff
@@ -147,10 +147,42 @@ namespace MicrosoftInstrumentationEngine
         static class CRefRBMap g_map;
         static CRITICAL_SECTION g_crst;
 
-        static void RecordCreation(CRefCount *pCRefCount);
-        static void RecordDestruction(CRefCount *pCRefCount);
-        static void RecordAddRef(CRefCount *pCRefCount);
-        static void RecordRelease(CRefCount *pCRefCount);
+        static void __forceinline RecordCreation(CRefCount* pCRefCount)
+        {
+            if (g_fRecordingInitialized)
```
wiktork, line 152: I'm assuming this is still wrapped in an ifdef DEBUG? Also does the compiler honor __forceinline in Debug builds?

delmyers: Yes, in the caller.

It seems a little faster __forceinline, but the callstack doesn't indicate that it is inlined. It is hard to tell. I'll also try creating a macro.

delmyers: OK, so I wrapped the functions in a macro instead, and it seems to save about 1 second of time for the InstrEngineTests overall when recording is turned off.

After more testing, it does turn out that having recording on causes a relatively small overhead. The InstrEngine tests run about 1/2 second slower in total when recording is turned on. That being said, these tests are pretty small and create/destroy a relatively small number of objects. When we talked offline, I thought that the tests had a much larger overhead because I had accidentally had a debugger attached to one of them... oops.

Anyway, I'm not going to try to squeeze any more out of this, it is probably as good as I can do.

src/InstrumentationEngine.Lib/refcount.cpp
```diff
@@ -129,82 +125,76 @@ namespace MicrosoftInstrumentationEngine
...
 
-            LeaveCriticalSection(&g_crst);
-        }
+        LeaveCriticalSection(&g_crst);
     }
 
     void CRefCount::RecordDestruction(CRefCount* pCRefCount)
     {
-        if (g_fRecordingInitialized)
+        if (pCRefCount->m_ulcRef <= 0)
```
wiktork, line 144: Did we actually hit this condition? That suggests a pretty bad bug.

delmyers: I don't know. It was probably in there when it was first written. 

I'll fold this back into the caller so that it is more similar to how it was before, but this check us used everywhere the CModuleRefCount pattern is used.

wiktork (approved):

Issue #311, opened 2020-10-20
Improve type safety of internal classes


delmyers:
There are a number of places where we cast from a base class to a derived class in our internal code, which could cause problems if a caller passes us an instance of an interface that we did not create internally. I suggest we:

1) Update our code to use virtual inheritance (instead of multiple inheritance). This will help us catch locations where we cast from a base class.

2) QI for the derived class. This will give us type safety.

PR #325, opened 2020-11-06
Fix static analysis errors

clguiman:
Most of the changes are minor to fix Roslyn analyzer and preFAST warnings. From my point of view the biggest change is in InstructionGraph, where multiple methods take `IInstruction*` arguments and immediately down cast to `CInstruction`. Wherever I could, I updated the internal field type to `IInstruction`, but in certain cases I had to QI to `CInstruction`.


inc/clr/prof/corprof.h
```diff
@@ -1,4 +1,6 @@
```
wiktork, line 1: This file is pulled from the clr repo, we may want to just supress these warnings. The next time we updated it, we'll have to go through this again.

clguiman: I had to fix the SAL annotations in this file because otherwise preFAST was complaining the overridden methods in the derived classes had mismatched annotations. I had three options: disable SAL annotation checks, remove the annotations in the derived classes or add them in corprof.h
I did find some wrong SAL annotations, so I guess disabling the warning wouldn't be the best thing.

delmyers: SAL warnings should just be suppressed for this file. I think we do something similar in the aerobee repository. Maybe we can have a chat about how to fix this.

build/yaml/steps/codeanalysis/prefast.yaml
```diff
@@ -8,6 +8,10 @@ steps:
 - task: ms-vseng.MicroBuildShipTasks.9ef02907-718e-41b7-a57e-b10f86130bb8.MicroBuildPrefastPlugin@2
   displayName: 'Install Prefast Plugin'
 
+- powershell: |
+    Write-Host "##vso[task.setvariable variable=_PREFAST_CYCLOMATIC;] 83"
+  displayName: Set _PREFAST_CYCLOMATIC=83
```
delmyers, line 13: Can you add comment about what this does?

inc/clr/prof/corprof.h
delmyers: In the aerobee repository, we added a separate file called corprofiler.h:

```
#pragma once
#pragma warning(push) // corprof.h from the coreclr does not include SAL anotations
#pragma warning(disable : 6313 25057 25155 28718 28722 28723 28729 28740)
#include "corprof.h"
#pragma warning(pop)
```

We never reference corprof.h directly, instead we include "corprofiler.h" I would suggest doing the same here.


---
In reply to: [520710879](https://github.com/microsoft/CLRInstrumentationEngine/pull/325#discussion_r520710879) [](ancestors = 520710879)

src/Common.Lib/CriticalSectionHolder.h
```diff
@@ -50,11 +50,11 @@ namespace CommonLib
...
 
     --*/
-    inline BOOL IsCriticalSectionOwned(
-        _In_ CRITICAL_SECTION* lpCriticalSection
+    inline bool IsCriticalSectionOwned(
+        _In_ const CRITICAL_SECTION* lpCriticalSection
     )
     {
-        BOOL fOwned = FALSE;
+        bool fOwned = false;
```
delmyers, line 57: this variable isn't really needed.

src/Extensions.Base/HostExtension.h
```diff
@@ -30,8 +30,8 @@ class ATL_NO_VTABLE CHostExtension :
 
 	DECLARE_PROTECT_FINAL_CONSTRUCT()
 
-	HRESULT FinalConstruct() { return S_OK; }
-	void FinalRelease() {}
+	static HRESULT FinalConstruct() { return S_OK; }
+	static void FinalRelease() {}
```
delmyers, line 34: Make sure that this isn't part of a public interface. It looks like it might be. If yes, just disable the warnings for this file.

src/ExtensionsCommon/DataItemContainer.h
```diff
@@ -47,12 +47,12 @@ class ATL_NO_VTABLE CDataItemContainer :
 		return  this->m_dataItem;
 	}
 
-	HRESULT FinalConstruct()
+	static HRESULT FinalConstruct()
```
delmyers, line 50: >static HRESULT FinalConstruct() [](start = 1, length = 31)

same comment here. Make sure that this isn't a public interface. It doesn't look like it is, otherwise it should override, and be stdcall, but that might be a mistake, too.

src/InstrumentationEngine/AppDomainCollection.h
```diff
@@ -55,7 +55,7 @@ namespace MicrosoftInstrumentationEngine
         // NOTE: These helpers will search all appdomains
         STDMETHOD(GetAssemblyInfoById)(_In_ AssemblyID assemblyID, _Out_ IAssemblyInfo** ppAssemblyInfo) override;
         STDMETHOD(GetModuleInfoById)(_In_ ModuleID moduleID, _Out_ IModuleInfo** ppModuleInfo) override;
-        STDMETHOD(GetModuleInfosByMvid)(_In_ GUID mvid, _Out_ IEnumModuleInfo** ppEnum) override;
+        STDMETHOD(GetModuleInfosByMvid)(GUID mvid, _Out_opt_ IEnumModuleInfo** ppEnum) override;
```
delmyers, line 58: >GUID mvid [](start = 40, length = 9)

Should this be, _In_ GUID& instead? or _In_ REFGUID?

src/InstrumentationEngine.Api/InstrumentationEngine.h
```diff
@@ -7433,7 +7433,7 @@ EXTERN_C const IID IID_IInstructionFactory;
         
         virtual HRESULT STDMETHODCALLTYPE DecodeInstructionByteStream( 
             /* [in] */ DWORD cbMethod,
-            /* [in] */ __RPC__in LPCBYTE instructionBytes,
+            /* [size_is][in] */ __RPC__in_ecount_full(cbMethod) LPCBYTE instructionBytes,
```
delmyers, line 7436: >__RPC__in_ecount_full(cbMethod) [](start = 32, length = 31)

I don't see a corresponding change to the idl.

src/InstrumentationEngine/AppDomainCollection.h
delmyers: Ugh!!! but that will be a change to the interface. That was a bad mistake in the interface definition :-(. Do we need to udate the idl to fix this?

---
In reply to: [520723341](https://github.com/microsoft/CLRInstrumentationEngine/pull/325#discussion_r520723341) [](ancestors = 520723341)

src/InstrumentationEngine/CorHeaders.h
```diff
@@ -178,10 +178,10 @@ typedef struct tagCOR_ILMETHOD_TINY : IMAGE_COR_ILMETHOD_TINY
 {
     bool     IsTiny () const         { return (Flags_CodeSize & (CorILMethod_FormatMask >> 1)) == CorILMethod_TinyFormat; }
     unsigned GetCodeSize () const    { return ((unsigned) Flags_CodeSize) >> (CorILMethod_FormatShift-1); }
-    unsigned GetMaxStack () const    { return 8; }
+    static unsigned GetMaxStack ()   { return 8; }
```
delmyers, line 181: >static unsigned GetMaxStack ()   { return 8; } [](start = 4, length = 46)

I would suggest just disabling the "function could be static" warning. There are often reasons that functions are not static when they could be.

src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
```diff
@@ -160,13 +160,14 @@ ULONG MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::AddRef()
 
 ULONG MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::Release()
 {
-    if (InterlockedDecrement(&m_refcount) == 0)
```
delmyers, line 163: >InterlockedDecrement(&m_refcount) == 0 [](start = 8, length = 38)

what was the error here?

src/InstrumentationEngine/Instruction.cpp
```diff
@@ -522,17 +522,18 @@ HRESULT MicrosoftInstrumentationEngine::CInstruction::GetIsRemoved(_Out_ BOOL* p
     return S_OK;
 }
 
-HRESULT MicrosoftInstrumentationEngine::CInstruction::GetInstructionSize(_Out_ DWORD* pdwSize)
+HRESULT MicrosoftInstrumentationEngine::CInstruction::GetInstructionSize(_In_ IInstruction* pInstruction, _Out_ DWORD* pdwSize)
```
delmyers, line 525: >GetInstructionSize(_In_ IInstruction* pInstruction, _Out_ DWORD* pdwSize) [](start = 54, length = 73)

I don't think that I understand the need for this. It is adding an additional if check for no reason.

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -404,14 +404,14 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::EncodeIL(
     CAutoVectorPtr<BYTE> pILArray;
     if (cbBuffer > 0)
     {
-        pILArray.Attach(new BYTE[cbBuffer]);
+        pILArray.Attach(new BYTE[cbBuffer]());
```
delmyers, line 407: >() [](start = 42, length = 2)

?

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -404,14 +404,14 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::EncodeIL(
...
     }
 
     // Allocate the il map.
     CAutoVectorPtr<COR_IL_MAP> pCorILMap;
     if(cCorILMap > 0)
     {
-        pCorILMap.Attach(new COR_IL_MAP[cCorILMap]);
+        pCorILMap.Attach(new COR_IL_MAP[cCorILMap]());
```
delmyers, line 414: >() [](start = 50, length = 2)

?

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -811,8 +785,10 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::InsertBefore(_In_ IIn
     IfNullRetPointer(pInstructionOrig);
     IfNullRetPointer(pInstructionNew);
 
-    CInstruction* pInstrOrig = (CInstruction*)pInstructionOrig;
-    CInstruction* pInstrNew = (CInstruction*)pInstructionNew;
+    CComPtr<CInstruction> pInstrOrig;
+    IfFailRet(pInstructionOrig->QueryInterface(&pInstrOrig));
```
delmyers, line 789: >IfFailRet(pInstructionOrig->QueryInterface(&pInstrOrig)); [](start = 4, length = 57)

These QI's will always fail because QI is not implemented for CInstruction. It currently can't be because of diamond inheritance. See #311

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -845,16 +821,22 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::InsertBefore(_In_ IIn
...
     CCriticalSectionHolder lock(&m_cs);
 
     IfNullRetPointer(pInstructionNew);
 
-    CInstruction* pInstrOrig = (CInstruction*)pInstructionOrig;
-    CInstruction* pInstrNew = (CInstruction*)pInstructionNew;
+    CComPtr<CInstruction> pInstrOrig;
+    if (pInstructionOrig != nullptr)
+    {
+        IfFailRet(pInstructionOrig->QueryInterface(&pInstrOrig));
```
delmyers, line 835: >pInstructionOrig->QueryInterface(&pInstrOrig) [](start = 18, length = 45)

Again, these QIs will fail.

src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
clguiman: I don't remember the exact error, but it was complaining that m_refcount might have changed from the time the atomic operation was performed

src/InstrumentationEngine.Attach/CommandHandler/AttachCommandHandler.cs
```diff
@@ -233,9 +233,12 @@ internal class AttachCommandHandler
             }
 
             XmlSerializer engineConfigSerializer = new XmlSerializer(typeof(InstrumentationEngineConfiguration));
-            MemoryStream memStream = new MemoryStream();
-            engineConfigSerializer.Serialize(memStream, configuration);
-            byte[] bytes = memStream.ToArray();
+            byte[] bytes;
```
delmyers, line 236: >byte[] bytes; [](start = 12, length = 13)

?

tests/ApplicationInsightsCompatibility/Intercept.Shared.Tests/TestEngine.cs
```diff
@@ -23,16 +23,14 @@ class TestEngine : TestEngineBase
...
-        private const string RawProfilerHookComponentId = "{4BCBE156-37EC-4179-B051-0183C57BACF9}";
 
         private TestEngine()
         {
         }
 
-        public static ITestResult ExecuteTest<T>()
+        public static ITestResult ExecuteTest<T>() where T : new()
         {
+            var _ = new T(); // instantiate the test class so the static analyzer doesn't think it's never used
```
delmyers, line 33: >var _ = new T(); // instantiate the test class so the static analyzer doesn't think it's never used [](start = 12, length = 99)

This is strange. I don't think I understand what the static analyzer is complaining about here. I don't think that we should be instantiating something here. It looks like a bug in the static analyzer. Can we disable the warning?

tests/InstrEngineTests/InstrEngineTests/NativeMethods.cs
```diff
@@ -11,6 +13,6 @@ internal class NativeMethods
         public static extern int GetComPlusPackageInstallStatus();
 
         [DllImport("kernel32.dll")]
-        public static extern int SetComPlusPackageInstallStatus(int flag);
+        public static extern bool SetComPlusPackageInstallStatus(int flag);
```
delmyers, line 16: >bool [](start = 29, length = 4)

?

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
```diff
@@ -263,11 +263,23 @@ public static void DiffResultToBaseline(string output, string baseline, bool reg
...
-                        XmlDocument baselineDocument = new XmlDocument();
-                        baselineDocument.LoadXml(baselineXmlDocStr);
-
-                        XmlDocument outputDocument = new XmlDocument();
-                        outputDocument.LoadXml(outputXmlDocStr);
+                        // https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca3075#solution-3
+                        XmlDocument baselineDocument = new XmlDocument() { XmlResolver = null };
+                        using (var baselineStringReader = new StringReader(baselineXmlDocStr))
+                        {
+                            using (var xmlReader = XmlReader.Create(baselineStringReader, new XmlReaderSettings() { XmlResolver = null }))
```
delmyers, line 270: > XmlReader.Create(baselineStringReader, new XmlReaderSettings() { XmlResolver = null } [](start = 50, length = 86)

Maybe we should add a helper function?

tests/RawProfilerHook/RawProfilerHook.Tests/TestEngine/TestEngine.cs
```diff
@@ -48,8 +48,9 @@ public class TestEngine : RemoteUnitTestExecutor.TestEngineBase
 
         private readonly string traceFilePath;
 
-        public static ITestResult ExecuteTest<T>()
+        public static ITestResult ExecuteTest<T>() where T : new()
         {
+            var _ = new T(); // instantiate the attribute class so the static analyzer doesn't think it's never used
```
delmyers, line 53: >var _ = new T() [](start = 12, length = 15)

same comment as in the other file.

delmyers (changes requested):
:clock1:

src/InstrumentationEngine/InstructionGraph.cpp
clguiman: it zero-initializes the array

delmyers: The typical way would be to use curly braces, I think.

src/InstrumentationEngine/Instruction.cpp
clguiman: While refactoring ```CExceptionClause::RenderExceptionClause()``` to only operate on ```IInstruction``` I noticed the only required method from ```CInstruction``` that's not part of ```IInstruction``` interface was ```CInstruction::GetInstructionSize``` so I added this public static method.

tests/InstrEngineTests/InstrEngineTests/NativeMethods.cs
clguiman: ```SetComPlusPackageInstallStatus``` returns a BOOL so to not confuse the return type with HRESULT I updated the signature

src/InstrumentationEngine.Attach/CommandHandler/AttachCommandHandler.cs
clguiman: I didn't want to extend the MemoryStream IDisposable scope until the end of the method, so I declared the bytes array like this.

src/InstrumentationEngine/InstructionGraph.cpp
clguiman: I updated ```CInstruction::QueryInterface```.

tests/ApplicationInsightsCompatibility/Intercept.Shared.Tests/TestEngine.cs
clguiman: This was the error: ``` tests/ApplicationInsightsCompatibility/Intercept.Shared.Tests/InterceptDecorateTests.cs(19,23): Warning CA1812: Avoid uninstantiated internal classes: InterceptDecorateTests.SimpleDecorate is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it static (Shared in Visual Basic).```

I noticed I can make it go away just with the ```new()``` constraint, so I'll do that

src/InstrumentationEngine/Instruction.h
```diff
@@ -64,6 +64,17 @@ namespace MicrosoftInstrumentationEngine
         DEFINE_DELEGATED_REFCOUNT_RELEASE(CInstruction);
         STDMETHOD(QueryInterface)(_In_ REFIID riid, _Out_ void **ppvObject) override
         {
+            // Workaround for the diamond inheritence problem.
```
clguiman, line 67: This is not enough, I have to add similar support in derived classes

src/InstrumentationEngine/AppDomainCollection.h
clguiman: Yes, I guess we won't be able to change it without breaking compatibility .

src/InstrumentationEngine.Api/InstrumentationEngine.h
clguiman: I've updated the idl file

src/InstrumentationEngine/AppDomainCollection.h
delmyers: We don't want to break compatibility, for sure. Oh well, this is unfortunate.

---
In reply to: [521739359](https://github.com/microsoft/CLRInstrumentationEngine/pull/325#discussion_r521739359) [](ancestors = 521739359)

src/InstrumentationEngine/Instruction.cpp
delmyers: GetInstructionSize is called a lot in tight loops when decoding methods. Adding any additional overhead to it might cause slow down. I'm OK with this, but I might have to change it in my perf updates if it takes too long.

---
In reply to: [520888584](https://github.com/microsoft/CLRInstrumentationEngine/pull/325#discussion_r520888584) [](ancestors = 520888584)

src/InstrumentationEngine/Instruction.h
```diff
@@ -137,18 +138,40 @@ namespace MicrosoftInstrumentationEngine
...
         constexpr CInstruction* OriginalNextInstructionInternal() { return m_pOriginalNextInstruction.p; }
         constexpr CInstruction* OriginalPreviousInstructionInternal() { return m_pOriginalPreviousInstruction.p; }
 
+    protected:
+        template<typename... Args>
+        HRESULT QueryInterfaceInternal(_In_ REFIID riid, _Out_ void** ppvObject, Args... pThisArgs)
+        {
+            // Workaround for the diamond inheritence problem.
+            // Both CDataContainer and IInstruction inherit from IUnknown and
+            //    ImplQueryInterface() doesn't know which one to pick when casting CInstruction* to IUnknow*.
```
delmyers, line 159: >IUnknow [](start = 100, length = 7)

typo

src/InstrumentationEngine/Instruction.cpp
clguiman: The compiler should be able to inline the indirection, but if it has a perf impact we'll figure out another solution

src/InstrumentationEngine/Instruction.h
```diff
@@ -137,18 +138,40 @@ namespace MicrosoftInstrumentationEngine
...
+        HRESULT LogInstruction(bool ignoreTest);
 
         constexpr CInstruction* NextInstructionInternal() { return m_pNextInstruction.p; }
         constexpr CInstruction* PreviousInstructionInternal() { return m_pPreviousInstruction.p; }
         constexpr CInstruction* OriginalNextInstructionInternal() { return m_pOriginalNextInstruction.p; }
         constexpr CInstruction* OriginalPreviousInstructionInternal() { return m_pOriginalPreviousInstruction.p; }
 
+    protected:
+        template<typename... Args>
+        HRESULT QueryInterfaceInternal(_In_ REFIID riid, _Out_ void** ppvObject, Args... pThisArgs)
```
delmyers, line 155: >QueryInterfaceInternal [](start = 16, length = 22)

This function seems not necessary. You could just put the if check inside QueryInterface, then you wouldn't have to forward the args template.

build/CommonBuild.Private.props
```diff
@@ -58,4 +58,5 @@
     <PropertyGroup Label="Configuration">
         <SpectreMitigation>Spectre</SpectreMitigation>
     </PropertyGroup>
+    <Import Project =".\CommonBuild.Cpp.props" Condition="'$(MSBuildProjectExtension)'=='.vcxproj'"/>
```
delmyers, line 61: ><Import Project =".\CommonBuild.Cpp.props" Condition="'$(MSBuildProjectExtension)'=='.vcxproj'"/> [](start = 4, length = 97)

There is already a ClrInstrumentationEngine.Cpp.props that is meant to be included by vcxproj files.

build/CommonBuild.Cpp.props
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Condition="'$(VSBuild_Prefast)'=='true'">
+    <Microbuild_Prefast>true</Microbuild_Prefast>
+  </PropertyGroup>
+  <Import Project=".\CommonBuild.Prefast.props" Condition="'$(Microbuild_Prefast)'=='true'"/>
```
delmyers, line 6: ><Import Project=".\CommonBuild.Prefast.props" Condition="'$(Microbuild_Prefast)'=='true'"/> [](start = 2, length = 91)

Could this be added to ClrInstrumentation.cpp.props?

tests/InstrumentationEngine.Lib.Tests/InstrumentationEngine.Lib.Tests.vcxproj
```diff
@@ -78,6 +78,7 @@
       <PreprocessorDefinitions Condition="'$(Platform)'=='Win32'">X86;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <UseFullPaths>true</UseFullPaths>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+      <ControlFlowGuard>Guard</ControlFlowGuard>
```
delmyers, line 81: ><ControlFlowGuard>Guard</ControlFlowGuard> [](start = 6, length = 42)

Import ClrInstrumentationEngine.cpp.props to get this.

tests/InstrumentationEngine.ProfilerProxy.Tests/InstrumentationEngine.ProfilerProxy.Tests.vcxproj
```diff
@@ -78,6 +78,7 @@
       <PreprocessorDefinitions Condition="'$(Platform)'=='Win32'">X86;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <UseFullPaths>true</UseFullPaths>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+      <ControlFlowGuard>Guard</ControlFlowGuard>
```
delmyers, line 81: > <ControlFlowGuard>Guard</ControlFlowGuard> [](start = 5, length = 43)

Import ClrInstrumentationEngine.cpp.props

src/ExtensionsCommon/DecorationCallbackInfo.cpp
```diff
@@ -9,7 +9,7 @@
 CDecorationCallbackInfo::CDecorationCallbackInfo(
 	const std::wstring& strMethodName,
 	DWORD dwAttributes,
-	PCCOR_SIGNATURE pCorSig,
+    _In_reads_(ulCorSigSize) PCCOR_SIGNATURE pCorSig,
```
WilliamXieMSFT, line 12: >     [](start = 0, length = 4)

nit mixing tabs/spaces

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -52,7 +52,7 @@ CProfilerManager::CProfilerManager() :
         tstringstream mboxStream;
         DWORD pid = GetCurrentProcessId();
         mboxStream << _T("MicrosoftInstrumentationEngine ProfilerAttach. PID: ") << pid;
-        MessageBox(NULL, mboxStream.str().c_str(), L"", MB_OK);
```
WilliamXieMSFT, line 55: >MessageBox [](start = 8, length = 10)

Interesting, I'd assume this would have redirected to MessageBoxA or MessageBoxW internally

clguiman: it does, there's a define statement in a Windows header file that picks the right one, but prefast was complaining - it wanted the actual exported name.

build/yaml/jobs/codeanalysis/apiscan.yaml
```diff
@@ -44,15 +48,21 @@ jobs:
...
 
+  # Patch the binary version in the exclusion file.
+  - powershell: |
+      (Get-Content $(APIScanExclusionFilePath)).Replace("[version]", "$(Build.BuildNumber)") | Set-Content $(APIScanExclusionFilePath)
+    displayName: Set binary version in APIScan exclusion list
+
   # Run API Scan
 
-  - task: securedevelopmentteam.vss-secure-development-tools.build-task-apiscan.APIScan@2
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-apiscan.APIScan@1
```
clguiman, line 58: APIScan@2 is fails when providing an exclusion list, that's why I switched to APIScan@1. I'm trying to get help from the task maintainers and see if I can switch back to the higher version.

tests/InstrumentationEngine.Lib.Tests/InstrumentationEngine.Lib.Tests.vcxproj
```diff
@@ -107,6 +107,7 @@
     </ProjectReference>
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), EnlistmentRoot.marker))\build\Common.targets" />
+  <Import Project="$(EnlistmentRoot)\build\ClrInstrumentationEngine.cpp.props" />
```
delmyers, line 110: This is typically done at the top of the project, not after the targets.

build/yaml/jobs/codeanalysis/apiscan.yaml
delmyers: Use the baseline instead of the exclusions. That's why this isn't working. The way that you are expected to exclude false positives is to pass a file to the task that does the build check, not directly to APIScan.

---
In reply to: [522557817](https://github.com/microsoft/CLRInstrumentationEngine/pull/325#discussion_r522557817) [](ancestors = 522557817)

tests/InstrumentationEngine.Lib.Tests/InstrumentationEngine.Lib.Tests.vcxproj
clguiman: It's actually at the middle of the project due to some properties that have to be defined before. I replicated the order from InstrumentationEngine.vcxproj

build/yaml/steps/codeanalysis/postanalysis.yaml
```diff
@@ -29,4 +29,6 @@ steps:
     GdnBreakOutputBaseline: MyBL1
     GdnBreakOutputSuppressionFile: MySupp1
     GdnBreakOutputSuppressionSet: MySupp1
+    GdnBreakBaselineFiles: $(Build.SourcesDirectory)\build\yaml\steps\codeanalysis\baseline
+    GdnBreakBaselines: ApiScanBaseline
```
delmyers, line 33: OK for now, but we may need to parameterize this based on the tool that was run. This baseline assumes that all of the tools were run in the same job, which they are not.

delmyers (approved):
:shipit:

build/yaml/steps/codeanalysis/postanalysis.yaml
clguiman: I'll leave that for when we'll need to set a baseline for another tool

PR #326, opened 2020-11-12
Add instrumentation engine unit tests.

delmyers:
Adds the ability for unit tests to test the single return instrumentation
method. This is useful because it is a complex instrumentation method
that rearanges IL in multiple ways, and generates a complex output. It is
also public instrumentation funcitonality that we don't want to break.

1) Updates the Naggler profiler to recognize a <MakeSingleRet /> command.
2) Fixes issue in instrumentation engine in which we were mis-reporting the original IL during method dumps.
3) Adds tests for single returns, testing if, switch, and exceptions.

Also adds tests for special edge-case around inserting branch instructions. When using insert before, and retargetting branches, the branch target can sometimes cause an infinite loop. This tests to ensure that doesn't happen. Finally, some failures in instrumentation may cause the instrumented function to hang or loop forever, so I've added a timeout for the process.


tests/InstrEngineTests/TestScripts/MultiReturn_IfTest.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<InstrumentationTestScript>
+  <InstrumentMethod>
+    <ModuleName>MultiReturnTests_Release_x64.exe</ModuleName>
+    <MethodName>MultiIfTest</MethodName>
+    <!-- reset to single return -->
```
WilliamXieMSFT, line 6: > [](start = 4, length = 31)

nit remove unless this is important for the MultiIfTest?

delmyers: Yeah, I was originally going to add instrumentation, and then do the single return after that, but it ended up being more complicated than necessary. I'll remove this.

---
In reply to: [522466577](https://github.com/microsoft/CLRInstrumentationEngine/pull/326#discussion_r522466577) [](ancestors = 522466577)

delmyers (commented):
I'm approving these changes myself. They are additional tests.

WilliamXieMSFT (approved):

PR #327, opened 2020-11-16
Reduce overhead of retargeting branches.

delmyers:
Add back-links to branches from targets.

Makes branch targets track what branches target them.
This reduces the overhead of retargeting branches after
an insert because it means that the entire graph does
not need to be searched for branches. Instead the
branch target holds its own list of branches that target
it, and that list can be iterated over instead. This reduces
the cost of InsertBeforeAndRetargetOffsets from O(G) to O(B)
where B is the number of branches that target the Instruction
being inserted before. The amortized cost should therefore be
near O(1) because the average instruction is not a branch
target, and most branch targets will have only 1 branch tareting
them. I have not done the math to be certain of this, though.

The list of branches targeting a single instruction is stored
as a CDataItem so that we don't have to add additional pointers
to instructions where it isn't necessary. The list is a list
of raw pointers because every pointer in the list should be
kept alive by the graph, and using CComPtrs would increase overhead
while iterating through the list.

Also adds a couple of smaller performance improvements including:
1) Remove message logging from CInstruction. Some of the functions there are called in tight loops, and the additional overhead of message logging adds up.
2) Optimize GetInstructionSize(). Again, this is called in tight loops, and performing additional calculations for checking inputs and returning HResults adds up.

### DIFF
--- a/src/InstrumentationEngine/BranchTargetInfo.cpp
+++ b/src/InstrumentationEngine/BranchTargetInfo.cpp
@@ -1,0 +1,146 @@
+#include "stdafx.h"
+#include "refcount.h"
+#include "Instruction.h"
+#include "BranchTargetInfo.h"
+
+namespace MicrosoftInstrumentationEngine
+{
+    CBranchTargetInfo::CBranchTargetInfo(_In_ CInstruction* pInstruction) : m_pInstruction(pInstruction)
+    {
+
+    }
+
+    HRESULT CBranchTargetInfo::GetInstance(_In_ CInstruction* pInstruction, _Outptr_ CBranchTargetInfo** ppResult)
+    {
+        IfNullRet(pInstruction);
+        IfNullRet(ppResult);
+
+        *ppResult = nullptr;
+
+        const GUID* uuid = &__uuidof(CBranchTargetInfo);
+        CComPtr<IUnknown> pUnknown;
+        HRESULT hr;
+        // Don't assert, it is common for there to not be target info.
+        IfFailRetNoLog(pInstruction->GetDataItem(uuid, uuid, &pUnknown));
+        CComPtr<CBranchTargetInfo> pTargetInfo;
+        IfFailRet(pUnknown.QueryInterface(&pTargetInfo));
+
+        IfFalseRet(pTargetInfo != nullptr, E_UNEXPECTED);
+        *ppResult = pTargetInfo.Detach();
+
+        return S_OK;
+    }
+
+    HRESULT CBranchTargetInfo::GetOrCreateInstance(_In_ CInstruction* pInstruction, _Out_ CBranchTargetInfo** ppResult)
+    {
+        IfNullRet(pInstruction);
+        IfNullRet(ppResult);
+        if (!SUCCEEDED(GetInstance(pInstruction, ppResult)))
+        {
+            CComPtr<CBranchTargetInfo> pResult;
+            pResult.Attach(new CBranchTargetInfo(pInstruction));
+            const GUID* uuid = &__uuidof(CBranchTargetInfo);
+            pInstruction->SetDataItem(uuid, uuid, pResult);
+            *ppResult = pResult.Detach();
+        }
+
+        return S_OK;
+    }
+
+    HRESULT CBranchTargetInfo::SetBranchTarget(_In_ CInstruction* pBranch, _In_ CInstruction* pTarget, _In_opt_ CInstruction* pOldTarget)
+    {
+        IfNullRet(pBranch);
+        IfNullRet(pTarget);
+
+        HRESULT hr;
+
+        if (pOldTarget != nullptr)
+        {
+            CComPtr<CBranchTargetInfo> pOldInfo;
+
+            // There may not be an instance of CBranchTargetInfo if we are in the
+            // middle of retargetting a branch. That is OK, just ignore it.
+            if (SUCCEEDED(CBranchTargetInfo::GetInstance(pOldTarget, &pOldInfo)))
+            {
+                pOldInfo->Remove(pBranch);
+            }
+        }
+
+        CComPtr<CBranchTargetInfo> pInfo;
+        IfFailRet(CBranchTargetInfo::GetOrCreateInstance(pTarget, &pInfo));
+        pInfo.p->m_branches.emplace(pBranch);
+        return S_OK;
+    }
+
+    HRESULT CBranchTargetInfo::RetargetBranches(_In_ CInstruction* pOriginalInstruction, _In_ CInstruction* pNewInstruction)
+    {
+        IfNullRet(pOriginalInstruction);
+        IfNullRet(pNewInstruction);
+
+        CComPtr<CBranchTargetInfo> pTargetInfo;
+        if (SUCCEEDED(CBranchTargetInfo::GetInstance(pOriginalInstruction, &pTargetInfo)))
+        {
+            return pTargetInfo->Retarget(pNewInstruction);
+        }
+        return S_OK;
+    }
+
+    HRESULT CBranchTargetInfo::Retarget(_In_ CInstruction* pNewInstruction)
+    {
+        HRESULT hr = S_OK;
+
+        // keep this alive, since disconnecting branch targets will cause
+        // this info to be removed from the data container.
+        CComPtr<CBranchTargetInfo> _this = this;
+
+        // Remove this CBranchTargetInfo from the data container. If a new one
+        // is needed, it will be regenerated when branch targets are set.
+        Disconnect();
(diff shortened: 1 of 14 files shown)


delmyers:
closes #328

src/InstrumentationEngine/BranchTargetInfo.cpp
```diff
@@ -0,0 +1,146 @@
...
+        CComPtr<CBranchTargetInfo> _this = this;
+
+        // Remove this CBranchTargetInfo from the data container. If a new one
+        // is needed, it will be regenerated when branch targets are set.
+        Disconnect();
+        for (CInstruction* pInstr : m_branches)
+        {
+            if (pInstr->GetIsBranchInternal())
+            {
+                CBranchInstruction* pBranch = (CBranchInstruction*)pInstr;
```
wiktork, line 103: nit: Did you want to QI here as well like you do for the switch?

src/InstrumentationEngine/BranchTargetInfo.h
```diff
@@ -0,0 +1,45 @@
...
+
+        virtual ~CBranchTargetInfo() = default;
+    private:
+
+        // This is a raw pointer to avoid circular references.
+        CInstruction* m_pInstruction;
+
+        // Using raw pointers for speed. The graph should keep these pointers
+        // alive.
+        std::unordered_set<CInstruction*> m_branches;
```
wiktork, line 39: Have you tried some of the extremely large code bases to see what kind of memory impact this has?

src/InstrumentationEngine/Instruction.cpp
```diff
@@ -362,221 +356,144 @@ HRESULT MicrosoftInstrumentationEngine::CInstruction::SetIsRemoved()
...
-
-    CLogging::LogMessage(_T("End CInstruction::GetIsRemoved"));
-
     return S_OK;
 }
 
-HRESULT MicrosoftInstrumentationEngine::CInstruction::GetInstructionSize(_Out_ DWORD* pdwSize)
+DWORD MicrosoftInstrumentationEngine::CInstruction::GetInstructionSize()
 {
-    HRESULT hr = S_OK;
```
wiktork, line 527: Maybe add a comment to indicate that GetInstructionSize is no longer implemented in terms of GetOpcodeLength/GetOperandLength

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -1140,23 +1104,9 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::RemoveAll()
     while (pInstr != NULL)
     {
         pInstr->SetIsRemoved();
-
-        CInstruction* pPreviousInstruction = pInstr->PreviousInstructionInternal();
         CInstruction* pNextInstruction = pInstr->NextInstructionInternal();
-
-        if (pPreviousInstruction != NULL)
```
wiktork, line 1147: The original code does seem redundant, but I'm curious did this cause issues? I'm trying to figure out what we were trying to do here originally.

src/InstrumentationEngine/Instruction.cpp
```diff
@@ -1154,20 +995,17 @@ HRESULT MicrosoftInstrumentationEngine::CBranchInstruction::GetTargetOffset(_Out
...
     HRESULT hr = S_OK;
-
-    CLogging::LogMessage(_T("Starting CBranchInstruction::SetBranchTarget"));
     IfNullRetPointer(pInstruction);
 
+    CComPtr<CInstruction> pOldInstruction = m_pBranchTarget;
     m_pBranchTarget = (CInstruction*)pInstruction;
+    IfFailRet(CBranchTargetInfo::SetBranchTarget(this, (CInstruction*)pInstruction, pOldInstruction));
 
     if (m_pBranchTarget == nullptr)
```
wiktork, line 1010: We seem to be ok with a null pInstruction, but SetBranchTarget will return E_FAIL in that situation.

src/InstrumentationEngine/BranchTargetInfo.cpp
wiktork: Nvm I see this is the same as the original code.

src/InstrumentationEngine/BranchTargetInfo.cpp
```diff
@@ -0,0 +1,146 @@
...
+        return S_OK;
+    }
+
+    HRESULT CBranchTargetInfo::Retarget(_In_ CInstruction* pNewInstruction)
+    {
+        HRESULT hr = S_OK;
+
+        // keep this alive, since disconnecting branch targets will cause
+        // this info to be removed from the data container.
+        CComPtr<CBranchTargetInfo> _this = this;
```
wiktork, line 94: I don't see why this is needed. You are calling Retarget via a ComPtr<CBranchTargetInfo>, which should have added a reference to CBranchTargetInfo for the duration of this function.

wiktork (approved):

jakubch1 (approved):
I've checked on our side and now instrumentation is very fast. Overall statistics it's +3-4%.
The most time we spend on OnModuleLoad when we calculate all blocks - we will try to make it faster on our side. Thank you a lot for all help on making code coverage instrumentation linear!

src/InstrumentationEngine/BranchTargetInfo.h
delmyers: No, I have not checked the memory impact. It will be linear in the number of branches, though. We could do a scan of mscorlib to find all of the branch instructions, and find an average or something. I don't know the best way to disassemble, though.

src/InstrumentationEngine/InstructionGraph.cpp
delmyers: Yes, it did cause issues. The removal of the pointers on the previous instruction caused the last instruction to ref count down to zero before the last iteration, causing an AV. I don't know what the original purpose was of the additional removals.

src/InstrumentationEngine/BranchTargetInfo.h
delmyers: OK, so I ran an experiment with a program that has about 11000 if instructions, and it looks like the memory is about 10x(pointer_size) per branch instruction. We might be able to make that smaller by reducing the initial size of this set (I can give it a try). We could probably get away with an initial size of 1, since most instructions won't be targeted by many branches.

delmyers: All right. I've updated the constructor to this set, and the map used in CDataContainer to free unused memory correctly, and now the memory impact is zero compared to the baseline of what is in the develop branc.

src/InstrumentationEngine/BranchTargetInfo.cpp
delmyers: fixed.

src/InstrumentationEngine/Instruction.cpp
```diff
@@ -1297,41 +1126,31 @@ HRESULT MicrosoftInstrumentationEngine::CSwitchInstruction::SetBranchTarget(_In_
     HRESULT hr = S_OK;
     CLogging::LogMessage(_T("Starting CSwitchInstruction::SetBranchTarget"));
 
-    m_branchTargets[index] = pTarget;
-
-    CLogging::LogMessage(_T("End CSwitchInstruction::SetBranchTarget"));
+    CComPtr<CInstruction> oldTarget = m_branchTargets[index];
+    m_branchTargets[index] = (CInstruction*)(pTarget);
+    CBranchTargetInfo::SetBranchTarget(this, (CInstruction*)(pTarget), oldTarget);
```
wiktork, line 1131: Did you want to propagate the hresult here?

src/InstrumentationEngine/Instruction.h
```diff
@@ -159,25 +163,25 @@ namespace MicrosoftInstrumentationEngine
         bool GetIsBranchInternal() const { return IsFlagSet(s_ilOpcodeInfo[m_opcode].m_flags, ILOpcodeFlag_Branch); }
 
     protected:
-        template<typename... Args>
-        HRESULT QueryInterfaceInternal(_In_ REFIID riid, _Out_ void** ppvObject, Args... pThisArgs)
+
+        template<typename TBase>
+        HRESULT QueryBaseType(_In_ REFIID riid, _Out_ void** ppvObject)
```
wiktork, line 168: Would it be easier if we folded QueryBaseType and QueryInterfaceInternal into one function? Something like:
```c++
template<typename TBase, typename... Args>
QueryInterfaceInternal(...)
```
````

src/InstrumentationEngine/Instruction.cpp
delmyers: Yup, fixed.

PR #332, opened 2020-11-30
Remove RPH if Initialize returns error

WilliamXieMSFT:
If RPH returns an error hresult, rather than failing Initialize altogether for CLRIE, this change will allow CLRIE to just remove the RPH instead.

### DIFF
--- a/src/InstrumentationEngine/ProfilerManager.cpp
+++ b/src/InstrumentationEngine/ProfilerManager.cpp
@@ -974,11 +974,16 @@ HRESULT CProfilerManager::Initialize(
         {
             if (m_attachedClrVersion != ClrVersion_2)
             {
-                IfFailRet(pCallback->Initialize((IUnknown*)(m_pWrappedProfilerInfo.p)));
+                IfFailLog(pCallback->Initialize((IUnknown*)(m_pWrappedProfilerInfo.p)));
             }
             else
             {
-                IfFailRet(pCallback->Initialize((IUnknown*)(m_pRealProfilerInfo.p)));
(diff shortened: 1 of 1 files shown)


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -974,11 +974,16 @@ HRESULT CProfilerManager::Initialize(
...
-                IfFailRet(pCallback->Initialize((IUnknown*)(m_pWrappedProfilerInfo.p)));
+                IfFailLog(pCallback->Initialize((IUnknown*)(m_pWrappedProfilerInfo.p)));
             }
             else
             {
-                IfFailRet(pCallback->Initialize((IUnknown*)(m_pRealProfilerInfo.p)));
+                IfFailLog(pCallback->Initialize((IUnknown*)(m_pRealProfilerInfo.p)));
+            }
+
+            if (FAILED(hr))
```
wiktork, line 984: If RPH is turned on and fails, but no instrumentation methods are on, do we fail gracefully?

wiktork (approved):

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: I'll double check this, the desired behavior is that the process should have CLRIE running without any instrumentation methods.

WilliamXieMSFT: Confirmed that nothing happens in that case

PR #333, opened 2020-12-10
Securely configure package source files

WilliamXieMSFT:


NuGet.config
```diff
@@ -3,7 +3,11 @@
      Licensed under the MIT License. -->
 <configuration>
     <packageSources>
+        <!-- Use single feed to mitigate substitution attacks and allow easier supply chain scanning.
+             For more information: https://www.1eswiki.com/wiki/How_to_Securely_Configure_Package_Source_Files -->
+        <clear />
         <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+        <!-- Azure Artifacts does not support myget sources as upstreams. -->
         <add key="ApplicationInsights-MyGet" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
```
WilliamXieMSFT, line 11: We pull the packages for testing, I could consider rehosting them in an upstream feed at ClrInstrumentationEngine-Dependency which is what our Unix build is using.

wiktork (approved):

NuGet.config
```diff
@@ -3,6 +3,7 @@
      Licensed under the MIT License. -->
 <configuration>
     <packageSources>
+        <clear />
         <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
         <add key="ApplicationInsights-MyGet" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
```
WilliamXieMSFT, line 8: We pull the packages from this MyGet for testing, I could consider rehosting them in an upstream feed at ClrInstrumentationEngine-Dependency which is what our Unix build is using.

PR #337, opened 2020-12-15
Use dependency feed

WilliamXieMSFT:


NuGet.config
```diff
@@ -5,7 +5,8 @@
     <packageSources>
         <clear />
         <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-        <add key="ApplicationInsights-MyGet" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
+        <add key="ClrInstrumentationEngine-Dependency"
```
delmyers, line 8: Is this feed open to the public? Or will this prevent individuals from building locally? Also, is nuget.org upstreamable, so that we only have one feed?

delmyers (dismissed):

NuGet.config
WilliamXieMSFT: The CLRIE-Dependency feed is public (https://dev.azure.com/ms/CLRInstrumentationEngine/_packaging?_a=feed&feed=ClrInstrumentationEngine-Dependency), and public feeds currently don't support upstreaming.

delmyers: Thanks. So that means, there is no way to use a single feed in this repo?

WilliamXieMSFT: Correct, until we can upstream our public feed to nuget.org, we'll have to live with this setup.

jander-msft (changes requested):
nuget.org should be listed after the dependency feed.

jander-msft (approved):

PR #338, opened 2020-12-17
Improve RPH performance by removing critical section lock

WilliamXieMSFT:
Callbacks for raw profiler can be bottlenecked by the critical section lock around obtaining the raw profiler.

CLRIE doesn't remove the RawProfiler during runtime (except during Initialize callback) and this change makes the public RemoveRawProfiler API to no-op for extra insurance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Please format the changes as follows:
 + Updates:
 
 ## 1.0.36
++ Perf:
+  + Remove critical section around rawprofiler handle which can bottleneck objectalloc callbacks.
 
 ## 1.0.35
 + New:
@@ -29,7 +31,6 @@ Please format the changes as follows:
   + Update PR/CI stage dependencies
   + Use NuGet 5.8 when restoring
   + Use IfFailLog when rawprofiler fails to initialize.
-
 
 ## 1.0.34
 + BugFixes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Please format the changes as follows:
 
 ## 1.0.36
 + Perf:
-  + Remove critical section around rawprofiler handle which can bottleneck objectalloc callbacks.
+  + use atomic shared_ptr instead of critical section around rawprofiler handle which can bottleneck objectalloc callbacks.
 
 ## 1.0.35
 + New:
--- a/src/InstrumentationEngine/ProfilerManager.cpp
+++ b/src/InstrumentationEngine/ProfilerManager.cpp
@@ -186,6 +186,7 @@ HRESULT CProfilerManager::AddRawProfilerHook(
     HRESULT hr = S_OK;
     IfNullRetPointer(pUnkProfilerCallback);
 
+    std::unique_lock<std::shared_mutex> srwLock(m_sharedMutexRawProfiler);
     if (m_profilerCallbackHolder != nullptr)
     {
         CLogging::LogError(_T("CAppDomainInfo::AddRawProfilerHook - Raw profiler hook is already initialized"));
@@ -314,7 +315,10 @@ HRESULT CProfilerManager::AddRawProfilerHook(
 HRESULT CProfilerManager::RemoveRawProfilerHook(
     )
 {
-    // Prevent external clients from removing RawProfilerHook during lifetime of process.
+    std::unique_lock<std::shared_mutex> srwLock(m_sharedMutexRawProfiler);
+
+    m_profilerCallbackHolder = nullptr;
+
     return S_OK;
 }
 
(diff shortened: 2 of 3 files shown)


src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -368,15 +368,9 @@ namespace MicrosoftInstrumentationEngine
...
-            // events and calls will block. Copy the collection under the lock, then release it and finally call the callbacks
+            if (m_profilerCallbackHolder != nullptr)
             {
-                CCriticalSectionHolder lock(&m_cs);
-
-                if (m_profilerCallbackHolder != nullptr)
-                {
-                    pCallback = (TInterfaceType*)(m_profilerCallbackHolder->GetMemberForInterface(__uuidof(TInterfaceType)));
-                }
+                pCallback = (TInterfaceType*)(m_profilerCallbackHolder->GetMemberForInterface(__uuidof(TInterfaceType)));
```
wiktork, line 373: This is not the only place we get the callback under a lock. See COMClassicVTableCreated.

wiktork, line 373: I think this is safe since m_profilerCallbackHolder should just get initialized in the Initialize method and either become null or stay non-null for the rest of its lifetime. If that doesn't end up being the case we could also use Interlocked functions instead of critical section.

wiktork (approved):

src/InstrumentationEngine/ProfilerManager.h
WilliamXieMSFT: Thanks

wiktork: Per our discussion, if we want to maintain the RemoveRawProfilerHook, we do need to synchronize m_profilerCallbackHolder, just not with a critical section.

WilliamXieMSFT: We discussed several options offline:

atomic shared_ptr provides thread safety only when writing against different shared_ptr instances. Writing to the same shared_ptr instance must be exclusively locked:
```
p1 = p; p2 = p; // allowed, Reads of p
p3 = p1; p1 = nullptr; // not allowed, Read/Write of p1
p2 = p; p2 = nullptr; // not allowed, multiple Writes of p2
```

atomic bool guards are not sufficient as atomic bools do not guarantee the pointer safety:
```
// Thread 1
if (atomic_bool) { /* use ptr */ }
// Thread 2
atomic_bool = false
/* modify ptr */

// we may succeed the check in Thread 1 but before using ptr, Thread 2 might set the atomic bool and modify ptr
```

WilliamXieMSFT: I've replaced critical_sections with unique_lock (for writes) and shared_lock (for reads) on a shared_mutex.

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -171,6 +172,7 @@ namespace MicrosoftInstrumentationEngine
         // This is the client that asked to receive the raw cor profiler event
         // model rather than using the simplified instrumentation method model. .
         unique_ptr<CProfilerCallbackHolder> m_profilerCallbackHolder;
+        mutable std::shared_mutex m_sharedMutexRawProfiler;
```
WilliamXieMSFT, line 175: >m_sharedMutexRawProfiler [](start = 34, length = 24)

nit m_sharedRawProfilerMutex

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Please format the changes as follows:
 + Updates:
 
 ## 1.0.36
++ Perf:
+  + Remove critical section around rawprofiler handle which can bottleneck objectalloc callbacks.
```
WilliamXieMSFT, line 12: >Remove critical section [](start = 4, length = 23)

update once we settle on an implementation

src/InstrumentationEngine/ProfilerManager.h
jander-msft: Have you tried building these on Linux? I'm not certain the C++ library that our version of clang understands is new enough for these.

wiktork: I think this is a good solution, but we'll have to validate the performance.

WilliamXieMSFT: atomic shared_ptr functions compile so I'll go ahead with that implementation.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -306,19 +307,16 @@ HRESULT CProfilerManager::AddRawProfilerHook(
...
 HRESULT CProfilerManager::RemoveRawProfilerHook(
     )
 {
-    HRESULT hr = S_OK;
-
-    CCriticalSectionHolder lock(&m_cs);
-
-    m_profilerCallbackHolder = nullptr;
+    atomic_store(&m_profilerCallbackHolder, shared_ptr<CProfilerCallbackHolder>(nullptr));
+    //m_profilerCallbackHolder.reset();
```
wiktork, line 319: nit: remove commented line

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -963,28 +961,28 @@ HRESULT CProfilerManager::Initialize(
         return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
     }
 
-    // Take the lock that protects the raw profiler callback and the instrumentation methods. This keeps the collection from changing out from under the iterator
+    CComPtr<ICorProfilerCallback2> pCallback;
 
-    CCriticalSectionHolder lock(&m_cs);
+    shared_ptr<CProfilerCallbackHolder> pProfilerCallbackHolder = atomic_load(&m_profilerCallbackHolder);
```
wiktork, line 966: do we want to have explicit memory_order flags for these calls? I don't know what the defaults are.

WilliamXieMSFT: atomic_load is equivalent to atomic_load_explicit(p, std::memory_order_seq_cst)
atomic_store is equivalent to atomic_store_explicit(p, r, memory_order_seq_cst)

Reading seq_cst ordering (https://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering),
I am a bit concerned about this note:

`Total sequential ordering requires a full memory fence CPU instruction on all multi-core systems. This may become a performance bottleneck since it forces the affected memory accesses to propagate to every core.`

WilliamXieMSFT: I was able to get confirmation that the default flags meet performance requirements.

src/InstrumentationEngine/ProfilerManager.h
delmyers: FYI, these will be replaced with atomic<shared_ptr> in C++ 20, not that we will be switching to that any time soon.

---
In reply to: [546068525](https://github.com/microsoft/CLRInstrumentationEngine/pull/338#discussion_r546068525) [](ancestors = 546068525)

PR #340, opened 2021-01-11
Use interlockedexchange instead of atomic shared_ptr

WilliamXieMSFT:
It appears atomic shared_ptr functions can still cause a context switch due to the underlying spin-lock implementation on Windows. This change removes locks entirely in favor of interlockedexchange functions. However, this also means there's no lock that can guarantee proper deletion of the object which means we are left with a memory leak.

### DIFF
--- a/src/InstrumentationEngine/ProfilerManager.cpp
+++ b/src/InstrumentationEngine/ProfilerManager.cpp
@@ -37,7 +37,8 @@ CProfilerManager::CProfilerManager() :
     m_bIsInitializingInstrumentationMethod(false),
     m_dwInstrumentationMethodFlags(0),
     m_bValidateCodeSignature(true),
-    m_bAttach(false)
+    m_bAttach(false),
+    m_profilerCallbackHolder(nullptr)
 {
 #ifdef PLATFORM_UNIX
     PAL_Initialize(0, NULL);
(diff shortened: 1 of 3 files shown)


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2799,7 +2810,11 @@ HRESULT CProfilerManager::COMClassicVTableCreated(
...
 
-    shared_ptr<CProfilerCallbackHolder> pProfilerCallbackHolder = atomic_load(&m_profilerCallbackHolder);
+    CProfilerCallbackHolder* pProfilerCallbackHolder = static_cast<CProfilerCallbackHolder*>(InterlockedCompareExchangePointer(
+        (volatile PVOID*)&m_profilerCallbackHolder,
+        nullptr,
+        nullptr));
+
     if (pProfilerCallbackHolder != nullptr)
     {
         pCallback = (ICorProfilerCallback*)(m_profilerCallbackHolder->GetMemberForInterface(__uuidof(ICorProfilerCallback)));
```
wiktork, line 2820: We should be using pProfilerCallbackHolder here, instead of the field directly

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -2828,7 +2843,11 @@ HRESULT CProfilerManager::COMClassicVTableDestroyed(
...
```
wiktork, line 2853: Same issue here: we should not be using the field directly

src/InstrumentationEngine/ProfilerManager.h
```diff
@@ -368,7 +373,7 @@ namespace MicrosoftInstrumentationEngine
 
             CComPtr<TInterfaceType> pCallback;
 
-            shared_ptr<CProfilerCallbackHolder> pProfilerCallbackHolder = atomic_load(&m_profilerCallbackHolder);
+            CProfilerCallbackHolder* pProfilerCallbackHolder = InterlockedCompareExchangePointer(&m_profilerCallbackHolder, nullptr, nullptr);
```
wiktork, line 376: The function here does not seem to require the cast on the parameter or the return value while other invocations do.

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Thanks for the catch

src/InstrumentationEngine/ProfilerManager.h
WilliamXieMSFT: Right, I wasn't sure why the compiler was ok with this. I'll add the explicit static_cast.

wiktork (approved):

PR #341, opened 2021-01-11
Fix CComBSTR not appending the null-terminator

WilliamXieMSFT:
`CComBSTR(_In_opt_z_ LPCOLESTR pSrc)` calls `SysAllocString(pSrc)` which doesn't append the null-terminator
`CComBSTR(_In_ int nSize, _In_reads_opt_(nSize) LPCOLESTR sz)` calls `SysAllocStringLen(sz, nSize)` which does append the null-terminator.


WilliamXieMSFT:
Abandoning this PR as the current behavior in CLRIE is correct. 

The code that's failing is:
```
static WCHAR PInvokeStubTypeName[] = _T("1234567890123456789012345678901234567890123456789"); // 49 characters + null terminator

hr = pDeclaringType->GetName(&pName);
CompareStringOrdinal(pName.m_str, pName.Length(), PInvokeStubTypeName, _countof(PInvokeStubTypeName), FALSE) == CSTR_EQUAL;
```

_countof will return the value of 50 (49 characters + null terminator) as it treats the string as a data array whereas pName.Length will return 49 (the length of the BSTR). This fix is unnecessary.

Issue #348, opened 2021-01-26
Configuration Variables are Case Sensitive


plnelson:
CLRInstrumentationEngine Doesn't recognize configuration unless casing is exact.  For example `MicrosoftInstrumentationEngine_ConfigPath64_*` is recognized.  In contrast `microsoftinstrumentationengine_configpath64_*`.  is _not_ recognized.  The CLR is case insensitive for variables such as `COR_ENABLE_PROFILING` so CLRInstrumentationEngine should be as well.

This is causing an issue for me because the IISExpress launcher is changing environment variables to lower case blocking dynamic instrumentation profiling of ASP.NET applications.  Rather than fix that code, it seems more appropriate to make the CLRInstrumentationEngine config case insensitive.

wiktork:
It looks like we are inconsistent with how we process environment variable names:
- Most go through `GetEnvironmentVariable`, and are case insensitive
- Some have explicit checks, like `MicrosoftInstrumentationEngine_ConfigPath64_` mentioned above so they end up case sensitive.
- We should also consider whether we care about the case sensitivity for values. For example, `MicrosoftInstrumentationEngine_LogLevel` and `MicrosoftInstrumentationEngine_FileLog` have case sensitive values.

plnelson:
I was planning to leave the values case sensitive - as well as leave the names case sensitive on Unix.  IMO, the log path should be case sensitive.  If we want to be case insensitive about the meaning of log level, that is a different issue to consider.

wiktork:
> I was planning to leave the values case sensitive - as well as leave the names case sensitive on Unix. IMO, the log path should be case sensitive. If we want to be case insensitive about the meaning of log level, that is a different issue to consider.

Sounds good. 

> Most go through GetEnvironmentVariable, and are case insensitive

This is for Windows only, GetEnvironmentVariable is case sensitive for Unix.

PR #349, opened 2021-01-26
Make environment variable names case insensitive on Windows

plnelson:
Make environment variable names case insensitive on Windows.
This fixes [348](https://github.com/microsoft/CLRInstrumentationEngine/issues/348).  Fix is needed for internal VS issue 1264553.

The IISExpress launcher is making environment variables lower case including "cor_enable_profiling".  After discussing with @wiktork, we decided the better solution is to fix CLRInstrumentationEngine rather than the IISExpress launcher.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Please format the changes as follows:
 + Updates:
 
 ## 1.0.36
++ BugFixes:
+  + Make configuration environment variables case-insensitive on Windows.
 + Perf:
   + use atomic shared_ptr instead of critical section around rawprofiler handle which can bottleneck objectalloc callbacks.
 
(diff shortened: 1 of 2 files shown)


wiktork (approved):
Can you also update https://github.com/microsoft/CLRInstrumentationEngine/blob/develop/CHANGELOG.md to reflect this change?

WilliamXieMSFT:
+1

---
In reply to: [576705848](https://github.com/microsoft/CLRInstrumentationEngine/pull/349#issuecomment-576705848) [](ancestors = 576705848)

WilliamXieMSFT (approved):

plnelson:
Done

---
In reply to: [576705848](https://github.com/microsoft/CLRInstrumentationEngine/pull/349#issuecomment-576705848) [](ancestors = 576705848)

PR #352, opened 2021-01-27
Adjust feeds

WilliamXieMSFT:


NuGet.config
```diff
@@ -6,7 +6,7 @@
         <clear />
         <add key="ClrInstrumentationEngine-Dependency"
              value="https://pkgs.dev.azure.com/ms/CLRInstrumentationEngine/_packaging/ClrInstrumentationEngine-Dependency/nuget/v3/index.json" />
-        <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+        <add key="Dotnet-Public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
```
delmyers, line 9: Is this what Concord does? We are trying to get rid of multiple feeds.

delmyers (approved):

PR #353, opened 2021-01-28
Test that demonstrate adding an exception handler

robertpi:
Set of unit test that show how an instrumentation method can be used to
wrap a method being jitted in an exception handler.

I got errors when for some of the tests, as CInstructionGraph::Replace
doesn't call CBranchTargetInfo::RetargetBranches. This means if you
replace an instruction that is a branch target, the branches will point
to the beginning of the method. This seems like a bug so I added a call
to CBranchTargetInfo::RetargetBranches in CInstructionGraph::Replace.


tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
```diff
@@ -338,7 +339,12 @@ private static void DiffResultToBaselineNode(XmlNode baselineNode, XmlNode outpu
             {
                 if (CompareOrdinalNormalizeLineEndings(baselineNode.Value, outputNode.Value) != 0)
                 {
-                    Assert.Fail("Baseline value does not equal output value\n" + baselineNode.Value + "\n" + outputNode.Value);
+                    int index = baselineNode.Value.Zip(outputNode.Value, (c1, c2) => c1 == c2).TakeWhile(b => b).Count() + 1;
```
delmyers, line 342: >(c1, c2) => c1 == c2 [](start = 73, length = 20)

In order to match the check above, should we normalize the endings first?

robertpi: I spent quite a lot of time trying to work out why two nearly identical strings didn't match. Turns out instructions node needs to end with exactly 4 spaces.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1187,6 +1202,133 @@ HRESULT CInstrumentationMethod::PerformSingleReturnInstrumentation(IMethodInfo*
...
+    CComPtr<IMetaDataImport> spIMetaDataImport;
+    IfFailRet(pModuleInfo->GetMetaDataImport(reinterpret_cast<IUnknown**>(&spIMetaDataImport)));
+
+    CComPtr<IMetaDataAssemblyImport> spIMetaDataAssemblyImport;
+    IfFailRet(pModuleInfo->GetMetaDataImport(reinterpret_cast<IUnknown**>(&spIMetaDataAssemblyImport)));
+
+    mdAssembly resolutionScope;
+    spIMetaDataAssemblyImport->GetAssemblyFromScope(&resolutionScope);
+
+    std::wstring strExceptionTypeName(_T("System.Exception"));
```
delmyers, line 1236: >std::wstring strExceptionTypeName(_T("System.Exception")); [](start = 4, length = 58)

We don't currently execute these tests on Linux, but I believe that std::wstring will give the wrong character width for calling functions that take a WCHAR in the ICorProfiler interfaces in Linux and Mac. We have another string type: tstring to handle this.

delmyers: Actually, never mind that comment. It seems that we don't handle it in general in this file. We should probably test and do a general cleanup.

---
In reply to: [566286936](https://github.com/microsoft/CLRInstrumentationEngine/pull/353#discussion_r566286936) [](ancestors = 566286936)

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
delmyers: Yes, we could do a better job with this, for sure.

---
In reply to: [566280346](https://github.com/microsoft/CLRInstrumentationEngine/pull/353#discussion_r566280346) [](ancestors = 566280346)

tests/InstrEngineTests/Baselines/AddExceptionHandler_ReturnMultiVoidMethodTest.xml
```diff
@@ -0,0 +1,81 @@
...
+    <CodeRVA>0x0000223e</CodeRVA>
+    <MethodImpl>0x00000000</MethodImpl>
+    <LocalSignature>0x07 0x00 </LocalSignature>
+    <MaxStack>0x00000002</MaxStack>
+    <Instructions>
+        <![CDATA[
+IL_0000 ldsfld 0x4000001
+IL_0005 ldc.i4.m1
+IL_0006 bne.un IL_0010
+IL_000b br IL_001b
```
delmyers, line 31: Something is confusing me about this test. When I compile this on my box, the IL I get is the following:

.method private hidebysig static void  ReturnMultiVoidMethodTest() cil managed
{
  // Code size       16 (0x10)
  .maxstack  8
  IL_0000:  ldsfld     int32 AddExceptionHandlerTests.Program::'error'
  IL_0005:  ldc.i4.m1
  IL_0006:  bne.un.s   IL_0009
  IL_0008:  ret
  IL_0009:  ldc.i4.0
  IL_000a:  stsfld     int32 AddExceptionHandlerTests.Program::'error'
  IL_000f:  ret
} // end of method Program::ReturnMultiVoidMethodTest

You can see that in the original set of instructions, this branch instruction is a return. I can't find anywhere in your instrumentation where you replace the return instruction. It is unclear to me how you are correctly handling multiple returns. Am I missing something?

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -899,6 +909,11 @@ HRESULT CInstrumentationMethod::InstrumentMethod(_In_ IMethodInfo* pMethodInfo,
         PerformSingleReturnInstrumentation(pMethodInfo, pInstructionGraph);
     }
 
+    if (pMethodEntry->IsAddExceptionHandler())
+    {
+        AddExceptionHandler(pMethodInfo, pInstructionGraph);
+    }
```
delmyers, line 915: Should this be applied after the other instrumentation? Otherwise, people who are writing test scripts will have to infer what transformations are going to be performed by the exception handler rather than just looking at the original IL.

tests/InstrEngineTests/Baselines/AddExceptionHandler_ReturnMultiVoidMethodTest.xml
robertpi: I don't handle multi return, I add the <MakeSingleRet /> node to the test script. I do this because I plan to convert any methods I instrument to single return, but I think it's interesting to test the results of an conversion.

See the test script here: https://github.com/microsoft/CLRInstrumentationEngine/pull/353/files#diff-11750276abb662ba0b270967e0279ef2b4056ef6baec3ddfeff8a6fd13fd6382R7

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Yes, you're right it should probably be last. I've just been testing it on it's own for now, so it doesn't really matter.

tests/InstrEngineTests/Baselines/AddExceptionHandler_ReturnMultiVoidMethodTest.xml
delmyers: OK, that makes sense. Can you swap the ordering in the XML of the MakeSingleRet, and add a comment to make it explicit? I'll add comments to the files that I'm thinking about as well.

---
In reply to: [566315167](https://github.com/microsoft/CLRInstrumentationEngine/pull/353#discussion_r566315167) [](ancestors = 566315167)

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1187,6 +1202,133 @@ HRESULT CInstrumentationMethod::PerformSingleReturnInstrumentation(IMethodInfo*
...
+    if (targetTypeRef == mdTypeRefNil)
+    {
+        return E_FAIL;
+    }
+
+
+    CComPtr<IInstructionFactory> pInstructionFactory;
+    IfFailRet(pMethodInfo->GetInstructionFactory(&pInstructionFactory));
+
+    CComPtr<IInstruction> sptrExitLabel;
```
delmyers, line 1263: >sptr [](start = 26, length = 4)

nit: our coding style doesn't typically mark pointers as "smart" we just use pVariableName.

tests/InstrEngineTests/TestScripts/AddExceptionHandler_ReturnMultiMethodTest.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<InstrumentationTestScript>
+  <InstrumentMethod>
+    <ModuleName>AddExceptionHandlerTests_Release_x64.exe</ModuleName>
+    <MethodName>ReturnMultiMethodTest</MethodName>
+    <AddExceptionHandler>True</AddExceptionHandler>
+    <MakeSingleRet />
```
delmyers, line 7: ><MakeSingleRet /> [](start = 4, length = 17)

Just for clarity, can you put this tag above the exception handler, and add a comment saying that exception instrumentation does not handle multi returns?

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1187,6 +1202,133 @@ HRESULT CInstrumentationMethod::PerformSingleReturnInstrumentation(IMethodInfo*
     return S_OK;
 }
 
+HRESULT CInstrumentationMethod::AddExceptionHandler(IMethodInfo* pMethodInfo, IInstructionGraph* pInstructionGraph)
```
delmyers, line 1205: >AddExceptionHandler [](start = 32, length = 19)

Add a comment somewhere in here that this function doesn't handle multiple returns.

delmyers: Also, consider adding some checks to ensure that there isn't more than one return instruction, and failing if there is. Otherwise, the execution of the test may end up throwing an exception during the jit of the method, and it would be difficult to track down why.

---
In reply to: [566343023](https://github.com/microsoft/CLRInstrumentationEngine/pull/353#discussion_r566343023) [](ancestors = 566343023)

delmyers: Your tests don't cover it, but tail calls can be a problem as well. You cannot have a tail call instruction inside try/catch/finally blocks.

---
In reply to: [566348196](https://github.com/microsoft/CLRInstrumentationEngine/pull/353#discussion_r566348196) [](ancestors = 566348196,566343023)

delmyers (changes requested):
Looks mostly very good. Just a couple of minor issues. Thanks for this change!

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: I think I copy pasted this prefix from here: https://github.com/microsoft/CLRInstrumentationEngine/blob/develop/tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp#L1009

But I mix several prefix styles in the same method, so I've tried to standardize on a single "p" in this commit: https://github.com/robertpi/CLRInstrumentationEngine/commit/a22fc6c204007911e9e2ff5fcfac6261382369df

tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs
robertpi: Fixed with: https://github.com/robertpi/CLRInstrumentationEngine/commit/473acd0848dfffc4e41f36ad561a365ea2698a7e

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Yes, it's technique I found used elsewhere.

tests/InstrEngineTests/Baselines/AddExceptionHandler_ReturnMultiVoidMethodTest.xml
robertpi: Okay, fixed we this commit: https://github.com/robertpi/CLRInstrumentationEngine/commit/d01b06ef411a69720c80d5d59f4f997120d2c092

I didn't add a comment as I changed all the test and file names, so I think that makes it clear.

tests/InstrEngineTests/TestScripts/AddExceptionHandler_ReturnMultiMethodTest.xml
robertpi: Done in this commit: https://github.com/robertpi/CLRInstrumentationEngine/commit/d01b06ef411a69720c80d5d59f4f997120d2c092

I renamed the files, so don't think a comment is necessary.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Added checks for this here: https://github.com/robertpi/CLRInstrumentationEngine/commit/3b9aedb3812b7f8fc56fc937dd0a9b02ecab0105

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Fixed here: https://github.com/robertpi/CLRInstrumentationEngine/commit/b1e30e9e351fdd8bb1a9fa7d61d8b301a1bec158

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1202,10 +1202,51 @@ HRESULT CInstrumentationMethod::PerformSingleReturnInstrumentation(IMethodInfo*
...
+
+    if (tailcallCount > 0)
+    {
+        ATLASSERT(L"Trying to add exception handle to method with tailcall");
+        return E_FAIL;
+    }
+
+    if (returnCount > 1) // note it would be valid to have method with zero returns, as we could end in a throw
+    {
+        ATLASSERT(L"Trying to add exception handle to method with more than one return");
```
delmyers, line 1246: typo: handler

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1202,10 +1202,51 @@ HRESULT CInstrumentationMethod::PerformSingleReturnInstrumentation(IMethodInfo*
...
+        }
+
+        CComPtr<IInstruction> pCurr = pInstruction;
+        pInstruction.Release();
+        pCurr->GetNextInstruction(&pInstruction);
+    }
+
+    if (tailcallCount > 0)
+    {
+        ATLASSERT(L"Trying to add exception handle to method with tailcall");
```
delmyers, line 1240: typo: handler

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1202,10 +1202,51 @@ HRESULT CInstrumentationMethod::PerformSingleReturnInstrumentation(IMethodInfo*
     return S_OK;
 }
 
+// this function does handle multiple returns or tailcalls
```
delmyers, line 1205: Do you mean "this function *does not handle*..."?

delmyers (changes requested):
Just a couple of typos. Sorry to nit-pick.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Fixed in: https://github.com/robertpi/CLRInstrumentationEngine/commit/83506ba636b3aa43e6860b9755dca696667ddfc0

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Fixed in: https://github.com/robertpi/CLRInstrumentationEngine/commit/83506ba636b3aa43e6860b9755dca696667ddfc0

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
robertpi: Fixed in: https://github.com/robertpi/CLRInstrumentationEngine/commit/83506ba636b3aa43e6860b9755dca696667ddfc0

delmyers (approved):
Looks good! Thanks.

PR #355, opened 2021-02-03
Multi-target the InstrEngineTests project.

jander-msft:
The following changes multi-target the InstrEntryTests project to bring it a step closer to running the tests for .NET Core on Windows and finally run them on Linux once the NaglerInstrumentationMethod is compiled for Linux.

Overview of changes:
- Multi-target the InstrEngineTests project.
- Use Microsoft.CodeAnalysis (Roslyn) instead of CodeDom.
- Set new baselines due to compiler changes.
- Remove unnecessary newlines from LogDumpMessage calls.
- Write method instructions and signatures on single line.
- Fix exception type token initialization.
- Temporarily do not run the new netcoreapp3.1 and net50 tests as they are not complete.


tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/DynamicCodeTests.cs
```diff
@@ -45,22 +49,57 @@ static void Main(string[] args)
...
         {
-            string[] assemblies = { "System.dll" };
-            CompilerParameters cp = new CompilerParameters(assemblies);
-            cp.OutputAssembly = "DynamicCodeAssembly.dll";
-            cp.GenerateInMemory = true;
-            CodeDomProvider provider = CodeDomProvider.CreateProvider("CSharp");
-            CompilerResults results = provider.CompileAssemblyFromSource(cp, CallForwarderSource);
-
-            Type t = results.CompiledAssembly.GetType("CompiledCode.CallForwarder");
+            SourceText sourceText = SourceText.From(CallForwarderSource);
```
jander-msft, line 52: Rather than having this debuggee invoke Roslyn to compile the assembly dynamically, I think I'll have it precompiled by the test. My understanding is that the CallCompiledAssembly test is focused on making such that assemblies that are loaded after the initial managed load can be instrumented. This type of change will simplify this test such that (1) Roslyn is not needed by the test anymore, (2) additional compiled-time references do not need to be calculated, and (3) the static config can be removed.

tests/InstrEngineTests/InstrEngineTests/TargetAppCompiler.cs
```diff
@@ -53,18 +55,32 @@ internal class TargetAppCompiler
...
-                string sourceCode = GetEmbeddedFile("InstrEngineTests.EmbeddedResources", prefix + ".cs");
+                string sourceCode = GetEmbeddedSourceFile(prefix);
                 SourceText sourceText = SourceText.From(sourceCode);
 
-                GenerateExecutable(sourceText, path, prefix, isDebug: true, is64bit: false);
-                GenerateExecutable(sourceText, path, prefix, isDebug: true, is64bit: true);
-                GenerateExecutable(sourceText, path, prefix, isDebug: false, is64bit: false);
-                GenerateExecutable(sourceText, path, prefix, isDebug: false, is64bit: true);
+                result = CompileTestAppPrefix(sourceText, directoryPath, prefix, isDebug: true, is64bit: false);
+                Assert.IsTrue(result.Success);
```
wiktork, line 69: Consider logging the errors from the compilation into the test result

tests/InstrEngineTests/InstrEngineTests/TestMethodReplacement.cs
```diff
@@ -28,44 +28,32 @@ public static void Initialize(TestContext context)
         [Timeout(TestConstants.TestTimeout)]
         public void ReplaceMethodTest()
         {
-            //0,0,3,3,6,6,a,a,e,e
+            // Note: despite the code having two locals, Roslyn will
+            // optimize one of the locals out during compilation.
```
wiktork, line 32: I believe you can turn this off, even in retail.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -1663,9 +1683,12 @@ HRESULT CInstrumentationMethod::ExceptionUnwindFunctionEnter(
 
     CComPtr<IProfilerManagerLogging> spLogger;
     m_pProfilerManager->GetLoggingInstance(&spLogger);
-    spLogger->LogDumpMessage(_T("<ExceptionUnwindFunctionEnter>"));
-    spLogger->LogDumpMessage(bstrFullName);
-    spLogger->LogDumpMessage(_T("</ExceptionUnwindFunctionEnter>\r\n"));
+
+    tstring strMessage(_T("    <ExceptionUnwindFunctionEnter>"));
```
wiktork, line 1687: Maybe a helper, like WriteTag(string buffer, tagName, indentLevel, optionalValue)?

wiktork (approved):
LGTM, although I didn't go through the baseline updates.

tests/InstrEngineTests/InstrEngineTests/TargetAppCompiler.cs
jander-msft: Done

tests/InstrEngineTests/InstrEngineTests/TestMethodReplacement.cs
jander-msft: I can't see how to do that. The only optimization settings I see are setting it to Debug or Release, which corresponds to /optimize- and /optimize+.

wiktork: I think you want /optimize+ and /debug+

jander-msft: Fixed, the option is hidden on the CSharpCompilationOptions class as internal methods and properties :-/

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
jander-msft: I think I'll leave it as-is for now and follow up with another change. I think there's more opportunity for XML logging unification using the CLogging::XmlDumpHelper class.

PR #356, opened 2021-02-05
Enable Nagler tests on .NET Core.

jander-msft:
These changes enable running the Nagler tests on .NET Core 3.1 and .NET 5 for Windows.

### DIFF
--- a/tests/InstrEngineTests/InstrEngineTests/TestAddInstruction.cs
+++ b/tests/InstrEngineTests/InstrEngineTests/TestAddInstruction.cs
@@ -6,12 +6,19 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 namespace InstrEngineTests
 {
     [TestClass]
+    // "dotnet test" runs tests in-place (instead of copying them to a deployment
+    // directory). DeploymentItems that copy from a source path to a destination
+    // that is the same location are not necessary and cause file locking issues
+    // when initializing the tests. Only use DeploymentItem for paths with different
+    // destinations for .NET Core test assemblies.
+#if !NETCOREAPP
     [DeploymentItem(PathUtils.BaselinesBinPath, PathUtils.BaselinesBinPath)]
     [DeploymentItem(PathUtils.TestScriptsBinPath, PathUtils.TestScriptsBinPath)]
+#endif
     [DeploymentItem(PathUtils.InstrumentationEngineX64BinPath)]
     [DeploymentItem(PathUtils.InstrumentationEngineX86BinPath)]
-    [DeploymentItem(PathUtils.InstrumentationConfigX64BinPath)]
-    [DeploymentItem(PathUtils.InstrumentationConfigX86BinPath)]
+    [DeploymentItem(PathUtils.NaglerInstrumentationConfigX64BinPath)]
+    [DeploymentItem(PathUtils.NaglerInstrumentationConfigX86BinPath)]
     [DeploymentItem(PathUtils.NaglerInstrumentationMethodX64BinPath)]
     [DeploymentItem(PathUtils.NaglerInstrumentationMethodX86BinPath)]
     public class TestAddInstruction
--- a/tests/InstrEngineTests/InstrEngineTests/TestDynamicCode.cs
+++ b/tests/InstrEngineTests/InstrEngineTests/TestDynamicCode.cs
@@ -6,12 +6,19 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 namespace InstrEngineTests
 {
     [TestClass]
+    // "dotnet test" runs tests in-place (instead of copying them to a deployment
+    // directory). DeploymentItems that copy from a source path to a destination
+    // that is the same location are not necessary and cause file locking issues
+    // when initializing the tests. Only use DeploymentItem for paths with different
+    // destinations for .NET Core test assemblies.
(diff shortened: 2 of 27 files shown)


tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -747,21 +747,26 @@ HRESULT CInstrumentationMethod::OnModuleLoaded(_In_ IModuleInfo* pModuleInfo)
...
             IfFailRet(pModuleInfo->GetMetaDataImport((IUnknown**)&pMetadataImport));
 
+            BOOL bIsRejit = FALSE;
+            IfFailRet(pInstrumentMethodEntry->GetIsRejit(&bIsRejit));
 
-            MicrosoftInstrumentationEngine::CMetadataEnumCloser<IMetaDataImport2> spTypeDefEnum(pMetadataImport, nullptr);
-            mdTypeDef typeDef = mdTypeDefNil;
-            ULONG cTokens = 0;
-            while (S_OK == (hr = pMetadataImport->EnumTypeDefs(spTypeDefEnum.Get(), &typeDef, 1, &cTokens)))
+            if (bIsRejit)
```
jander-msft, line 753: I'll add comments for this change. Our working theory is that .NET Core has an optimization where if ReJIT is called for a method before the JIT process has started, it will only callback to the profiler for ReJIT and not for the initial JIT (because the runtime already knows that initial JIT is not sufficient).

The test assembly used to blindly call ReJIT for each method the test cared about and would cause both the JIT and ReJIT callbacks to occur, however on .NET Core we only get the ReJIT callback. Thus, on .NET Core, if a test is expecting just the JIT, it will not get the expected callback and the test will fail.

The fix here is to only ask the method to be ReJIT'd if the test expects the callback for ReJIT.

WilliamXieMSFT:
> using System;

nit, add copyright header

---
Refers to: tests/InstrEngineTests/TestAppRunner/Program.cs:1 in 5f0d285. [](commit_id = 5f0d28559522ee4133a1a3b880eaddbb143e392b, deletion_comment = False)

tests/InstrEngineTests/InstrEngineTests/TestDynamicCode.cs
```diff
@@ -6,12 +6,19 @@
 namespace InstrEngineTests
 {
     [TestClass]
+    // "dotnet test" runs tests in-place (instead of copying them to a deployment
+    // directory). DeploymentItems that copy from a source path to a destination
+    // that is the same location are not necessary and cause file locking issues
+    // when initializing the tests. Only use DeploymentItem for paths with different
+    // destinations for .NET Core test assemblies.
+#if !NETCOREAPP
     [DeploymentItem(PathUtils.BaselinesBinPath, PathUtils.BaselinesBinPath)]
```
delmyers, line 15: >DeploymentItem [](start = 5, length = 14)

Is DeploymentItem an inheritable attribute? If yes, you could potentially just make a baseclass with all of these attributes attached instead of needing to repeat them.

tests/InstrEngineTests/TestAppRunner/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+                Action<string[]> entryPointDelegate = (Action<string[]>)assembly.EntryPoint
+                    .CreateDelegate(typeof(Action<string[]>));
+
+                entryPointDelegate(entryPointArgs);
+
+                return 0;
+            }
+            else
+            {
+                Func<string[], int> entryPointDelegate = (Func<string[], int>)assembly.EntryPoint
```
WilliamXieMSFT, line 52: >int [](start = 31, length = 3)

Would we ever have a Task or Task<int> main function? Maybe add a comment about this assumption?

tests/InstrEngineTests/InstrEngineTests/TestDynamicCode.cs
jander-msft: Unfortunately, it does not appear to be inheritable.

tests/InstrEngineTests/InstrEngineTests/TestAddInstruction.cs
```diff
@@ -6,12 +6,19 @@
 namespace InstrEngineTests
 {
     [TestClass]
+    // "dotnet test" runs tests in-place (instead of copying them to a deployment
+    // directory). DeploymentItems that copy from a source path to a destination
+    // that is the same location are not necessary and cause file locking issues
+    // when initializing the tests. Only use DeploymentItem for paths with different
+    // destinations for .NET Core test assemblies.
+#if !NETCOREAPP
```
wiktork, line 14: Not necessary but it might be cleaner if you create something like

[Conditional("...")]
ConditionalDeploymentItemAttribute : DeploymentItem

wiktork (approved):

WilliamXieMSFT (approved):
:shipit:

tests/InstrEngineTests/InstrEngineTests/TestAddInstruction.cs
jander-msft: DeploymentItemAttribute is sealed. It also doesn't implement any type of interface or base class, so I don't think I could create another attribute to mimic the behavior and defer to pre-existing behavior.

PR #362, opened 2021-02-09
Make CModuleRefCount available for Nagler

delmyers:
This is the first in a number of changes required to enable our native
tests to run on Linux. In order to do that, we will need to decouple
the Nagler instrumentation method from the ATL libraries, which
are only available on Windows. This change starts that work.

1) Move refcount.h/refcount.cpp into Common.Lib
2) Decouple Common.Lib from Windows/PAL (where necessary)
    a) Update refcount.cpp to use standard libraries for threading
        instead of ATL
    b) Create a header-only "mincom" library that defines all of the
        types necessary to create types based on IUnknown and
        do reference counting using CComPtr.
    c) Disable reference count logging in Linux. This is done
        because there is no implementation of debug print on
        Linux, and enabling one without the PAL would require
        additional tests.
3) Move header-only includes out of Common.Lib. This is
    essentially to make them separated from the files that are
    compiled into the Common.Lib binary.
4) Decouple Macros.h from CLogging. The library that uses
    Macros.h is now required to supply a function called
    AssertLogFailure. In InstrumentationEngine, this forwards
    to CLogging.
5) Reordered some header files.
6) Build and Link Common.Lib for Linux
7) Added -Interactive switch to DockerLocalBuild.ps1 so that
    we can have an interactive bash shell to run local builds in.


delmyers:
azp /run

src/Common.Lib/refcount.cpp
```diff
@@ -70,27 +72,26 @@ namespace MicrosoftInstrumentationEngine
...
+            OutputDebugStringW(szTrace.data());
         }
     }
-#endif
 
     // ----------------------------------------------------------------------------
     // CRefRBMap
-#ifdef DEBUG
-    class CRefRBMap : public ATL::CRBMap<CRefCount*, ULONG> {};
+    class CRefRBMap : public std::map<CRefCount*, ULONG> {};
```
wiktork, line 83: I'm assuming our implementation here diverges from the original because we don't want to bring in atl/pal into the common library?

src/Common.Lib/refcount.cpp
```diff
@@ -109,124 +109,108 @@ namespace MicrosoftInstrumentationEngine
...
 
                 TRACE(_T(">> End ref count dump <<"));
             }
-
-            LeaveCriticalSection(&g_crst);
         }
     }
 
-#endif // DEBUG
+#endif // DEBUG & WINDOWS
```
wiktork, line 201: Why only do the refcount on windows?

wiktork: Nvm i see this is answered below

src/Common.Lib/Macros.h
```diff
@@ -22,15 +22,21 @@
...
+// Asserts and optionally logs that a failure occured with the given hresult.
+// Note, this library does not define the implementation of this function. It
+// is up to the binary executable that links this library to supply a definition
+// otherwise, linkage errors will be expected.
+void AssertLogFailure(_In_ const WCHAR* wszError, ...);
+
     // Log an error without returning.
 #ifndef IfFailLog
 #define IfFailLog(EXPR) \
-    do { if (FAILED(hr = (EXPR))) { CLogging::LogError(_T("IfFailLog(") _T(#EXPR) _T(") failed in function ") __FUNCTIONT__ _T(": %d"), hr); } } while (false)
```
wiktork, line 28: Are we actually calling any of these in Common.lib (i'm not sure how this could have worked if we were)? Maybe we can move this to common.headers and skip the intermediate AssertLogFailure function.

src/Common.Lib/refcount.cpp
delmyers: Yes, exactly. If we use PAL calls in this lib, it means that it can't be used in multiple dlls on musl systems because the dynamic linker will get name clashes. I would like to build the Nagler instrumentation method without a dependency on the PAL.

---
In reply to: [573075612](https://github.com/microsoft/CLRInstrumentationEngine/pull/362#discussion_r573075612) [](ancestors = 573075612)

src/Common.Lib/Macros.h
delmyers: This could be moved to Common.Headers regardless, honestly. I left it here because it isn't technically a header-only library. You need to supply an implementation of AssertLogFailure.

Honestly, I mostly did this so that the header can be reused inside Nagler. But Nagler won't have access to CLogging, so I needed a work-around. Hence, AssertLogFailure. I could not use this in Nagler, but it would mean duplicating the dll exports, etc.




---
In reply to: [573096085](https://github.com/microsoft/CLRInstrumentationEngine/pull/362#discussion_r573096085) [](ancestors = 573096085)

src/InstrumentationEngine.Lib/Logging.cpp
```diff
@@ -232,4 +232,17 @@ void CLogging::XmlDumpHelper::WriteUlongNode(const WCHAR* name, const ULONG32 va
         << setfill(_T('0')); // fill with 0s
 
     m_result << m_childrenIndent << _T('<') << name << _T('>') << hex << setw(10) << value << _T("</") << name << _T('>') << _T("\r\n");
+}
+
+// Implemenation of Assert functions found in Macros.h
+void AssertLogFailure(_In_ const WCHAR* wszError, ...)
+{
```
WilliamXieMSFT, line 239: Should there be a IfNotInitRet(s_initialize); call?

src/InstrumentationEngine.ProfilerProxy.Lib/stdafx.h
```diff
@@ -13,12 +13,12 @@
...
-#include "../Common.Lib/tstring.h"
-#include "../Common.Lib/CriticalSectionHolder.h"
-#include "../Common.Lib/InitOnce.h"
-#include "../Common.Lib/Singleton.h"
+#include "../Common.Headers/tstring.h"
+#include "../Common.Headers/CriticalSectionHolder.h"
+#include "../Common.Headers/InitOnce.h"
+#include "../Common.Headers/Singleton.h"
+#include "../Common.Headers/SafeFindFileHandle.h"
+#include "..//Common.Headers/banned.h"
```
WilliamXieMSFT, line 21: >/ [](start = 13, length = 1)

nit extra slash

src/unix/inc/mincom/ccomptrs.h
```diff
@@ -0,0 +1,146 @@
...
+};
+
+
+//CComPtrBase provides the basis for all other smart pointers
+//The other smartpointers add their own constructors and operators
+template <class T>
+class CComPtrBase
+{
+protected:
+	CComPtrBase()
```
WilliamXieMSFT, line 30: >	 [](start = 0, length = 1)

nit tabs/spaces

src/unix/inc/mincom/ccomptrs.h
```diff
@@ -0,0 +1,146 @@
...
+	{
+		p = NULL;
+	}
+	CComPtrBase(int nNull)
+	{
+		(void)nNull;
```
WilliamXieMSFT, line 36: >(void)nNull; [](start = 2, length = 12)

what's the purpose of this expression?

src/unix/inc/mincom/ccomptrs.h
```diff
@@ -0,0 +1,146 @@
...
+		p = NULL;
+	}
+	CComPtrBase(T* lp)
+	{
+		p = lp;
+		if (p != NULL)
+			p->AddRef();
+	}
+public:
+	typedef T _PtrClass;
```
WilliamXieMSFT, line 46: >_PtrClass [](start = 11, length = 9)

not used?

src/InstrumentationEngine.Lib/Logging.cpp
delmyers: GetLoggingFlags will do that.

src/unix/inc/mincom/ccomptrs.h
delmyers: Probably not needed. I adapted this from the PAL. I can get rid of it.

src/unix/inc/mincom/ccomptrs.h
delmyers: It's a public type. It's a method of getting the type associated with the template. For example:

```CComPtr<CMyType>::_PtrClass obj;```

Will generate a local variable of type CMyType.

It is part of the ATL definition of this class, so I think I'll keep it.

File: src/Common.Lib/refcount.cpp [hunk 9]
```diff
@@ -4,16 +4,13 @@
 #include "stdafx.h"
 
 #include "refcount.h"
-#include "intrin.h"
-
 // TODO:
 // Clean this up. It was pulled from Debugger\inc and Debugger\common in order to get leak detection on
 // ref counted objects. I didn't just use the compiled lib becase it isn't clear to me where this code
 // is eventually going to live and if it turns out to be on a CDM server, the debugger lib won't exist.
```
WilliamXieMSFT, line 11: is this comment necessary?

WilliamXieMSFT (approved):
:shipit:

wiktork (approved):

PR #368, opened 2021-02-11
Fix or deliberately suppress more build warnings.

jander-msft:
Additional changes:
- Switch nupkg to use packaged icon instead of URL.
- Update MSBuild version in build pipeline to 16.0.


WilliamXieMSFT (approved):
:shipit:

PR #380, opened 2021-03-11
Ignore cert/strongname verification for test artifacts

WilliamXieMSFT:


build/CodeSign/certificate.ignore
```diff
@@ -15,4 +18,6 @@ bin\*\*.wixlib, Ignore wix artifact
 bin\*\*.wixpdb, Ignore wix artifact
 bin\*\*.msidir\*.cab, Ignore embedded cab files
 
-binaries-windows-release\*, Ignore downloaded binary artifacts
\ No newline at end of file
+binaries-windows-release\*, Ignore downloaded binary artifacts
+
+bin\*\newtonsoft.json.dll, Ignore Newtonsoft.Json
```
jander-msft, line 23: Can you briefly check why we have Newtonsoft.Json in the output? It's probably a test artifact, but I don't recall us doing JSON parsing in managed code.

build/CodeSign/certificate.ignore
```diff
@@ -3,6 +3,9 @@ obj\*, Ignore Intermediates
 bin\*\instrumentationengine.lib.tests.dll, Ignore test assembly
```
jander-msft, line 3: Not that I'm asking you to do this, but I wonder if we should move all test artifacts under a Tests folder and then just exclude the folder rather than excluding individual files.

build/CodeSign/certificate.ignore
```diff
@@ -3,6 +3,9 @@ obj\*, Ignore Intermediates
...
 bin\*\instrumentationengine.profilerproxy.tests.dll, Ignore test assembly
 bin\*\instrenginetests_anycpu.dll, Ignore test assembly
+bin\*\instrenginetests.dll, Ignore test assembly
+bin\*\testapprunner.dll, Ignore test assembly
+bin\*\testapprunner.exe, Ignore test assembly
```
jander-msft, line 8: I'll have to check, but we might be able to get rid of this apphost during build.

jander-msft (approved):

build/CodeSign/certificate.ignore
WilliamXieMSFT: I believe this came in from a dependency package, let me double check

build/CodeSign/certificate.ignore
WilliamXieMSFT: Absolutely agree

build/CodeSign/certificate.ignore
WilliamXieMSFT: Yep, it's a dependency of Microsoft.TestPlatform.TestHost: https://www.nuget.org/packages/Microsoft.TestPlatform.TestHost/16.10.0-preview-20210219-03

build/CodeSign/certificate.ignore
WilliamXieMSFT: Hmm, seems it's needed for the netcore3.1 tests

jander-msft: Should not be needed since the test code used the dll not the exe: https://github.com/microsoft/CLRInstrumentationEngine/blob/584c9f7bd3b66f774c527e6a31584fca9dcb09d4/tests/InstrEngineTests/InstrEngineTests/ProfilerHelpers.cs#L204

WilliamXieMSFT: I agree, but for some reason removing the apphost breaks the build: https://dev.azure.com/ms/CLRInstrumentationEngine/_build/results?buildId=146483&view=logs&j=eb496c93-f86e-53a7-8d99-a19210289e6d&t=b9d4457c-6a08-54a4-58a2-8ae830dc4b33

I'll try to investigate this later

PR #381, opened 2021-03-15
Re-include microbuild assets for signing

WilliamXieMSFT:


jander-msft:
Why is this change being made?

WilliamXieMSFT:
The microbuild package contains props/targets that are needed for signing plugin. Without these imports, our builds were erroring out with 'The target "SignFiles" does not exist in the project'

---
In reply to: [799580867](https://github.com/microsoft/CLRInstrumentationEngine/pull/381#issuecomment-799580867) [](ancestors = 799580867)

jander-msft (approved):

PR #384, opened 2021-03-25
Update docs for LogFilePath

WilliamXieMSFT:
Setting a single log file when multiple processes are profiled causes only the first process to log correctly. This documents the ability to set LogFilePath to a directory path which allows each process to generate its own log file.

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -40,7 +40,7 @@ These are the environment variables and values that aid in troubleshooting the C
 | Variable | Value | Description |
 |-|-|-|
 MicrosoftInstrumentationEngine_DebugWait|1|Suspends the process until the debugger is attached.
-MicrosoftInstrumentationEngine_FileLogPath|"[FULL PATH TO LOGGING FILE]"|File to host the event logs. This requires LogLevel to be set.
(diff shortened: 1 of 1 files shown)


docs/environment_variables.md
```diff
@@ -40,7 +40,7 @@ These are the environment variables and values that aid in troubleshooting the C
 | Variable | Value | Description |
 |-|-|-|
 MicrosoftInstrumentationEngine_DebugWait|1|Suspends the process until the debugger is attached.
-MicrosoftInstrumentationEngine_FileLogPath|"[FULL PATH TO LOGGING FILE]"|File to host the event logs. This requires LogLevel to be set.
+MicrosoftInstrumentationEngine_FileLogPath|"[FULL PATH TO LOGGING FILE]"|File to host the event logs. This requires LogLevel to be set. If multiple processes are profiled, then use a directory (end with slash) so that each process generates its own log file `ProfilerLog_[processId].txt`
```
jander-msft, line 43: I believe that `<path>\*` is also supported, which is probably safer from getting the ending slash from being accidentally forgotten or parsed off.

jander-msft (approved):

wiktork (approved):

docs/environment_variables.md
WilliamXieMSFT: Good catch, double checked the code and it allows `<path>\.` so I'll update this comment accordingly

Issue #382, opened 2021-03-17
Logging causes issues on exiting on Ubuntu


jakubch1:
I was testing on ubuntu 18.04 our instrumentation method with latest CLR IE. We see issues when logging is enabled (when logging is disabled it works fine). For investiagation we disabled all our logic so our instrumentation method is just returning S_OK on all callbacks.

The scenario is that we have 3 dotnet processes. 
1) first process is starting second process and prints exit code of second process
2) Second process is starting third process and prints exit code of third process
3) Third process is just printing "Hello world"

When CLR IE is not attached or CLR IE is attached but logging is disabled all works fine and exit codes are correctly set to 0.

In case that CLR IE is enabled and logging is enabled (through env variable) we can see such exit codes:
1) p1: 139, p2: 139, p3: 134
OR
2) p1: 139, p2:139, p3: hangs, after killing it we got 137
OR
3) p1: 139, p2:134, p3: hangs, after killing it we got 137'
OR
4) p1: 139, p2:134, p3: 134
OR
5) p1: 139, p2:134, p3: 139

If process is finishing with 134 we get such thing on output:
**Fatal error: glibc detected an invalid stdio handle**

Last messages in logs are:
```
LogMessage[14:54:25]:Ending ProfilerCallback AssemblyLoadFinished
LogMessage[14:54:25]:Starting ProfilerCallback Shutdown
LogMessage[14:54:25]:Sending event to Instrumentation Method
LogMessage[14:54:25]:Finished Sending event to Instrumentation Method. hr=0000
```

jakubch1:
Sample full output:

```
jakubch@Jachocho-5820:~/mstest/artifacts/test/8e3daa21-f12f-4424-95d6-b7cb2e9f353c/childprocesses$ dotnet test --settings dotnet.runsettings
  Determining projects to restore...
  All projects are up-to-date for restore.
  childprocesses -> /home/jakubch/mstest/artifacts/test/8e3daa21-f12f-4424-95d6-b7cb2e9f353c/childprocesses/bin/Debug/net5.0/childprocesses.dll
Test run for /home/jakubch/mstest/artifacts/test/8e3daa21-f12f-4424-95d6-b7cb2e9f353c/childprocesses/bin/Debug/net5.0/childprocesses.dll (.NETCoreApp,Version=v5.0)
Microsoft (R) Test Execution Command Line Tool Version 16.9.1
Copyright (c) Microsoft Corporation.  All rights reserved.

[...]

PR #387, opened 2021-03-30
Flush log file before closing it via smart pointer.

jander-msft:
The `m_pOutputFile` smart pointer has a deleter on it that closes the file. The CFileLoggerSink::CloseLogFile method used to clear the pointer (which would close the file), then try to flush the writes to the file, and close the file. On Windows, this is benign even though it is not the correct behavior; however, on Linux, this will generate a SIGSEGV and crash the process.

The fix is to flush the writes to the file and then clear the pointer (which will close the file).

closes: #382

cc: @jakubch1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Please format the changes as follows:
 ## 1.0.39
 + BugFixes:
   + Remove MethodInfo creation from JitComplete events. This fixes a race condition when accessing method info map.
+  + Fix CFileLoggerSink to flush writes before closing the file.
 
 ## 1.0.38
 + New:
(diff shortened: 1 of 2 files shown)


wiktork (approved):

jander-msft:
@jakubch1, would you be able to test this branch with your scenario that you mentioned in #382? I recreated it myself with a different instrumentation method and saw similar symptoms before this fix; after the fix, all of the processes now return exit code 0. I'd like for you to test with your scenario before I merge the PR to be sure I've fixed it correctly.

jakubch1:
I've verified that fix works. Could you please merge it and release version 39?

jakubch1 (approved):

PR #397, opened 2021-05-05
Use CModuleRefCount in NaglerInstrumentationMethod

schmittjoseph:
### Changes
- `NaglerInstrumentationMethod`
  - Replace `CComCoClass` with `CMoudleRefCount` and remove `CComObjectRootEx`.
  - Update DLL exports to use the Unix-friendly `DLLEXPORT` macro.
- `InstrEngineTests` 
  - Improve `ProfileHelpers` to include the process's launch command and stderr in the test output to help debug test failures.


tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -14,6 +14,14 @@ const WCHAR CInstrumentationMethod::TestScriptFileEnvName[] = L"Nagler_TestScrip
...
 const WCHAR CInstrumentationMethod::IsRejitEnvName[] = L"Nagler_IsRejit";
 
+ILOpcodeInfo ilOpcodeInfo[] =
+{
+#define OPDEF(ord, code, name,  opcodeLen, operandLen, type, alt, flags, pop, push) \
+    { name, (DWORD)opcodeLen, (DWORD)operandLen, ##type, alt, flags, (DWORD)pop, (DWORD)push},
+#include "ILOpcodes.h"
+#undef OPDEF
+};
+
```
WilliamXieMSFT, line 24: What were the reasons for moving this from NaglerInstrumentMethodEntry.h?

tests/InstrEngineTests/NaglerInstrumentationMethod/stdafx.h
```diff
@@ -46,4 +46,10 @@ using namespace ATL;
     do { if (FAILED(hr = (EXPR))) { ATLASSERT(FALSE); return hr; } } while (false)
 #endif
 
-#include "../../../src/Common.Headers/tstring.h"
\ No newline at end of file
+#include "../Common.Headers/tstring.h"
```
WilliamXieMSFT, line 49: >../ [](http://example.com/codeflow?start=10&length=3)

I don't think these parent directories are correct... Does it work because we're including the Common.Lib.Lib in the AdditionalDependencies?

WilliamXieMSFT:
>         { L"System.Boolean", ELEMENT_TYPE_BOOLEAN },

I think these widestrings should be wrapped using `_T("System.Boolean")`

---
Refers to: tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.h:53 in dcff7ab. [](commit_id = dcff7abe460a1c2fde47306638eb52a4a1d347f9, deletion_comment = False)

tests/InstrEngineTests/NaglerInstrumentationMethod/stdafx.h
jacdavis: Header files impact the compiler, libs impact the linker. So they should be unrelated. Are you sure tstring is actually getting pulled in via the precompiled header? Most likely, this isn't being found and anything using tstring is getting the header pulled in some other way.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
schmittjoseph: This was moved here since `dllmain.cpp` will indirectly pull in `NaglerInstrumentMethodEntry.h`, resulting in an error regarding `ilOpCodeInfo` already being defined. I'm open to alternative solutions to this if you have an idea.

tests/InstrEngineTests/NaglerInstrumentationMethod/stdafx.h
schmittjoseph: These paths work since the project has an `AdditionalIncludeDirectories` on `$(InstrumentationEngineApiInc);` though it is confusing. I'll go ahead and update these paths to be relative to this project itself.

schmittjoseph:
>         { L"System.Boolean", ELEMENT_TYPE_BOOLEAN },

Good catch. I looked into updating all instances in this dll and it'd require more changes than I'd prefer for this PR. I'll create another task to track this.

---
In reply to: [833089924](https://github.com/microsoft/CLRInstrumentationEngine/pull/397#issuecomment-833089924) [](http://example.com/codeflow?ancestors=833089924,833089924)

---
Refers to: tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.h:53 in dcff7ab. [](commit_id = dcff7abe460a1c2fde47306638eb52a4a1d347f9, deletion_comment = False)

WilliamXieMSFT (approved):
:shipit:

tests/InstrEngineTests/NaglerInstrumentationMethod/dllmain.cpp
```diff
@@ -5,17 +5,79 @@
...
+    {
+        return S_OK;
+    }
+
+    return S_FALSE;
 }
+
+// Returns a class factory to create an object of the requested type.
+_Check_return_
+STDAPI DLLEXPORT(DllGetClassObject, 12)(_In_ REFCLSID rclsid, _In_ REFIID riid, _Outptr_ LPVOID FAR* ppv)
```
delmyers, line 73: >DLLEXPORT [](http://example.com/codeflow?start=7&length=9)

Since you were not able to do it in this PR, coudl you add an issue to GitHub for exporting DllGetClassObject, etc. in Nagler for Linux?

delmyers:
> #include <atlcoll.h>

While you are here, could you take these out and replace them with:

```
#include "mincom/mincom.h"
#include "mincom/ccomptrs.h"
#define __ATL_MIN_COM__
```

Take note of what breaks because we will have to fix all of those breaks in order to get this built on Linux.

---
Refers to: tests/InstrEngineTests/NaglerInstrumentationMethod/stdafx.h:21 in 94b5ce2. [](commit_id = 94b5ce264493d64bd3833c3d311afc07de42da86, deletion_comment = False)

tests/InstrEngineTests/NaglerInstrumentationMethod/stdafx.h
```diff
@@ -46,4 +46,10 @@ using namespace ATL;
     do { if (FAILED(hr = (EXPR))) { ATLASSERT(FALSE); return hr; } } while (false)
 #endif
 
-#include "../../../src/Common.Headers/tstring.h"
\ No newline at end of file
+#include "../../../src/Common.Headers/tstring.h"
+#include "../../../src/Common.Lib/Macros.h"
+#include "../../../src/Common.Lib/refcount.h"
+#include "../../../src/InstrumentationEngine.Lib/ImplQueryInterface.h"
```
delmyers, line 52: We shouldn't be including anything from InstrumentationEngine.Lib since we will not link to that library. Is there a way that we can share this code instead? Maybe just move the header into the common headers?

delmyers (changes requested):
:clock1:

PR #398, opened 2021-05-10
Update ubuntu image to use newer CMake

delmyers:
Updates the ubuntu image to add the kitware repo so that
we can get a newer version of CMake.

Also updates the DockerLocalBuild.ps1 so that the local
Dockerfiles can be used to generate images for building Linux
rather than depending on images that are only available to Microsoft
build pipelines.


src/Scripts/DockerLocalImage.ps1
```diff
@@ -0,0 +1,90 @@
...
+    # Delete the current image
+    docker image rm $imageName
+
+    if (-not $?) {
+        Write-Error "Error removing image '$imageName'"
+        exit 1
+    }
+}
+
+$EnlistmentRoot = $(Resolve-Path -Path "$PSScriptRoot\..\..")
```
jander-msft, line 69: This is already a parameter to the script so this should not be necessary.

src/Scripts/DockerLocalBuild.ps1
```diff
@@ -12,23 +12,37 @@
 #       "musl" CLib for the Alpine 3.7 sample. More generally, use "gnu" build output for glibc-
 #       based distros and "musl" build output for musl-libc-based distros.
 
+<#
```
WilliamXieMSFT, line 15: Yay PowerShell comment blocks! Would you mind also moving the above Usage comment to this? probably in an `.EXAMPLE` section

src/Scripts/DockerLocalImage.ps1
```diff
@@ -0,0 +1,90 @@
...
+    exit 1
+}
+
+if ($exists -eq "yes") {
+    Write-Output "Image '$imageName' already exists."
+    if (-not $Rebuild) {
+        exit 0
+    }
+
+    # Delete the current image
```
jander-msft, line 60: I don't think the deletion is necessary; building the image should overwrite any existing image.

src/Scripts/DockerLocalBuild.ps1
```diff
@@ -12,23 +12,37 @@
...
 param
 (
+    # Optionally change the enlistment root. 
     [Parameter(Mandatory=$false)]
     [ValidateNotNullOrEmpty()]
     [ValidateScript({$_ -and (Test-Path $_)})]
     [String] $EnlistmentRoot = $(Resolve-Path -Path "$PSScriptRoot\..\.."),
 
+    # Indicates what version of the c runtime will be used for the build.
+    # This determines the version of Linux that will be used for the build.
```
WilliamXieMSFT, line 28: >version [](http://example.com/codeflow?start=26&length=7)

consider the term "flavor" or "distro" of Linux

src/Scripts/DockerLocalImage.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+$EnlistmentRoot = $(Resolve-Path -Path "$PSScriptRoot\..\..")
+$dockerDir = "$EnlistmentRoot\src\unix\docker\dockerfiles\build\$linux"
+
+if (-not (Test-Path "$dockerDir\DockerFile" -PathType Leaf)) {
+    Write-Error "Could not find path to required DockerFile in '$dockerDir"
+    exit 1
+}
+
+pushd $dockerDir
```
jander-msft, line 77: The `pushd` and `pop` might not be necessary. I think you can do `docker build $dockerDir -f "$dockerDir\Dockerfile" -t "$imageName"`

src/Scripts/DockerLocalImage.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+try {
+    docker build -t "$imageName" .
+    if (-not $?) {
+        Write-Error "Error building image '$imageName'"
+        exit 0
```
jander-msft, line 83: exit 1

src/Scripts/DockerLocalBuild.ps1
```diff
@@ -12,23 +12,37 @@
...
+    # What flavor to build. Debug or release. Default is Debug.
     [Parameter()]
     [ValidateNotNullOrEmpty()]
     [ValidateSet('Debug', 'Release')]
     [String] $Type="Debug",
 
+    # Use a local docker image. Default is to download the official image
+    # that is used for released builds.
+    [Parameter()]
+    [Switch] $LocalDockerImage,
```
jander-msft, line 43: nit: BuildDockerImage, the name LocalDockerImage almost sounds like I should point it to a local Docker image.

jander-msft (changes requested):

src/Scripts/DockerLocalBuild.ps1
```diff
@@ -41,17 +55,39 @@ param
...
+        exit 1
+    }
+}
+
 if ($CLib -eq "gnu")
 {
-    $BaseImage = "proddiagbuild.azurecr.io/clrie-build-ubuntu:latest"
+    if ($LocalDockerImage)
+    {
+        $BaseImage="clrielocal:gnu"
```
WilliamXieMSFT, line 70: >$BaseImage="clrielocal:gnu" [](http://example.com/codeflow?start=8&length=27)

Since the DockerLocalImage.ps1 script generates this, you could have the script `Write-Output $imageName` (you'll want to use Write-Verbose or Write-Host for logging) and store that value into $BaseImage.

src/Scripts/DockerLocalImage.ps1
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Usage:
+#   .\DockerLocalImage.ps1 [musl|gnu]
+#   Creates a local image based on the DockerFiles in this repository. The local image can be used for 
+#   performing local docker builds.
```
WilliamXieMSFT, line 7: Move this into .EXAMPLE

src/Scripts/DockerLocalImage.ps1
delmyers: If I understand correctly, if we don't change directories, then docker will add whatever files are under the current working directory to the image. The easiest way to change that is to switch to the directory that the dockerfile is in.

src/Scripts/DockerLocalBuild.ps1
delmyers: I thought about that, but I didn't want to use BuildDockerImage since it only builds the local image if it doesn't already exist.

jander-msft: BuildLocalImage? BuildImageLocally?

src/Scripts/DockerLocalBuild.ps1
delmyers: I don't need to add the parameters to the documentation comment because the parameters are documented inline.

src/Scripts/DockerLocalBuild.ps1
delmyers: That command didn't seem to work. But if I pipe the content of the file to docker build, then I don't need to set a context. I'll do that instead.

src/Scripts/DockerLocalImage.ps1
```diff
@@ -0,0 +1,79 @@
...
+
+$Expression = "Get-Content '$dockerDir\DockerFile' | docker build - -t '$imageName'"
+Write-Host $Expression
+Invoke-Expression $Expression | Write-Host
+if (-not $?) {
+    Write-Error "Error building image '$imageName'"
+    exit 1
+}
+
+Write-Host "$Created '$imageName'"
```
WilliamXieMSFT, line 77: >$ [](http://example.com/codeflow?start=12&length=1)

nit

WilliamXieMSFT (approved):
:shipit:

PR #401, opened 2021-05-14
Update alpine image

delmyers:
Update alpine docker image to use the latest version of cmake.

### DIFF
--- a/src/unix/docker/dockerfiles/build/alpine/Dockerfile
+++ b/src/unix/docker/dockerfiles/build/alpine/Dockerfile
@@ -5,6 +5,8 @@
 # REFERENCE: https://github.com/dotnet/coreclr/blob/v2.1.0-preview2-26406-04/buildpipeline/pipelines.json
 FROM microsoft/dotnet-buildtools-prereqs:alpine-3.6-3148f11-20171119021156
 
+WORKDIR /
+
 # Install .NET Core 2.1 SDK
 # REFERENCE: https://github.com/dotnet/dotnet-docker/blob/master/2.1/sdk/alpine3.10/amd64/Dockerfile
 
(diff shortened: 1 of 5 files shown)


jander-msft:
Alpine 3.9 uses clang 5.0 (https://pkgs.alpinelinux.org/package/v3.9/main/x86_64/clang) which it too new for ABI compatibility with the PAL package.

delmyers:
> Alpine 3.9 uses clang 5.0 (https://pkgs.alpinelinux.org/package/v3.9/main/x86_64/clang) which it too new for ABI compatibility with the PAL package.

Is the incompatibility documented somewhere?

jander-msft:
> > Alpine 3.9 uses clang 5.0 (https://pkgs.alpinelinux.org/package/v3.9/main/x86_64/clang) which it too new for ABI compatibility with the PAL package.
> 
> Is the incompatibility documented somewhere?

The compiler switch is barely documented [here](https://clang.llvm.org/docs/ClangCommandLineReference.html) as `-fclang-abi-compat=<version>`

The ABI versions are [here](https://github.com/llvm/llvm-project/blob/5bb7e81c64bd29edd1c9ebadd4e4717919def0bf/clang/include/clang/Basic/LangOptions.h#L136) in the source code. You might have to go back to the 5.0 release to see what the options were at that time.

All of the language around the ABI back compat descriptions seem to use "attempt" a lot so there isn't much of a guarantee that a later clang version using the -fclang-abi-compat switch for 4.0 has any guarantee that it would actually generate back compat code.

src/unix/docker/dockerfiles/build/alpine/Dockerfile
```diff
@@ -38,4 +40,20 @@ RUN dotnet help
 
 # CLR Instrumentation Engine Build Prerequisites
 RUN apk update \
-    && apk add --no-cache libxml2-dev
\ No newline at end of file
+    && apk add --no-cache libxml2-dev
+
+# Install newer version of cmake. We need to run make commands inside the directory
+# that it is extracted to, so WORDIR is the easiest context to use.
```
jander-msft, line 46: nit: `WORKDIR`

src/unix/docker/dockerfiles/build/alpine/Dockerfile
```diff
@@ -38,4 +40,20 @@ RUN dotnet help
...
+WORKDIR /usr/share/cmakeinst
+RUN wget -O cmake.tar.gz https://github.com/Kitware/CMake/releases/download/v3.20.2/cmake-3.20.2.tar.gz \
```
jander-msft, line 48: nit: Maybe refactor the the version into an ENV command so that it can be reused here and in the `cd cmake-3.20.2` lines. Here's an example that dotnet-docker uses for this: https://github.com/dotnet/dotnet-docker/blob/720a40b1aa348ad4c9fc5575668090849ac3172b/src/monitor/5.0/alpine/amd64/Dockerfile#L8

jander-msft (approved):

src/unix/docker/dockerfiles/build/alpine/Dockerfile
jander-msft, line 48: Consider having some way of reading the cgmanifest file for the URL or version. That way we don't declare to CG that we are using one version and are actually using another version.

This is not something you have to do now, but is something to think about (if possible, since I know this is building within a Docker container).

src/unix/docker/dockerfiles/build/alpine/Dockerfile
```diff
@@ -38,4 +40,29 @@ RUN dotnet help
...
+# Install newer version of cmake. We need to run make commands inside the directory
+# that it is extracted to, so WORKDIR is the easiest context to use.
+WORKDIR /usr/share/cmakeinst
+
+# Copy manifest json to the target so that we can extract the download url and
+# version of cmake
+COPY "components/cmake/cgmanifest.json" /usr/share/cmakeinst
+COPY "scripts/readattr.sh" /usr/share/cmakeinst
+RUN chmod 775 ./readattr.sh \
+    && cmakeurl=$(./readattr.sh cgmanifest.json 'DownloadUrl') \
```
jander-msft, line 54: I don't know how much more you want to iterate on this but a more concise and widely used option might be to use the [jq](https://pkgs.alpinelinux.org/package/v3.6/main/x86_64/jq) command-line tool.

jander-msft: By concise, I mean you won't need a custom shell script to do the parsing.

src/unix/docker/context/scripts/readattr.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+manifest=$1
```
WilliamXieMSFT, line 3: Mind adding a copyright header?

```
```

WilliamXieMSFT (changes requested):
:clock1:

WilliamXieMSFT (approved):
:shipit: (with suggestions)

WilliamXieMSFT (approved):
:shipit:

PR #402, opened 2021-05-15
Decouple XmlDocWrapper from PAL

### LINKED ISSUE #359: Decouple XML reading from the CLR PAL
Nagler needs to read XML to run its scripts. Currently, it uses msxml to read xml, but we have code inside InstrumentationEngine that can read xml using msxml or libxml. Unfortunately, this code is coupled with the PAL to do encoding/decoding and to pass strings around as BSTR types. This is not necessary. We should decouple this xml reading code from bstrs, and use iconv to do the encoding/decoding if necessary on Linux. The encoding/decoding will not be needed on Windows.

delmyers:
Moves XmlDocWrapper out of InstrumentationEngine.Lib so that it can
be used by Nagler as well, as part of the effort to get Nagler running
on Windows and Linux.

There are multiple steps to this.

1) Remove CEncoding and replace it with systemstring
    -XmlDocWrapper depended on CEncoding, but CEncoding depended
     on ATL libraries that depended on the PAL. CEncoding has been
     been replaced by systemstring which uses only functions that
     are native to the OS.
2) Remove XmlDocWrapper's dependency on BSTR and AutoVectorPointer
    - tstring is a better cross-plat solution. XmlDocWrapper now uses
      systemstring and tstring instead of depending on Windows types.
3) Move XmlDocWrapper into Common.Lib
4) While I was here, I simplified ConfigurationLoader to
    use XmlDocWrapper so that all of our xml reading is consistent.

To support these changes, additional cross-platform unit tests have been
added. This involves:

1) Adding a dependency on googletest for running cross-plat c++ tests.
2) Updating the minimum version of cmake to 3.14 (needed for googletest)
4) Adding unit tests for systemstring, and XmlDocWrapper
5) Moving the ConfigurationLoaderTests into a new cross-plat test.

While doing this work, I found a couple of bugs:

1) XmlDocWrapper behaved differently on Windows and Linux. This has been
    fixed.
2) Previous move of ImplQueryInterface makes an undefined reference to
    IID_IUnknown. The necessary IID has been added for Linux systems.

closes #359


delmyers:
Depends on PR #401 for the pr builds to be successful.

build/ClrInstrumentationEngine.cpp.props
```diff
@@ -34,11 +34,17 @@
     -->
   <PropertyGroup>
     <CodeAnalysisWarningsToDisable>/wd25007 /wd25021 /wd25031 /wd25046 /wd25090 /wd25129 /wd25164 /wd25165 /wd25166</CodeAnalysisWarningsToDisable>
+    <AdditionalWarningsToDisable>/wd4103</AdditionalWarningsToDisable>
```
delmyers, line 37: > <AdditionalWarningsToDisable>/wd4103</AdditionalWarningsToDisable> [](http://example.com/codeflow?start=3&length=67)

Undo this change. It was added because of a break in the C++ compiler.

Config.xml
```diff
@@ -0,0 +1,10 @@
+<xml/>
```
delmyers, line 1: Remove

src/build.sh
```diff
@@ -110,6 +109,59 @@ locate_llvm_exec()
...
+    cmakeMinor=$(echo $cmakeVersion | sed 's/^.*[^0-9]*\.\([0-9]*\)\..*$/\1'/)
+    if (( $cmakeMajor > 3 )); then
+        if (( $cmakeMinor >= 14 )); then
+            # found sufficient cmake, continue.
+            echo $cmakeLocation
+            return
+        fi
+    fi
+
+    echo "Insufficient cmake version. Ensure cmake version 3.8 or later is installed"
```
delmyers, line 160: >3.8 [](http://example.com/codeflow?start=59&length=3)

3.14

src/Common.Lib/systemstring.cpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        result = buffer.get();
+        return S_OK;
+    }
+#else
+
+    SystemString::SystemString(_In_ const string& original)
+    {
+        m_hresult = FromString(original.c_str(), *this);
```
delmyers, line 162: >m_hresult = FromString(original.c_str(), *this) [](http://example.com/codeflow?start=8&length=47)

It would probably make more sense to reverse the order of these. That is, have FromString call the argumented constructor.

src/build.sh
```diff
@@ -110,6 +109,59 @@ locate_llvm_exec()
     fi
 }
 
+get_download_url()
+{
+    manifest=$EnlistmentRoot/build/components/$1/cgmanifest.json
```
schmittjoseph, line 114: >$EnlistmentRoot/build/components/$1/cgmanifest.json [](http://example.com/codeflow?start=13&length=51)

nit: this path and instances of `$manifest` should be enclosed in quotes incase `$EnlistmentRoot` contains spaces.

src/Common.Lib/systemstring.h
```diff
@@ -0,0 +1,119 @@
...
+#else
+    using osstring = std::wstring;
+#endif
+
+    /// <summary>
+    /// Class for converting between operating system strings and standard strings
+    /// or wide-character (Windows) strings. Useful for converting strings between
+    /// Linux and Windows.
+    /// </summary>
+    class SystemString : public osstring
```
wiktork, line 35: Could this be done with a typedef similar to tstring instead of inheritence? (And you could potentially expand the set of static functions on CEncoding or just make all the conversion functions on this class static). 
Imo having a string class with an error state is pretty confusing.

src/Tests/InstrumentationEngineLibTests/InstrumentationEngineLibTests.vcxproj
```diff
@@ -0,0 +1,139 @@
...
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{B1FFBD51-80A9-41D6-BC54-07B7E0DE7CF6}</ProjectGuid>
+    <Keyword>Win32Proj</Keyword>
+    <WindowsTargetPlatformVersion>10.0.19041.0</WindowsTargetPlatformVersion>
```
wiktork, line 26: This is 10.0.18362.0 for other projects. We probably want to keep it consistent.

src/CMakeLists.txt
```diff
@@ -221,6 +225,15 @@ function(generate_exports_file inputFilename versionName outputFilename)
                               PROPERTIES GENERATED TRUE)
 endfunction()
 
+macro (fetch_google_test)
+    include(FetchContent)
+    FetchContent_Declare(
+        googletest
+        URL ${GOOGLE_TEST_URL}
```
wiktork, line 232: So where does this actually pull from?

delmyers: See build/components/googletest/cgmanifest.json and src/build.sh. Build.sh reads the cgmanifest.json file to find the url snd sets this variable.

We need the cgmanifest file because our OSS component detection tools can't detect components downloaded by cmake.

build/components/readme.md
```diff
@@ -0,0 +1,12 @@
+# Build-Time Components
+
+This folder contains component manafests for build-time components that cannot be downloaded via nuget.
```
WilliamXieMSFT, line 3: >manafests [](http://example.com/codeflow?start=31&length=9)

nit spelling

Config.xml
```diff
@@ -0,0 +1,10 @@
...
+<InstrumentationEngineConfiguration>
```
WilliamXieMSFT, line 2: >InstrumentationEngineConfiguration [](http://example.com/codeflow?start=1&length=34)

Can this be moved to the "test" folder or rename the folder to something like "TestConfig"?

src/Common.Lib/systemstring.cpp
```diff
@@ -0,0 +1,241 @@
...
+        // scan to a resonable length for the end of the string.
+        for (i = 0; i < MAX_STRING_LEN; ++i, ++p)
+        {
+            if (*p == (WCHAR)0)
+            {
+                break;
+            }
+        }
+
+        if (i > MAX_STRING_LEN)
```
WilliamXieMSFT, line 87: >> [](http://example.com/codeflow?start=14&length=1)

`==` ? Otherwise, the for-loop condition will never make this true

src/Common.Lib/systemstring.cpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+
+        Ciconv icnv(iconv_open("UTF-8", "UTF-16LE"));
+
+        // include the null terminator in the input string.
+        size_t inbytes = (i+1)*2;
+        size_t outbytes = (i+1)*4;
+        unique_ptr<char[]> buffer(new char[outbytes]);
+        char* input = (char*)lpzwStr;
```
WilliamXieMSFT, line 100: >char [](http://example.com/codeflow?start=23&length=4)

thoughts on using `std::byte`?

WilliamXieMSFT: Oh nvm, it's only available in C++17, and I think we're on C++11

src/Common.Lib/systemstring.h
```diff
@@ -0,0 +1,119 @@
...
+
+        /// <summary>
+        /// Construct a system string from a UTF8 (Linux) string.
+        /// </summary>
+        /// <param name="lpzStr">The original string</param>
+        SystemString(_In_ const CHAR* lpzStr) _CHARBASE(lpzStr);
+
+        /// <summary>Returns true if an error occurred when constructing this object </summary>
+
+        bool HasError() const { return m_hresult < 0; }
```
WilliamXieMSFT, line 72: nit newline

src/Common.Lib/XmlDocWrapper.cpp
```diff
@@ -0,0 +1,194 @@
...
+#ifndef PLATFORM_UNIX
+using namespace ATL;
+#endif
+
+namespace CommonLib
+{
+    CXmlDocWrapper::CXmlDocWrapper()
+    #ifdef PLATFORM_UNIX
+        : m_cleanupDoc(),
+        m_pDocument(nullptr)
```
WilliamXieMSFT, line 17: >m_pDocument(nullptr) [](http://example.com/codeflow?start=8&length=20)

should `m_pDocument` be initialized for windows also?

src/Common.Lib/XmlNode.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "systemstring.h"
+
+namespace CommonLib
+{
+    #ifndef PLATFORM_UNIX
+    CXmlNode::CXmlNode(_In_ IXMLDOMNode* pNode)
+    {
+        m_pNode = pNode;
+    }
+    #else
```
WilliamXieMSFT, line 15: >#else [](http://example.com/codeflow?start=4&length=5)

nit, wrap just the function signature

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -507,28 +507,29 @@ DWORD WINAPI CProfilerManager::ParseAttachConfigurationThreadProc(
...
     {
         CLogging::LogError(_T("CConfigurationLocator::ParseAttachConfigurationThreadProc - Invalid configuration. Root element should be InstrumentationEngineConfiguration"));
         return E_FAIL;
     }
 
     while (pCurrChildNode != nullptr)
     {
-        IfFailRet(pCurrChildNode->GetName(&bstrCurrChildNodeName));
-        if (wcscmp(bstrCurrChildNodeName, _T("InstrumentationEngine")) == 0)
+        tstring currChildNodeName;
```
WilliamXieMSFT, line 521: >currChildNodeName [](http://example.com/codeflow?start=16&length=17)

this is already defined in the outer scope

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -11,10 +11,10 @@
 
 
 MicrosoftInstrumentationEngine::CInstrumentationMethod::CInstrumentationMethod(
-    _In_ BSTR bstrModuleFolder,
-    _In_ BSTR bstrName,
-    _In_ BSTR bstrDescription,
-    _In_ BSTR bstrModule,
+    _In_ LPCWSTR bstrModuleFolder,
```
WilliamXieMSFT, line 14: >bstr [](http://example.com/codeflow?start=17&length=4)

nit update argument names

src/Tests/CommonLibTests/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+project(Common.LibTests)
+find_package(LibXml2 REQUIRED)
+
+# the midl directory is needed to replace the rpc headers referenced by midl generated headers.
+include_directories(. ${REPOSITORY_ROOT}/src/unix/inc/midl ${REPOSITORY_ROOT}/src ${LIBXML2_INCLUDE_DIR})
+
+build_init(CPP Common.LibTests)
+
+#note: we purposfully exclude windows-only files,
```
WilliamXieMSFT, line 17: >purposfully [](http://example.com/codeflow?start=10&length=11)

nit spelling

WilliamXieMSFT:
> <?xml version="1.0" encoding="utf-8"?>

nit copyright header
```
<!-- Copyright (c) Microsoft Corporation. All rights reserved.
```

---
Refers to: src/Tests/CommonLibTests/CommonLibTests.vcxproj:1 in 13ba350. [](commit_id = 13ba3505673ecab11c922529390aa3a6ada3f635, deletion_comment = False)

src/Tests/CommonLibTests/StringConversionTests.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "Common.Lib/systemstring.h"
+
+using namespace std;
+using namespace CommonLib;
+TEST(StringConversionTests, u8u16test) {
+    SystemString ss = u8"①Ⅻㄨㄩ 啊阿鼾齄丂丄狚狛狜狝﨨﨩ˊˋ˙– ⿻〇㐀㐁䶴䶵";
+    EXPECT_FALSE(ss.HasError());
+    tstring u16test = _WT("①Ⅻㄨㄩ 啊阿鼾齄丂丄狚狛狜狝﨨﨩ˊˋ˙– ⿻〇㐀㐁䶴䶵");
+    tstring tresult;
+    EXPECT_EQ(0, ss.ToString(tresult));
```
WilliamXieMSFT, line 14: >EXPECT_EQ [](http://example.com/codeflow?start=4&length=9)

EXPECT_OK

tests/InstrumentationEngine.Lib.Tests/InstrumentationEngine.Lib.Tests.vcxproj
```diff
@@ -91,7 +91,6 @@
     <ClInclude Include="TestLoggingHost.h" />
   </ItemGroup>
   <ItemGroup>
-    <ClCompile Include="ConfigurationLoaderTests.cpp" />
```
WilliamXieMSFT, line 94: >ConfigurationLoaderTests [](http://example.com/codeflow?start=24&length=24)

was this removal intentional?

WilliamXieMSFT:
> <?xml version="1.0" encoding="utf-8"?>

copyright header

---
Refers to: src/Tests/InstrumentationEngineLibTests/InstrumentationEngineLibTests.vcxproj:1 in 13ba350. [](commit_id = 13ba3505673ecab11c922529390aa3a6ada3f635, deletion_comment = False)

src/Tests/InstrumentationEngineLibTests/ConfigurationLoaderTests.cpp
```diff
@@ -0,0 +1,133 @@
...
+        remove(fileName);
+    }
+
+    sysfstream output;
+    output.open(fileName, ios_base::app | ios_base::out);
+    WCHAR error[256] = { 0 };
+
+    ASSERT_TRUE(!output.fail() && !output.bad());
+
+    //output << _ST("<?xml?>") << endl;
```
WilliamXieMSFT, line 74: >//output << _ST("<?xml?>") << endl; [](http://example.com/codeflow?start=4&length=35)

nit

tests/InstrumentationEngine.Lib.Tests/InstrumentationEngine.Lib.Tests.vcxproj
delmyers: Yes. It was moved to the cross plat tests.

delmyers: Looks like I forgot to delete the cpp file from this directory though.

Config.xml
delmyers: This was accidentally added from the output of a test that I ran. I will remove it.

src/Common.Lib/XmlDocWrapper.cpp
delmyers: Doesn't need to be because it isn't a pointer. CComPtr will initualize it to null. A cleaner way to do this might be to use initialization in the header file, though.

src/Common.Lib/systemstring.cpp
WilliamXieMSFT: maybe `if (*p != (WCHAR)0)`

delmyers: Forgot to push this change. Checking for not null is probably a bad way to do this, because it could be null by coincidence.

WilliamXieMSFT (approved):
:shipit:

PR #403, opened 2021-05-27
Adding docker file for centos 7

jakubch1:
Verified that build works on latest develop and on branch: dev/delmyers/nopalxml

Also checked build and that library works on ubuntu 18.04 commit: 8a8dec8

Using this docker image build should be invoke like this:
`scl enable llvm-toolset-7 './src/build.sh Release'`

`llvm-toolset-7` is environment where clang 4 is installed.


jakubch1:
Verified that build works on latest develop and on branch: dev/delmyers/nopalxml

Also checked build and that library works on ubuntu 18.04 commit: https://github.com/microsoft/CLRInstrumentationEngine/commit/8a8dec826e08ce632d1f0b179bfbe567c48c3d81

Using this docker image build should be invoke like this: 
`scl enable llvm-toolset-7 './src/build.sh Release'`

`llvm-toolset-7` is environment where clang 4 is installed.

PR #404, opened 2021-06-21
Add support for Win7

WilliamXieMSFT:
This includes a version bump to 1.0.41

### DIFF
--- a/src/Common.Lib/PathUtils.cpp
+++ b/src/Common.Lib/PathUtils.cpp
@@ -3,6 +3,7 @@
 
 #include "stdafx.h"
 #include "PathUtils.h"
+#include "StringUtils.h"
 
 
 namespace CommonLib
@@ -35,5 +36,21 @@ namespace CommonLib
         strFormat << strPath2;
 
         return strFormat.str();
+    }
+
+    HRESULT PathUtils::SafePathAppend(LPWSTR pwszPath, LPCWSTR pwszMore, size_t cBounds)
+    {
+        // If PathAppend fails, the destination buffer will be cleared. Check bounds before appending.
+        if (StringUtils::WStringLen(pwszPath) + StringUtils::WStringLen(pwszMore) >= cBounds)
+        {
+            return E_BOUNDS;
+        }
+
+        if (!PathAppend(pwszPath, pwszMore))
+        {
+            return E_FAIL;
+        }
+
+        return S_OK;
     }
 }
--- a/src/Common.Lib/PathUtils.cpp
+++ b/src/Common.Lib/PathUtils.cpp
@@ -3,7 +3,6 @@
 
 #include "stdafx.h"
 #include "PathUtils.h"
-#include "StringUtils.h"
 
 
 namespace CommonLib
@@ -36,21 +35,5 @@ namespace CommonLib
         strFormat << strPath2;
 
         return strFormat.str();
-    }
-
-    HRESULT PathUtils::SafePathAppend(LPWSTR pwszPath, LPCWSTR pwszMore, size_t cBounds)
-    {
-        // If PathAppend fails, the destination buffer will be cleared. Check bounds before appending.
-        if (StringUtils::WStringLen(pwszPath) + StringUtils::WStringLen(pwszMore) >= cBounds)
-        {
-            return E_BOUNDS;
-        }
-
-        if (!PathAppend(pwszPath, pwszMore))
-        {
-            return E_FAIL;
-        }
-
-        return S_OK;
     }
 }
--- a/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
+++ b/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
@@ -8,7 +8,7 @@
 #ifndef PLATFORM_UNIX
 #include "SignatureValidator.h"
 #endif
-#include "StringUtils.h"
+#include "../Common.Lib/PathUtils.h"
 
 
 MicrosoftInstrumentationEngine::CInstrumentationMethod::CInstrumentationMethod(
@@ -83,7 +83,7 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
 
     if ((m_bstrModuleFolder.Length() >= MAX_PATH))
     {
-        CLogging::LogError(_T("CInstrumentationMethod::Initialize - module folder path is too long, PID: %u"), GetCurrentProcessId());
+        CLogging::LogError(_T("CInstrumentationMethod::Initialize - module folder path is too long: '%s', PID: %u"), m_bstrModuleFolder.m_str, GetCurrentProcessId());
         return E_BOUNDS;
     }
 
@@ -96,17 +96,13 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
         return HRESULT_FROM_WIN32(dwLastError);
     }
 
-    // If PathAppend fails, the destination buffer will be cleared. Check bounds before appending.
-    if (StringUtils::WStringLen(wszModuleFullPath) + StringUtils::WStringLen(m_bstrModule) >= MAX_PATH)
+    hr = PathUtils::SafePathAppend(wszModuleFullPath, m_bstrModule, MAX_PATH);
+    if (FAILED(hr))
     {
-        CLogging::LogError(_T("CInstrumentationMethod::Initialize - method configuration fullpath is too long: '%s' + '%s', PID: %u"), m_bstrModuleFolder.m_str, m_bstrModule.m_str, GetCurrentProcessId());
-        return E_BOUNDS;
-    }
-
-    if (!PathAppend(wszModuleFullPath, m_bstrModule))
-    {
-        CLogging::LogError(_T("CInstrumentationMethod::Initialize - unable to append module to method configuration folder: '%s' + '%s', PID: %u"), m_bstrModuleFolder.m_str, m_bstrModule.m_str, GetCurrentProcessId());
-        return E_FAIL;
+        CLogging::LogError(
+            _T("CInstrumentationMethod::Initialize - failed to append method dll to method configuration folder, PID: %u, hr: %x, target: '%s' + '%s'"),
+            GetCurrentProcessId(), hr, m_bstrModuleFolder.m_str, m_bstrModule.m_str);
+        return hr;
     }
 
     m_hmod = ::LoadLibrary(wszModuleFullPath);
--- a/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
+++ b/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
@@ -8,7 +8,7 @@
 #ifndef PLATFORM_UNIX
 #include "SignatureValidator.h"
 #endif
-#include "../Common.Lib/PathUtils.h"
+#include "StringUtils.h"
 
 
 MicrosoftInstrumentationEngine::CInstrumentationMethod::CInstrumentationMethod(
@@ -83,7 +83,7 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
 
     if ((m_bstrModuleFolder.Length() >= MAX_PATH))
     {
-        CLogging::LogError(_T("CInstrumentationMethod::Initialize - module folder path is too long: '%s', PID: %u"), m_bstrModuleFolder.m_str, GetCurrentProcessId());
+        CLogging::LogError(_T("CInstrumentationMethod::Initialize - module folder path is too long, PID: %u"), GetCurrentProcessId());
         return E_BOUNDS;
     }
 
@@ -96,7 +96,7 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
         return HRESULT_FROM_WIN32(dwLastError);
     }
 
-    hr = PathUtils::SafePathAppend(wszModuleFullPath, m_bstrModule, MAX_PATH);
+    hr = StringUtils::SafePathAppend(wszModuleFullPath, m_bstrModule, MAX_PATH);
     if (FAILED(hr))
     {
         CLogging::LogError(
--- a/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
+++ b/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
@@ -84,8 +84,6 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
     memset(wszModuleFullPath, 0, MAX_PATH);
     wcscpy_s(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder);
 
-    // For Windows 7 support, we cannot use PathCch functions.
-    //PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
     WCHAR wszBuffer[2 * MAX_PATH]; // set to 2x the MAX_PATH in order to prevent buffer overflow.
     if (!PathCanonicalize(wszBuffer, wszModuleFullPath) ||
         !PathAppend(wszBuffer, m_bstrModule) ||
--- a/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
+++ b/src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
@@ -8,6 +8,7 @@
 #ifndef PLATFORM_UNIX
 #include "SignatureValidator.h"
 #endif
+#include "StringUtils.h"
 
 
 MicrosoftInstrumentationEngine::CInstrumentationMethod::CInstrumentationMethod(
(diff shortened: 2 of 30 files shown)


karpinsn:
@WilliamXieMSFT, I would also remove pathcch.lib from AdditionalDependencies in InstrumentationEngine.vcxproj. That way folks are less likely to accidentally pull in Win8+ code

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -83,7 +83,17 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
     WCHAR wszModuleFullPath[MAX_PATH];
     memset(wszModuleFullPath, 0, MAX_PATH);
     wcscpy_s(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder);
-    PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
+
+    // For Windows 7 support, we cannot use PathCch functions.
+    //PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
+    if (!wcslen(wszModuleFullPath) + m_bstrModule.Length() < MAX_PATH - 1 ||
```
wiktork, line 89: Did you mean `!`**(**`wcslen(...) + m_bstrModule.Length()`**)**

jander-msft, line 89: I assume you are doing `MAX_PATH - 1` to allow the null terminator. I don't think this accounts for the directory separators?

src/InstrumentationEngine.ProfilerProxy/dllmain.cpp
```diff
@@ -112,8 +113,19 @@ namespace ProfilerProxy
         WCHAR wszEngineFullPath[MAX_PATH];
         IfFailRetErrno_Proxy(wcscpy_s(wszEngineFullPath, MAX_PATH, wszProfilerPath));
 
-        IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, wszVersionFolder));
-        IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, profilerRelativeFileName));
+        if (!wcslen(wszEngineFullPath) + wcslen(wszVersionFolder) + wcslen(profilerRelativeFileName) < MAX_PATH - 1||
```
jander-msft, line 116: I assume you are doing MAX_PATH - 1 to allow the null terminator. I don't think this accounts for the directory separators?

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -83,7 +83,17 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
...
+        !PathCombine(wszModuleFullPath, wszModuleFullPath, m_bstrModule) ||
```
jander-msft, line 90: Is it safe for the destination buffer to be the same as one of the other inputs? I seem to remember that it is not always safe to do this, but i could be thinking of other file path APIs. Question applies to all of the new functions you are using.

WilliamXieMSFT:
Removing, thanks!

---
In reply to: [865344723](https://github.com/microsoft/CLRInstrumentationEngine/pull/404#issuecomment-865344723) [](http://example.com/codeflow?ancestors=865344723,865344723)

WilliamXieMSFT:
>         IfFailRet_Proxy(PathCchAppend(wszProfilerPath, MAX_PATH, instrumentationEngineFolder));

todo

---
Refers to: src/InstrumentationEngine.ProfilerProxy/dllmain.cpp:267 in 823edf5. [](commit_id = 823edf5ab7bd3a183f518f72391a2faa849964cd, deletion_comment = False)

WilliamXieMSFT:
>         IfFailRet_Proxy(PathCchAppend(wszProfilerPath, MAX_PATH, versionFolder.c_str()));

todo

---
Refers to: src/InstrumentationEngine.ProfilerProxy/dllmain.cpp:295 in 823edf5. [](commit_id = 823edf5ab7bd3a183f518f72391a2faa849964cd, deletion_comment = False)

WilliamXieMSFT:
>         IfFailRet_Proxy(PathCchAppend(wszProfilerPath, MAX_PATH, profilerRelativeFileName));

todo

---
Refers to: src/InstrumentationEngine.ProfilerProxy/dllmain.cpp:301 in 823edf5. [](commit_id = 823edf5ab7bd3a183f518f72391a2faa849964cd, deletion_comment = False)

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
WilliamXieMSFT: yep or swap `<` with `>`

WilliamXieMSFT: Correct about the null terminator, and there's a slight problem with PatchCanonicalize which can cause paths to be longer so I'll need to refactor this with more thought

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
WilliamXieMSFT: It is not safe! I'll refactor

src/InstrumentationEngine.ProfilerProxy/dllmain.cpp
```diff
@@ -108,12 +109,26 @@ namespace ProfilerProxy
...
+        //IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, wszVersionFolder));
+        //IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, profilerRelativeFileName));
+        WCHAR wszBuffer[2 * MAX_PATH]; // set to 2x the MAX_PATH in order to prevent buffer overflow.
+        if (!PathCanonicalize(wszBuffer, wszEngineFullPath) ||
+            !PathAppend(wszBuffer, wszVersionFolder) ||
+            !PathAppend(wszBuffer, profilerRelativeFileName) ||
+            wcslen(wszBuffer) >= MAX_PATH)
+        {
+            dError = GetLastError();
+            eventLogger.LogError(_T("dllmain::HasProfilerDll - Unable to generate fullpath to CLRIE dll: '%s' + '%s' + '%s'", wszEngineFullPath, wszVersionFolder, profilerRelativeFileName));
```
WilliamXieMSFT, line 127: >, [](http://example.com/codeflow?start=124&length=1)

nit misplaced `)`

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -83,8 +83,20 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
...
+    // For Windows 7 support, we cannot use PathCch functions.
+    //PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
+    WCHAR wszBuffer[2 * MAX_PATH]; // set to 2x the MAX_PATH in order to prevent buffer overflow.
+    if (!PathCanonicalize(wszBuffer, wszModuleFullPath) ||
+        !PathAppend(wszBuffer, m_bstrModule) ||
+        wcslen(wszBuffer) >= MAX_PATH)
+    {
+        DWORD dwLastError = GetLastError();
+        CLogging::LogError(_T("CInstrumentationMethod::Initialize - unable to generate method configuration fullpath '%s' + '%s', PID: %u"), m_bstrModuleFolder.m_str, m_bstrModule.m_str, GetCurrentProcessId());
+        return dwLastError == 0 ? E_FAIL : HRESULT_FROM_WIN32(dwLastError);
```
delmyers, line 96: >dwLastError == 0  [](http://example.com/codeflow?start=15&length=17)

Does one of the above not set the last error? If both do, then this check is not necessary. If one doesn't then I think HRESULT_FROM_WIN32 may retrieve an error value for  something other than the functions that were called above.

That is, if this check is necessary, then GetLastError() may be wrong.

It seems that we should either always just return HRESULT_FROM_WIN32(dwLastError), or always return E_FAIL, depending on the documentation of PathCanonicalize and PathAppend.

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -83,8 +83,20 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
...
+    }
+
+    wcscpy_s(wszModuleFullPath, MAX_PATH, wszBuffer);
```
delmyers, line 99: >wcscpy_s(wszModuleFullPath, MAX_PATH, wszBuffer); [](http://example.com/codeflow?start=4&length=49)

This string copy isn't strictly necessary, and probably wastes time. If the path append was successful, then wszBuffer is a valid string, and can just be used throughout the rest of the funciton.

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -83,8 +83,20 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
     WCHAR wszModuleFullPath[MAX_PATH];
     memset(wszModuleFullPath, 0, MAX_PATH);
     wcscpy_s(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder);
-    PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
 
+    // For Windows 7 support, we cannot use PathCch functions.
+    //PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
+    WCHAR wszBuffer[2 * MAX_PATH]; // set to 2x the MAX_PATH in order to prevent buffer overflow.
+    if (!PathCanonicalize(wszBuffer, wszModuleFullPath) ||
```
delmyers, line 90: >wszModuleFullPath [](http://example.com/codeflow?start=37&length=17)

can you just use m_bstrModuleFolder and avoid the string copy into wszModuleFullPath above?

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -83,8 +83,20 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
     WCHAR wszModuleFullPath[MAX_PATH];
     memset(wszModuleFullPath, 0, MAX_PATH);
     wcscpy_s(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder);
-    PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
 
+    // For Windows 7 support, we cannot use PathCch functions.
+    //PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModule);
+    WCHAR wszBuffer[2 * MAX_PATH]; // set to 2x the MAX_PATH in order to prevent buffer overflow.
```
delmyers, line 89: >// set to 2x the MAX_PATH in order to prevent buffer overflow [](http://example.com/codeflow?start=35&length=61)

Have we guaranteed that m_bstrModuleFolder.Lenghth() >= MAX_PATH elsewhere?

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
delmyers: Ah, I see. It is because of the wcslen(wszBuffer) check.

I would suggest checking that separately, and returning E_BOUNDS.

Also, I don't think we are supposed to be using wcslen. Can you use StringUtils::WStringLen instead?

src/InstrumentationEngine.ProfilerProxy/dllmain.cpp
```diff
@@ -108,12 +109,26 @@ namespace ProfilerProxy
...
+        DWORD dError = 0;
 
         WCHAR wszEngineFullPath[MAX_PATH];
         IfFailRetErrno_Proxy(wcscpy_s(wszEngineFullPath, MAX_PATH, wszProfilerPath));
 
-        IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, wszVersionFolder));
-        IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, profilerRelativeFileName));
+        // For Windows 7 support, we cannot use PathCch functions.
+        //IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, wszVersionFolder));
+        //IfFailRet_Proxy(PathCchAppend(wszEngineFullPath, MAX_PATH, profilerRelativeFileName));
```
delmyers, line 119: Remove comment.

src/InstrumentationEngine.ProfilerProxy/dllmain.cpp
```diff
@@ -108,12 +109,26 @@ namespace ProfilerProxy
...
+        WCHAR wszBuffer[2 * MAX_PATH]; // set to 2x the MAX_PATH in order to prevent buffer overflow.
+        if (!PathCanonicalize(wszBuffer, wszEngineFullPath) ||
+            !PathAppend(wszBuffer, wszVersionFolder) ||
+            !PathAppend(wszBuffer, profilerRelativeFileName) ||
+            wcslen(wszBuffer) >= MAX_PATH)
+        {
+            dError = GetLastError();
+            eventLogger.LogError(_T("dllmain::HasProfilerDll - Unable to generate fullpath to CLRIE dll: '%s' + '%s' + '%s'"), wszEngineFullPath, wszVersionFolder, profilerRelativeFileName);
+            return dError == 0 ? E_FAIL : HRESULT_FROM_WIN32(dError);
+        }
```
delmyers, line 129: Looks like duplicated logic. Can it be pulled into Common.Lib?

tests/InstrEngineTests/NaglerInstrumentationMethod/stdafx.h
```diff
@@ -36,7 +36,8 @@ do { if (FAILED(hr = (EXPR))) { ATLASSERT(!L"IfFailRet(" L#EXPR L") failed"); re
 #include <vector>
 #include <memory>
 #include <unordered_map>
-#include <Pathcch.h>
+#include <Shlwapi.h>
+//#include <Pathcch.h> // For Windows 7 support, we cannot use PathCch functions.
```
delmyers, line 40: No need to bother with this comment, I don't think. It probably won't prevent any mistakes anyway.

delmyers (changes requested):
:clock1:

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
delmyers: Ah, I see. It is because of the wcslen(wszBuffer) check.

I would suggest checking that separately, and returning E_BOUNDS.

Also, I don't think we are supposed to be using wcslen. Can you use StringUtils::WStringLen instead?

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
WilliamXieMSFT: Good point, actually I'll just use wszModuleFullPath

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
WilliamXieMSFT: PathCanonicalize will report outofbounds error

src/InstrumentationEngine.ProfilerProxy/dllmain.cpp
WilliamXieMSFT: refactoring made these functions slightly different

src/Common.Lib/PathUtils.cpp
```diff
@@ -36,4 +37,20 @@ namespace CommonLib
...
+
+    HRESULT PathUtils::SafePathAppend(LPWSTR pwszPath, LPCWSTR pwszMore, size_t cBounds)
+    {
+        // If PathAppend fails, the destination buffer will be cleared. Check bounds before appending.
+        if (StringUtils::WStringLen(pwszPath) + StringUtils::WStringLen(pwszMore) >= cBounds)
+        {
+            return E_BOUNDS;
+        }
+
+        if (!PathAppend(pwszPath, pwszMore))
```
WilliamXieMSFT, line 49: >PathAppend [](http://example.com/codeflow?start=13&length=10)

Not sure if this will work on Linux, we may need to update the pal

src/InstrumentationEngine.Lib/StringUtils.h
```diff
@@ -5,6 +5,11 @@
 
 #include <string>
 
+#ifndef E_BOUNDS
+// Apparantly, Linux builds don't have E_BOUNDS defined.
+#define E_BOUNDS 0x8000000BL
+#endif
```
delmyers, line 11: This is in unix.h now.

src/InstrumentationEngine.Lib/StringUtils.h
```diff
@@ -31,4 +36,10 @@ class StringUtils
     // Yields the same result as calling strnlen(szString, MAX_STRING_LEN);
     // For use with untrusted data which may not have a null terminator.
     static size_t StringLen(_In_ const char* szString);
+
+    // Modifies pwszPath by appending pwszMore.
+    // Checks cBounds before appending to prevent buffer overflows.
+    //
+    // Consider moving to CommonLib & make PathAppend available in the PAL.
+    static HRESULT SafePathAppend(_Inout_ LPWSTR pwszPath, _In_ LPCWSTR pwszMore, _In_ size_t cBounds);
```
delmyers, line 44: >_Inout_ [](http://example.com/codeflow?start=34&length=7)

`_Inout_updates_z_(cBounds)` ?

src/InstrumentationEngine.Lib/StringUtils.h
WilliamXieMSFT: It is, however this is going directly to the release branch which doesn't have the change (yet)

src/InstrumentationEngine.Lib/StringUtils.h
```diff
@@ -31,4 +36,10 @@ class StringUtils
     // Yields the same result as calling strnlen(szString, MAX_STRING_LEN);
     // For use with untrusted data which may not have a null terminator.
     static size_t StringLen(_In_ const char* szString);
+
+    // Modifies pwszPath by appending pwszMore.
+    // Checks cBounds before appending to prevent buffer overflows.
+    //
+    // Consider moving to CommonLib & make PathAppend available in the PAL.
```
delmyers, line 43: > PathAppend available in the PAL [](http://example.com/codeflow?start=42&length=32)

We want to keep CommLib free of the PAL so that we can reuse it in Nagler (eventually).

If we could update to using C++17, then we could just use std::filesystem. I don't think we can do that because we are limited to the versions available for the platforms that we support... I'm unsure on that, though.

src/InstrumentationEngine.Lib/StringUtils.h
WilliamXieMSFT: I'll move it to unix.h so it matches main

src/InstrumentationEngine.Lib/StringUtils.h
WilliamXieMSFT: Gotcha, I'll update the comment

jander-msft: > If we could update to using C++17, then we could just use std::filesystem. I don't think we can do that because we are limited to the versions available for the platforms that we support... I'm unsure on that, though. #Pending

Correct. The PAL limits us to certain clang versions, which then limits us to having most of C++14 and very little of C++17 available for use on Linux.

src/InstrumentationEngine.ProfilerProxy/InstrumentationEngine.ProfilerProxy.vcxproj
```diff
@@ -65,7 +65,7 @@
   </ImportGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <AdditionalIncludeDirectories>$(OutDir);..\Common.Lib;..\InstrumentationEngine.ProfilerProxy.Lib;..\..\inc;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(OutDir);..\Common.Lib;..\InstrumentationEngine.ProfilerProxy.Lib;..\..\inc;..\InstrumentationEngine.Lib;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
delmyers, line 68: >..\InstrumentationEngine.Lib [](http://example.com/codeflow?start=113&length=28)

Ugh... linking InstrumentationEngine.Lib into this dll isn't great. There is a lot of code in there that we don't need in this one. On Windows, I think that will be fine. The linker should strip out unused code for retail builds. I don't know what it will do on Linux, though.

delmyers: We could consider making StringUtils header-only (move all of the implementation into the .h file). That way, we wouldn't need to add this additional linkage.

WilliamXieMSFT: I can create a GitHub issue for that; currently the ProfilerProxy is only used in the MSI scenario but I can see it going xpat.

WilliamXieMSFT: [https://github.com/microsoft/CLRInstrumentationEngine/issues/417](https://github.com/microsoft/CLRInstrumentationEngine/issues/417)

delmyers: OK. I'll sign off on this for now, so that we can get it in and unblock our partners. I would like to see #417 fixed ASAP, though.

delmyers (approved):
:shipit:

PR #407, opened 2021-07-06
Fix roslyn static analysis

delmyers:
We were accidentally building some C++ unit tests
during the managed static analysis builds. This
was confusing the static analysis and causing it
to fail.


WilliamXieMSFT (approved):
:shipit:

PR #408, opened 2021-07-08
Enable CodeQL (Semmle) analysis for C++

delmyers:
CodeQL is a static analysis tool that finds common antipatterns in code. We are enabling it for C++.
This change:

1) Enables the newest version of CodeQL (previously called Semmle)
    in the build pipeline.
2) Fixes any issues that CodeQL found, or suppress false positives
3) Updates the code analysis stages of the build pipeline, so that they
    can be manually disabled for testing purposes.


src/InstrumentationEngine/MethodInfo.cpp
```diff
@@ -372,7 +370,7 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::CreateILFunctionBody()
 {
     HRESULT hr = S_OK;
 
-    if (m_pIntermediateRenderedMethod != nullptr)
+    if (m_pIntermediateRenderedMethod.empty())
```
delmyers, line 373: >m_pIntermediateRenderedMethod.empty() [](http://example.com/codeflow?start=8&length=37)

typo. This condition should be opposite.

build/yaml/pipelines/codeanalysis.yaml
```diff
@@ -64,18 +64,26 @@ stages:
...
-#    parameters:
-#      Platform: x86
-#      Analyzer: cpp
+  - template: ../jobs/codeanalysis/buildanalyzer.yaml
+    parameters:
+      Platform: x86
+      Analyzer: cpp
 
+- stage: Run_PreFast_Analyzers
+  displayName: Code Analysis (Build Time C++ PreFast )
```
jander-msft, line 83: nit:
```suggestion
  displayName: Code Analysis (Build Time C++ PreFast)
```

src/InstrumentationEngine.Lib/FileLoggerSink.cpp
```diff
@@ -183,7 +185,10 @@ void CFileLoggerSink::WritePrefix(_In_ LoggingFlags flags)
 #else
                 WCHAR szFormatted[MAX_PATH];
                 wcsftime(szFormatted, MAX_PATH, szFormat, &localTime);
-                fwprintf(pOutputFile, szFormatted);
+
+                // Static analysis will complain that the following line prints a non-constant
+                // string, but there are no inputs to the string. Ignore.
+                fwprintf(pOutputFile, szFormatted);  // lgtm[cpp/non-constant-format]
```
jander-msft, line 191: Is this the same error that you fixed in https://github.com/microsoft/CLRInstrumentationEngine/pull/408/files#diff-57c5f94f29184e8ca0b109256bc0adbb292ef774a8ed031a882753cc2067ce54

Why fix it there but suppress it here?

src/InstrumentationEngine/MethodInfo.cpp
```diff
@@ -1132,13 +1129,15 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::InitializeName(_In_ mdToken
 
         nameBuilder << (LPWSTR)m_bstrMethodName << _T("<");
 
-        for (ULONG i = 0; i < m_genericParameters.size(); i++)
+        ULONG cparams = static_cast<ULONG>(m_genericParameters.size());
+        IfFalseRet(cparams == m_genericParameters.size(), E_BOUNDS);
```
jander-msft, line 1133: This doesn't cause a static analysis error? Because it's checking equality of a `ULONG` vs `SIZE_T`.

src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -721,8 +721,9 @@ HRESULT CProfilerManager::LoadInstrumentationMethods(_In_ CConfigurationSource*
 }
 
 HRESULT CProfilerManager::RemoveInstrumentationMethod(
-    _In_ IInstrumentationMethod* pInstrumentationMethod)
+    _In_opt_ IInstrumentationMethod* pInstrumentationMethod)
```
jander-msft, line 724: Why is this `_In_opt_` but then we fail the function if it's `nullptr`? Is this because the IDL file and generated headers state its `_In_opt_`?

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentMethodEntry.h
```diff
@@ -160,13 +160,15 @@ class CInstrumentMethodEntry
 
     HRESULT GetModuleName(BSTR* pbstrModuleName)
     {
-        m_bstrModuleName.CopyTo(pbstrModuleName);
+        HRESULT hr;
```
jander-msft, line 163: Should these be:
```suggestion
        HRESULT hr = S_OK;
```

jander-msft (approved):

src/InstrumentationEngine.Lib/FileLoggerSink.cpp
delmyers: I was going to think about this, and potentially reconsider it.

For many of these, fixed them just in case they end up with format characters like '%' in the string. I didn't think that was possible here.  It's probably better to be safe than sorry.

src/InstrumentationEngine/MethodInfo.cpp
delmyers: This is an overflow check. The compiler should cast cparams to a size_t in order to do the comparison, so there shouldn't be a problem.

Truth be told, it isn't probably necessary at all. I can't think of a situation in which we would have more than 4M generic parameters. I was probably just being overcautious because the loop below was being flagged by static analysis (if the parameter count overflows, the loop will loop forever because max(ULONG) < max(size_t)). We would probably cause an out of memory error first.

src/InstrumentationEngine/ProfilerManager.cpp
delmyers: idl always generates _In_opt_. Potentially, it can't guarantee that the caller will not pass in a null pointer, so everything is marked as _In_opt_?

We are free to declare whatever contract we want, though, so it is fine to return an error.

tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentMethodEntry.h
delmyers: Not necessary. The hresult will be set immediately on the next line.

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -388,27 +388,33 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::EncodeIL(
...
+
+        cbBuffer = cbBufferTemp;
         pInstruction = pInstruction->NextInstructionInternal();
     }
 
     IfFailRet(CalculateInstructionOffsets());
 
     // Allocate the il buffer
-    CAutoVectorPtr<BYTE> pILArray;
+    vector<BYTE> pILArray;
```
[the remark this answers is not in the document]
delmyers: I'll leave this as-is, but there are some more simplifications I can do here.

PR #409, opened 2021-07-08
Convert to SDK style projects and add .NET 6 to InstrEngineTests

jander-msft:
Overview of changes:
- Update all managed projects to .NET SDK project format.
- Update pipelines to install .NET 6 runtimes and SDKs.
- Add .NET 6 to InstrEngineTests project.
- Use dotnet restore for managed build and test jobs.
- Disable some tests due to infrastructure assumptions.
- Regenerate files from T4 templates to ensure no differences.

The goal here is to add .NET 6 to the InstrEngineTests projects to get some test coverage on .NET 6. Since the builds are running on release versions of MSBuild, the msbuild tool cannot be used for building the managed projects that contain the .NET 6 target framework. This precludes the use of msbuild.exe and nuget.exe on the InstrumentationEngine.sln file in the pipelines for the `Any CPU` platform; the `x64` and `x86` platforms remain unaffected because they do not build for .NET 6.

The changes use `dotnet restore` for restoring NuGet packages, `dotnet build` to build the projects, and `dotnet test` to run tests, for the `Any CPU` platform. This requires that all of the managed projects are converted to .NET SDK style projects. To be able to restore the NuGet packages for the `x64` and `x86` platforms, a new solution file was created on which `nuget restore` will be issued. This is a short-term stopgap; the managed projects in the `x64` and `x86` platforms should be moved to the `Any CPU` platform in the long term, however that would have made this change significantly larger. This change to remove the new solution and move the managed projects can be made in a subsequent change.

These changes require the installation of .NET 6 SDK on build machines as well as the use of a Visual Studio install that allows the use of .NET 6 (e.g. Visual Studio 2019 Preview Channel). I can add some more code to the MSBuild projects to allow conditional opt-in or opt-out of .NET 6 builds (but always have it enabled on the CI machines) if that is desirable.


build/yaml/steps/codeanalysis/csharp.yaml
```diff
@@ -9,7 +9,7 @@ steps:
   inputs:
     continueOnError: true
     rulesetName: Custom
-    msBuildArchitecture: x86
+    msBuildArchitecture: x64
```
WilliamXieMSFT, line 12: >x64 [](http://example.com/codeflow?start=25&length=3)

I'm assuming the reason for changing this is performance?

build/yaml/steps/windows/dotnettest.yaml
```diff
@@ -0,0 +1,24 @@
...
+parameters:
+  SubFolder: ''
+  RunTitle: ''
+
+steps:
+- task: DotNetCoreCLI@2
+  displayName: 'Run Tests: $(Platform)\${{ parameters.SubFolder }}*Tests*.dll'
+  inputs:
+    command: test
+    arguments: >-
```
WilliamXieMSFT, line 15: >>- [](http://example.com/codeflow?start=15&length=2)

Is this yaml syntax or an argument for dotnet test?

WilliamXieMSFT:
> Microsoft Visual Studio Solution File, Format Version 12.00

I just realized these sln files don't have copyright headers (my bad), would you mind adding them?

---
Refers to: InstrumentationEngine.sln:2 in 6057792. [](commit_id = 6057792b58baf1357e883b5309df4740d66ee677, deletion_comment = False)

WilliamXieMSFT (approved):
:shipit:

build/yaml/steps/windows/dotnettest.yaml
jander-msft: This yaml syntax says that each item in the following section should be interpreted as a concatenation of each together using a single space between each element to produce a single string. It allows you to break apart really long strings so it's more readable/diff-able.

build/yaml/steps/codeanalysis/csharp.yaml
jander-msft: It might not be necessary to change it, but the installation of dotnet from the UseDotNet tasks with install for the native architecture of the OS, which is x64 for the build machines. I ran the static analysis build and it seemed to work fine.

jander-msft:
> # Copyright (c) Microsoft Corporation. All rights reserved.
> # Licensed under the MIT License.

Visual Studio refuses to load the solution if I try adding the copyright message.

build/yaml/steps/windows/dotnettest.yaml
WilliamXieMSFT: awesome, learned something new!

build/yaml/steps/windows/enginerestore.yaml
```diff
@@ -0,0 +1,34 @@
...
+# Long term, the managed projects in the 'x64' and 'x86' platforms should be
+# moved to the 'Any CPU' platform, which will allow usage of 'dotnet restore'
+# for all platforms. 
+
+steps:
+# nuget restore InstrumentationEngine.sln
+- template: dotnetrestore.yaml
+  parameters:
+    Solution: InstrumentationEngine.sln
+    Condition: and(succeeded(), eq(variables['Platform'], 'Any CPU'))
```
delmyers, line 29: >eq(variables['Platform'], 'Any CPU') [](http://example.com/codeflow?start=32&length=36)

Will dotnet restore not restore all projects? Just out of curiosity.

jander-msft: Seems that `dotnet restore` only restores projects in the `Any CPU` platform.

delmyers:
> // Copyright (c) Microsoft Corporation. All rights reserved.

Why are you removing this file?

---
Refers to: src/Extensions.Base.Api/NativeMethods.cs:1 in 6057792. [](commit_id = 6057792b58baf1357e883b5309df4740d66ee677, deletion_comment = True)

jander-msft:
> > // Copyright (c) Microsoft Corporation. All rights reserved.
> 
> Why are you removing this file?
> 
> Refers to: src/Extensions.Base.Api/NativeMethods.cs:1 in [6057792](https://github.com/microsoft/CLRInstrumentationEngine/commit/6057792b58baf1357e883b5309df4740d66ee677). [](commit_id = [6057792](https://github.com/microsoft/CLRInstrumentationEngine/commit/6057792b58baf1357e883b5309df4740d66ee677), deletion_comment = True)

It was not actually included in the project prior to my change, so it never compiled. I checked the binaries from previous builds to be sure. Since moving to the SDK style projects, all source files are included by default. Instead of explicitly excluding it, I'm just deleting it.

tests/RawProfilerHook/RawProfilerHook.Tests/TestsRawProfilerHookAttach.cs
```diff
@@ -61,7 +63,9 @@ public override void Execute()
             }
         }
 
-        [TestMethod]
+        // TODO: RemoteUnitTestExecutor does not run properly due to strong name signing verification
```
delmyers, line 66: > TODO: [](http://example.com/codeflow?start=10&length=6)

Did you add a work item for this?

delmyers (approved):
:shipit:

tests/RawProfilerHook/RawProfilerHook.Tests/TestsRawProfilerHookAttach.cs
jander-msft: I have one internally; I can create a GH issue too for tracking, although I'm working on it now :-)

Issue #410, opened 2021-07-09
Announcement: develop and main branch renames


jander-msft:
In order to be more consistent with typical branch naming the following branches will be renamed:
- `main` -> `release`
- `develop` -> `main`

The `main` branch will become the default branch.

This transition will be done by:
- Lock the `develop` branch from accepting any changes.
- Update the `main` branch yaml pipelines to trigger on new branch names.
- Create a new `release` branch from the current `main` branch.
- Update Azure DevOps pipelines and release definitions and GitHub policies with new branch names.
- Merge the `develop` branch into the `main` branch.
- Delete the `develop` branch after verifying all builds and triggers are correct.

This transition is planned to start on Thursday, July 15th and to be completed on Friday, July 16th.

jander-msft:
This work is completed and the default branch has been changed to `main`. Releases are now performed out of the `release` branch.

PR #415, opened 2021-07-14
Rename branches: main to release, develop to main

jander-msft:
These changes rename the branches as described in #410. This change targets the current main branch so that when the new release branch is forked from it, the changes are already in the release branch.

### DIFF
--- a/build/yaml/pipelines/codeanalysis.yaml
+++ b/build/yaml/pipelines/codeanalysis.yaml
@@ -15,7 +15,7 @@ schedules:
   displayName: Weekly Static Analysis
   branches:
     include:
-    - main
+    - release
 
 variables:
   TeamName: ClrInstrumentationEngine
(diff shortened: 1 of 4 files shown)


build/yaml/pipelines/codeanalysis.yaml
```diff
@@ -15,7 +15,7 @@ schedules:
   displayName: Weekly Static Analysis
   branches:
     include:
-    - main
+    - release
```
delmyers, line 18: We can probably get rid of this. We run static analysis as a CI after each completed pull request now.

delmyers (approved):
:shipit:

build/yaml/pipelines/codeanalysis.yaml
jander-msft: Removed

PR #423, opened 2021-07-21
Enable .NET Analyzers for all managed projects.

jander-msft:
Overview:
- Enable .NET Analyzers for all managed projects.
- Use SDL Recommended ruleset at Error level.
- Report all code analysis warnings as errors.
- Fix existing violations.

Since .NET code analysis is relatively lightweight, it is run unconditionally on all managed projects using the SDL recommended rulset; any violation will cause a build error and prevent PR builds from succeeding.


build/Managed.props
```diff
@@ -20,6 +20,9 @@
 
   <PropertyGroup>
     <DefineConstants Condition="'$(PublicRelease)' != 'True'">$(DefineConstants);ALLOWNOTSIGNED</DefineConstants>
+    <EnableNetAnalyzers>true</EnableNetAnalyzers>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysis\Sdl.Recommended.Error.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisTreatWarningsAsErrors>false</CodeAnalysisTreatWarningsAsErrors>
```
jander-msft, line 25: This should be `true`

build/yaml/steps/codeanalysis/csharp.yaml
```diff
@@ -10,6 +10,6 @@ steps:
     continueOnError: true
     rulesetName: Custom
     msBuildArchitecture: x64
-    customRuleset: DevDivRoslynRequired.ruleset
+    customRuleset: $(Build.SourcesDirectory)\build\CodeAnalysis\Sdl.Recommended.Error.ruleset
```
delmyers, line 13: > $(Build.SourcesDirectory)\build\CodeAnalysis\Sdl.Recommended.Error.ruleset [](http://example.com/codeflow?start=18&length=75)

Are we sure that these two rulesets are the same?

jander-msft: It is not the same; the prior behavior was referring to a non-existent rule set. There is no DevDivRoslynRequired.ruleset in the repo.

Compared to the hypothetical DevDivRoslynRequired, the new ruleset should be more restrictive (more enabled rules, and all rules at the same or more restrictive level).

delmyers: I'm not certain that the ruleset has to be in the repository. It might be included with the guardian toolset that gets downloaded for the task. I would have to look at the documentation for the task again, though.

jander-msft: The documentation says that the customRuleset should be "An accessible path to a ruleset to use." Using a built-in ruleset is only supported by specifying the resultName input. However, since we want the build analyzers and this task to agree on the ruleset, we should be using the path to the custom ruleset.

src/Tests/RemoteUnitTestExecutor/TestResult.cs
```diff
@@ -36,11 +38,18 @@ public void Serialize(string filename)
             }
         }
 
+        /// CA2300 is set a Info level, which will not cause a build break. However, RoslynAnalyzer and PostAnalysis
```
WilliamXieMSFT, line 41: >a [](http://example.com/codeflow?start=26&length=1)

super nit `at`

WilliamXieMSFT (approved):
:shipit:

build/yaml/steps/codeanalysis/csharp.yaml
jander-msft: I dug into this more this morning to figure out what was happening.

Prior to this change, there were a few code analysis warnings that the RoslynAnalyzers task would flag, for example, CA1062. If you check the actual DevDivRoslynRequired.ruleset or the Sdl.Recommended.Error.ruleset, this rule is set to None in both cases. Also, if you check the RoslynAnalyzers build, there are a bunch of "Could not find rule set file C:\a\1\s\DevDivRoslynRequired.ruleset" warnings. This leads me to believe that the RoslynAnalyzers task was just running all possible rules.

I wrote a quick tool to compare the Sdl.Recommended.Error.ruleset and DevDivRoslynRequired.ruleset files and found that the SDL one is a superset of the DevDiv one, with the exception of a handful of internal rules (they start with IA). Because they are internal, they require a special NuGet package to enable those rules, which I don't think we want to put up on the dependency feed. So I think we need to use the SDL ruleset.

I don't see any precedence for adding the DevDivRoslynRequired.rulset to any of the GitHub repos in the microsoft or dotnet orgs.

build/yaml/steps/codeanalysis/csharp.yaml
```diff
@@ -9,7 +9,7 @@ steps:
   inputs:
     continueOnError: true
     rulesetName: Custom
-    msBuildArchitecture: x64
-    customRuleset: DevDivRoslynRequired.ruleset
+    msBuildArchitecture: amd64
+    customRuleset: $(Build.SourcesDirectory)\build\CodeAnalysis\Sdl.Recommended.Error.ruleset
```
jander-msft, line 13: I will revert this back what it was before (even though it is broken) and then fix whatever warnings/errors come out of the RoslynAnalyzers task.

I will leave the changes to use the SDL ruleset during the build phase in order to aid in resolving most of the overlapping rules from the SDL ruleset and the one that this task should be using.

In a later change, I'll figure out how to run the appropriate ruleset with this task.

WilliamXieMSFT (approved):
:shipit:

PR #424, opened 2021-07-21
Fix signed builds

WilliamXieMSFT:


jander-msft:
What is the issue that is being fixed?

WilliamXieMSFT:
The error is due to the API rate limited by GitHub when downloading the CredentialProvider to install the Signing plugin.
Update the task from v2 to v3 solves this problem.

jander-msft (approved):

PR #427, opened 2021-07-23
Revert to using PathCchAppend for Linux & Version bump to 1.0.42

WilliamXieMSFT:


src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -89,15 +89,22 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
 
     WCHAR wszModuleFullPath[MAX_PATH];
     memset(wszModuleFullPath, 0, MAX_PATH);
+#ifndef PLATFORM_UNIX
     if (!PathCanonicalize(wszModuleFullPath, m_bstrModuleFolder))
+#else
+    if (FAILED(PathCchAppend(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder)))
```
jander-msft, line 95: Shouldn't this be PathCchCanonicalize, so long as that is implemented in the PAL?

WilliamXieMSFT: Unfortunately that's not available in the PAL so I just reverted to what the code was before.

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -89,15 +89,22 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
 
     WCHAR wszModuleFullPath[MAX_PATH];
     memset(wszModuleFullPath, 0, MAX_PATH);
```
jander-msft, line 91: I think this needs to be wmemset; otherwise it will only zero out the first 260 bytes instead of 520 bytes: sizeof(WCHAR) == 2 * sizeof(CHAR)

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
jander-msft: That's not quite correct. Prior to the changes in #404, it initialized wszModuleFullPath with:

```cpp
wcscpy_s(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder);
```

jander-msft: I'm not certain why PathCanonicalize was introduced here in that PR.

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
WilliamXieMSFT: Spoke offline, the PAL doesn't support wmemset, will log an issue to track this.

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
WilliamXieMSFT: True, I can just revert the entire behavior; it's not really needed

jander-msft (approved):

src/InstrumentationEngine.Lib/InstrumentationMethod.cpp
```diff
@@ -88,13 +88,8 @@ HRESULT MicrosoftInstrumentationEngine::CInstrumentationMethod::InitializeCore(
...
-    if (!PathCanonicalize(wszModuleFullPath, m_bstrModuleFolder))
-    {
-        DWORD dwLastError = GetLastError();
-        CLogging::LogError(_T("CInstrumentationMethod::Initialize - unable to canonicalize method configuration folder: '%s', PID: %u"), m_bstrModuleFolder.m_str, GetCurrentProcessId());
-        return HRESULT_FROM_WIN32(dwLastError);
-    }
+    ZeroMemory(wszModuleFullPath, MAX_PATH * sizeof(WCHAR));
+    wcscpy_s(wszModuleFullPath, MAX_PATH, m_bstrModuleFolder);
 
     hr = StringUtils::SafePathAppend(wszModuleFullPath, m_bstrModule, MAX_PATH);
```
delmyers, line 94: >wszModuleFullPath [](http://example.com/codeflow?start=37&length=17)

Without the canonicalization, we run a greater risk of running out of space in wszModuleFuillPath. If m_bstrModuleFolder isn't already canonicalized, then you can have paths like"

"C:\somePath\someSubPath\..\adifferentSubapth\someotherPathHere\,,\,,\realSubPath\\\"

When what you will want is:

"C:\somePath\realSubPath"

If only we could use c++17

delmyers: Note: on Windows, you can use PathCanonicalize, and on Linux, you can use "realpath"

delmyers: You would have to convert back-and-forth from Windows strings, though.

jander-msft: I would suggest we fix the canonicalization problem later. The behavior in the PR at this time is the behavior before the Win7 support change and we didn't run into the problems in prior versions.

While I agree that in general this can be a problem, we haven't run into it in practice and I don't think any customer has reported it.

I think we should get the Linux support restored (with the change as-is) and later make more of an effort to fix these types of scenarios.

delmyers: Agreed

delmyers:
:shipit:

delmyers (approved):

src/Common.Headers/StringUtils.h
```diff
@@ -77,5 +79,9 @@ class StringUtils
         }
 
         return S_OK;
+#else
+        LPWSTR pwszMore2 = const_cast<LPWSTR>(pwszMore);
+        return PathCchAppend(pwszPath, MAX_PATH, pwszMore2);
```
jander-msft, line 84: Isn't this defined as:

```
HRESULT PathCchAppend(
  PWSTR  pszPath,
  size_t cchPath,
  PCWSTR pszMore
);
```

so LPCWSTR should work for pszMore. Also, I think a lot of the code before the Win7 support change passed in all kinds of variations of `const WCHAR*`.

WilliamXieMSFT: It's defined in the palrt2.h as:

```
extern "C" HRESULT PathCchAppend(
    _Inout_  WCHAR*  wszPath,
    _In_     size_t cchPath,
    _In_opt_ WCHAR* wszMore
);
```

Should I update this instead?

jander-msft: We should check what the real signature was when the pal was built from dotnet/coreclr.

jander-msft (changes requested):
Blocking PR until PathCchAppend signature can be validated.

src/Common.Headers/StringUtils.h
jander-msft: The declaration is here: https://github.com/microsoft/CLRInstrumentationEngine/blob/43c4e97b8dbde7d8b2ca1a5e94c40ba8a53f5a22/src/unix/inc/palrt2.h#L646
The implementation is here: https://github.com/microsoft/CLRInstrumentationEngine/blob/43c4e97b8dbde7d8b2ca1a5e94c40ba8a53f5a22/src/unix/src/misc/file.cpp#L73

So this function is completely within out control.

I would say revert the header change that you made and just try to compile without having to do the const_cast. It was working before the Win 7 support change and it should work still (I don't see anything obvious that suggests it shouldn't work given how it worked before the Win 7 support change).

If it doesn't compile, then change both the declaration and implementation to match the signature that was defined for Windows:
```
HRESULT
PathCchAppend(
    _Inout_updates_(cchPath) PWSTR pszPath,
    _In_ size_t cchPath,
    _In_opt_ PCWSTR pszMore
    );
```

so that it same for Windows and non-Windows.

jakubch1 (approved):
Works good on ubuntu

jander-msft (approved):
:shipit:

Issue #60, opened 2019-04-30
Generalize Linux ExtensionHost


WilliamXieMSFT:
Current implementation of the Linux ExtensionsHost (which deals with loading Instrumentation Methods) specifically loads Snapshot Debugger's method "Production Breakpoints". This needs to be generalized to have functional parity with its Windows counterpart.

PR #428, opened 2021-08-03
Build fixes for OSX and recent Linux

dividedmind:
Hi, for https://github.com/applandinc/appmap-dotnet I needed CLRIE builds for Linux and OSX and found it a bit difficult to obtain them. I've discovered a couple of small issues (including a bug) that prevented those builds from proceeding on recent tools; each fix is in a separate commit for review convenience and the commit messages should make it pretty clear what the changes are. There's also a commit enabling build on OSX to proceed and, as a bonus, a github actions build job.

For the record, I am the sole author of these changes and I consider them trivial enough to not be copyrightable, and in any case, I disclaim any eventual copyrights for them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,42 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+name: build
+
+on: [push]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        # Note OSX isn't officially supported yet, but might be in the future.
+        # The build here will help ensure no changes are introduced that
+        # would make it difficult.
+        os: [ ubuntu-latest, macos-latest ]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/cache@v2
+      with:
+        path: out/packages
+        key: nuget
+
+    - run: src/build.sh x64 release
+
+    - name: archive so
+      if: matrix.os == 'ubuntu-latest'
+      uses: actions/upload-artifact@v2
+      with:
+        name: libInstrumentationEngine.so
+        path: out/Linux/bin/x64.Release/ClrInstrumentationEngine/libInstrumentationEngine.so
+
+    - name: archive dylib
+      if: matrix.os == 'macos-latest'
+      uses: actions/upload-artifact@v2
+      with:
+        name: libInstrumentationEngine.dylib
+        path: out/OSX/bin/x64.Release/ClrInstrumentationEngine/libInstrumentationEngine.dylib
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -153,6 +153,8 @@ macro (build_init build_language build_type)
     add_compile_options(-Wno-new-returns-null)
     add_compile_options(-Wno-deprecated-declarations)
     add_compile_options(-Wno-delete-incomplete)
+    add_compile_options(-Wno-pragma-pack)
+    add_compile_options(-Wno-unknown-warning-option)
 
     # Turn all enabled warnings into errors
     add_compile_options(-Werror)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -274,8 +274,8 @@ endif(CMAKE_SYSTEM_NAME STREQUAL Linux)
 
 
 if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
-    message(FATAL_ERROR "Darwin build not supported.")
-    return()
+    set(PLAT_DLL_EXT "dylib")
+    set(CLR_CMAKE_PLATFORM_UNIX 1)
 elseif(CMAKE_SYSTEM_NAME STREQUAL Linux)
     set(PLAT_DLL_EXT "so")
 else()
@@ -322,33 +322,39 @@ if (CMAKE_SYSTEM_NAME STREQUAL Linux)
     )
 
     add_custom_target(pal_redefines DEPENDS ${REDEFINES_FILE})
+elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+    message(NOTICE "Building for Darwin is currently experimental.")
 
-    function(add_library_pal target_name do_redefinition lib_location)
-        get_filename_component(lib_file_name ${lib_location} NAME)
-        set(redefine_lib_location ${CMAKE_CURRENT_BINARY_DIR}/${lib_file_name})
-
-        if(${do_redefinition})
-            # Copy the PAL lib from it's original location and do the redefine
-            add_custom_command(
-                OUTPUT ${redefine_lib_location}
-                COMMAND ${OBJCOPY} --redefine-syms ${REDEFINES_FILE} ${lib_location} ${redefine_lib_location}
-                DEPENDS pal_redefines ${lib_location} ${REDEFINES_FILE}
-                COMMENT "Redefining exported PAL symbols in ${lib_location}"
-            )
-        else()
-            configure_file(${lib_location} ${redefine_lib_location} COPYONLY)
-        endif(${do_redefinition})
-
-        add_custom_target(${target_name}_redefines DEPENDS ${redefine_lib_location})
-
-        add_library(${target_name} STATIC IMPORTED)
-        set_property(TARGET ${target_name} PROPERTY IMPORTED_LOCATION ${redefine_lib_location})
-
-        add_dependencies(${target_name} ${target_name}_redefines)
-    endfunction()
+    # Dummy noop target to prevent warnings.
+    add_custom_target(pal_redefines_header)
 else()
-    message(FATAL_ERROR "Linux is only supported platform.")
+    message(FATAL_ERROR "CMake build not supported for system '${CMAKE_SYSTEM_NAME}'")
 endif()
+
+function(add_library_pal target_name do_redefinition lib_location)
+    get_filename_component(lib_file_name ${lib_location} NAME)
+    set(redefine_lib_location ${CMAKE_CURRENT_BINARY_DIR}/${lib_file_name})
+
+    if((CMAKE_SYSTEM_NAME STREQUAL Linux) AND ${do_redefinition})
+        # Copy the PAL lib from it's original location and do the redefine
+        add_custom_command(
+            OUTPUT ${redefine_lib_location}
+            COMMAND ${OBJCOPY} --redefine-syms ${REDEFINES_FILE} ${lib_location} ${redefine_lib_location}
+            DEPENDS pal_redefines ${lib_location} ${REDEFINES_FILE}
+            COMMENT "Redefining exported PAL symbols in ${lib_location}"
+        )
+    else()
+        configure_file(${lib_location} ${redefine_lib_location} COPYONLY)
+    endif()
+
+    add_custom_target(${target_name}_redefines DEPENDS ${redefine_lib_location})
+
+    add_library(${target_name} STATIC IMPORTED)
+    set_property(TARGET ${target_name} PROPERTY IMPORTED_LOCATION ${redefine_lib_location})
+
+    add_dependencies(${target_name} ${target_name}_redefines)
+endfunction()
+
 add_library_pal(coreclrpal true ${CORECLR_PAL_ROOT}/lib/${CMAKE_BUILD_TYPE}/libcoreclrpal.a)
 add_library_pal(palrt true ${CORECLR_PAL_ROOT}/lib/${CMAKE_BUILD_TYPE}/libpalrt.a)
 add_library_pal(corguids false ${CORECLR_PAL_ROOT}/lib/${CMAKE_BUILD_TYPE}/libcorguids.a)
@@ -363,4 +369,3 @@ add_subdirectory(InstrumentationEngine)
 enable_testing()
 add_subdirectory(Tests/CommonLibTests)
 add_subdirectory(Tests/InstrumentationEngineLibTests)
-
--- a/src/Dependencies/NativeDependencies.csproj
+++ b/src/Dependencies/NativeDependencies.csproj
@@ -19,11 +19,11 @@
   <!-- This must be a relative or absolute path, not computed, so that dotnet restore on Linux finds this correctly. -->
   <Import Project="$(MSBuildThisFileDirectory)..\..\build\Common.props" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
-  <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
-  </PropertyGroup>
   <Choose>
     <When Condition="!$([MSBuild]::IsOsPlatform('Windows'))">
+      <PropertyGroup>
+        <TargetFramework>net50</TargetFramework>
+      </PropertyGroup>
       <ItemGroup>
         <PackageReference Include="Microsoft.VisualStudio.Debugger.CoreCLRPAL" Version="$(CoreClrPalVersion)">
           <ExcludeAssets>All</ExcludeAssets>
@@ -31,6 +31,9 @@
       </ItemGroup>
     </When>
     <Otherwise>
+      <PropertyGroup>
+        <TargetFramework>net461</TargetFramework>
+      </PropertyGroup>
       <ItemGroup>
         <PackageReference Include="Microsoft.ApplicationInsights.Agent.Intercept" Version="$(ApplicationInsightsAgentInterceptVersion)">
           <ExcludeAssets>All</ExcludeAssets>
--- a/src/InstrumentationEngine.Lib/Logging.cpp
+++ b/src/InstrumentationEngine.Lib/Logging.cpp
@@ -12,6 +12,29 @@ CSingleton<CLoggerService> CLogging::s_loggerService;
 atomic_size_t CLogging::s_initCount(0);
 atomic<LoggingFlags> CLogging::s_cachedFlags(LoggingFlags_None);
 
+namespace {
+    tstring FormatHex(ULONG32 value)
+    {
+        constexpr size_t width = 10;
+
+        tstring result(width, _T('0'));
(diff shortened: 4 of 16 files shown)


delmyers:
Thanks, @dividedmind, for your contribution. Before we can accept it, we will need you to read and accept our [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/CLRInstrumentationEngine?pullRequest=428). Would you mind going ahead and doing that.

delmyers:
> Thanks, @dividedmind, for your contribution. Before we can accept it, we will need you to read and accept our [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/CLRInstrumentationEngine?pullRequest=428). Would you mind going ahead and doing that.

Never mind. I see that you signed it. Thank you very much!

src/InstrumentationEngine/MethodInfo.cpp
```diff
@@ -1124,28 +1125,28 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::InitializeName(_In_ mdToken
 
     if (!m_genericParameters.empty())
     {
-        tstringstream nameBuilder;
```
delmyers, line 1127: I'm a little bit concerned with what the new performance characteristics of this change could be. Potentially hundres of thousands of method names could be initialized by the instrumentation engine, and doing string concatination like this may be slow.

dividedmind: AFAIK it isn't any different than what stringstream would be doing internally, which is pretty much string appending + formatting with stream semantics (which is actually more work if you think about this).

If you're concerned about this, it might make sense to prereserve some space in the string so it doesn't usually have to reallocate the buffer. However, this can turn out to be a pessimisation especially in the case where most method names fit into a small string buffer (although I don't think this is likely).

WilliamXieMSFT:
Not directly related to this PR, but I took a look at your appmap-dotnet repo and noticed that it mentions using "ProductionBreakpoints_x64.config". While that's fine for internal use, we have a long-standing issue (#60) to refactor this code to support the same logic as Windows (look-up all environment variables with prefix "MicrosoftInstrumentationEngine_ConfigPath32|64_")

src/InstrumentationEngine/MethodInfo.cpp
delmyers: Sorry for the slow follow-up.

You are right. I double-checked  the disassembly for this change, and it shouldn't be any worse that string streams.

I might recommend removing tstringstream altogether, then, so that we don't accidentally use it in locations that this might cause a problem.

delmyers: I think tstringstream might be used in tests, so you could move it into the tests. We can figure out what to do with it as we update the tests to run on other platforms.

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: build
+
+on: [push]
+
+jobs:
```
delmyers, line 5: We can't officially support mac at this time, but it is good to have this build running to ensure that we don't do anything that would prevent us from supporting it in the future.

Could you please add a comment saying that there are no official mac releases, and mac is not being tested?

Would you mind adding a comment here to that effect.

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: build
```
delmyers, line 1: Add comment:

src/InstrumentationEngine/AppDomainCollection.cpp
```diff
@@ -105,7 +105,7 @@ HRESULT MicrosoftInstrumentationEngine::CAppDomainCollection::GetAppDomainIDs(_I
     IfNullRetPointer(pcActual);
     IfNullRetPointer(pAppDomainIDs);
 
-    memset(pAppDomainIDs, cAppDomains, 0);
+    memset(pAppDomainIDs, 0, cAppDomains);
```
delmyers, line 108: > memset(pAppDomainIDs, 0, cAppDomains); [](http://example.com/codeflow?start=3&length=39)

Thanks for catching this.

src/CMakeLists.txt
```diff
@@ -320,33 +322,34 @@ if (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
-        add_custom_target(${target_name}_redefines DEPENDS ${redefine_lib_location})
-
-        add_library(${target_name} STATIC IMPORTED)
-        set_property(TARGET ${target_name} PROPERTY IMPORTED_LOCATION ${redefine_lib_location})
-
-        add_dependencies(${target_name} ${target_name}_redefines)
-    endfunction()
 else()
-    message(FATAL_ERROR "Linux is only supported platform.")
+    add_custom_target(pal_redefines_headers)  # dummy
```
delmyers, line 326: ># dummy [](http://example.com/codeflow?start=46&length=7)

Please be more descriptive in this comment. Thanks.

src/CMakeLists.txt
```diff
@@ -320,33 +322,34 @@ if (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
-
-        add_custom_target(${target_name}_redefines DEPENDS ${redefine_lib_location})
-
-        add_library(${target_name} STATIC IMPORTED)
-        set_property(TARGET ${target_name} PROPERTY IMPORTED_LOCATION ${redefine_lib_location})
-
-        add_dependencies(${target_name} ${target_name}_redefines)
-    endfunction()
 else()
-    message(FATAL_ERROR "Linux is only supported platform.")
```
delmyers, line 348: >message(FATAL_ERROR "Linux is only supported platform.") [](http://example.com/codeflow?start=4&length=56)

Rather than just allowing all builds to go through, could you do something like this:

elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
   message(NOTICE "Building for Darwin is for experimentation purposes")
else()
    message(FATAL_ERROR "CMake build not supported for system '${CMAKE_SYSTEM_NAME}'")
endif()

src/InstrumentationEngine.Lib/Logging.cpp
```diff
@@ -217,7 +217,28 @@ void CLogging::XmlDumpHelper::WriteStringNode(const WCHAR* name, const WCHAR* va
     {
         return;
     }
-    m_result << m_childrenIndent << _T("<") << name << _T(">") << value << _T("</") << name << _T(">") << _T("\r\n");
+    m_result += m_childrenIndent + _T("<") + name + _T(">") + value + _T("</") + name + _T(">") + _T("\r\n");
+}
+
+namespace {
+    tstring format_hex(ULONG32 value, size_t width = 10)
```
delmyers, line 224: >width = 10 [](http://example.com/codeflow?start=45&length=10)

This width is a little unintuitive. It doesn't represent zero padding, it represents the size of the string. I would expect it to represent the padding.

src/InstrumentationEngine.Lib/Logging.cpp
```diff
@@ -217,7 +217,28 @@ void CLogging::XmlDumpHelper::WriteStringNode(const WCHAR* name, const WCHAR* va
...
+        result[1] = _T('x');
+
+        size_t pos = width - 1;
+        while (value != 0) {
+            const auto nibble = value & 0xf;
+            if (nibble < 0xa)
+                result[pos] = _T('0') + nibble;
+            else
+                result[pos] = _T('a') + nibble - 0xa;
+            --pos;
```
delmyers, line 236: >--pos [](http://example.com/codeflow?start=12&length=5)

If width width isn't large enough to contain all of the digits, then pos will underflow, potentially causing a buffer overrun (std::basic_string::operator[] does not do bounds checking)

src/InstrumentationEngine.Lib/Logging.cpp
delmyers: The parameter is never used anyway. Consider removing it so that the underflow issue below can't happen.

delmyers, line 224: >format_hex [](http://example.com/codeflow?start=12&length=10)

put utility functions at the top, please. And use CamelCase for function names.

src/Tests/InstrumentationEngineLibTests/CMakeLists.txt
```diff
@@ -58,14 +65,21 @@ target_link_libraries(
...
-    :libunwind-x86_64.so
-    stdc++fs
     gtest_main
 )
 
+if (CMAKE_SYSTEM_NAME STREQUAL Linux)
+    target_link_libraries(${PROJECT_NAME} stdc++fs)
+elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+    target_link_libraries(${PROJECT_NAME}
+        iconv
```
delmyers, line 75: >iconv [](http://example.com/codeflow?start=8&length=5)

is this going to cause static linkage to libiconv? What is the license of libiconv on a mac?

delmyers (changes requested):
:clock1:

src/Tests/InstrumentationEngineLibTests/CMakeLists.txt
```diff
@@ -28,21 +28,28 @@ add_exe(
...
 # They have nothing to do with one another, each is just a quirk of the environment.
 if(CLR_CMAKE_PLATFORM_ALPINE_LINUX)
     target_link_libraries(${PROJECT_NAME} :libintl.so)
-else()
+elseif(CMAKE_SYSTEM_NAME STREQUAL Linux)
     # NOTE: Explicitly link to the .so version of libdl.so to make sure we don't somehow accidentally
     # start statically linking to these, as these are LGPL components.
     target_link_libraries(${PROJECT_NAME} :libdl.so)
+else()
+    target_link_libraries(${PROJECT_NAME} dl)
```
delmyers, line 52: >dl [](http://example.com/codeflow?start=42&length=2)

similar to below. I haven't tried building these on mac, but can we do something similar to above and link to :libdl.dylib and :libiconv.dylib so that we are explicit in our build system that we are linking agains they dynamic libraries instead of the static ones?

delmyers:
@dividedmind We have left several comments on this PR. Are you able to address them?

dividedmind:
> @dividedmind We have left several comments on this PR. Are you able to address them?

Hi, sorry for dropping the ball, I've been away for a couple of days. Thanks for your review! I'll try to deal with the comments and re-request review once it's ready.

.github/workflows/build.yml
dividedmind: I don't think this config file is in any way copyrightable, but I'll add it.

src/InstrumentationEngine/AppDomainCollection.cpp
dividedmind: No problem. clang did! :)

src/InstrumentationEngine.Lib/Logging.cpp
dividedmind: Good point. I was mimicking setw from the stringstream.

src/Tests/InstrumentationEngineLibTests/CMakeLists.txt
dividedmind: Yes, the linker uses a shared library unless only a static one is available or a `-static` option is given. I don't think it makes sense to try to force dynamic linking somehow. Many systems won't even have a static library; it would take a very unusual configuration to only have a static one. This is especially true of libdl (in fact I don't think I've ever seen a libdl.a).

A CI machine would have shared objects, so it'll get linked dynamically and the binaries can be freely redistributed. If a user has a weird configuration with only static libraries, it's their decision and responsibility not to redistribute the statically linked binary if it's against the license (which I believe is not 100% clear).

BTW, this iconv is for test code which isn't redistributed in a compiled form anyway.

src/Tests/InstrumentationEngineLibTests/CMakeLists.txt
dividedmind: See https://github.com/microsoft/CLRInstrumentationEngine/pull/428#discussion_r695125411

dividedmind:
> Not directly related to this PR, but I took a look at your appmap-dotnet repo and noticed that it mentions using "ProductionBreakpoints_x64.config". While that's fine for internal use, we have a long-standing issue (#60) to refactor this code to support the same logic as Windows (look-up all environment variables with prefix "MicrosoftInstrumentationEngine_ConfigPath32|64_")

Yup, I've noticed that but wanted to take the path of least resistance when starting developing appmap-dotnet :)
It's a good idea, I might take a look since I'm digging in this code anyway.

src/Tests/InstrumentationEngineLibTests/CMakeLists.txt
delmyers: IANAL, but if my understanding is correct, statically linking to an LGPL binary implies that the distribution of the new dependent binary must be redistributed as LGPL... not so if it is statically linked. There is, at least, some controversy around it. We would like to be able to release all of our binaries as MIT and avoid controversy. We would also like to redistribute them under alternate licenses (and allow other developers to redistribute under alternate licenses). In any case, the code remains MIT.

Also, it isn't entirely clear whether or not uploading the resulting binaries as artifacts of a CI counts as "distribution"... so there is that.

All-in-all, I think that we would like to make dynamic linkage as explicit as possible so that we don't accidentally make a mistake in the future.

delmyers:
> > @dividedmind We have left several comments on this PR. Are you able to address them?
> 
> Hi, sorry for dropping the ball, I've been away for a couple of days. Thanks for your review! I'll try to deal with the comments and re-request review once it's ready.

No worries. Thanks for the submission!

src/Tests/InstrumentationEngineLibTests/CMakeLists.txt
dividedmind: IANAL either ;)

FWIW, there doesn't seem to be a way to explicitly instruct ld64 (OSX linker) to only search for dynamic libraries. I tried using `-lazy-l` but it appears to be deprecated and falls back to normal link. I verified that on it indeed links dynamically to libiconv.dylib. Libdl appears to be a compatibility layer and it's not linked directly, instead the calls are replaced by _dl* system builtins (imported dynamically too, probably by the dyld interpreter on execution); it is kind of a special case library, because it really is an interface to the dynamic linker which loads and executes the binary image.

dividedmind:
I think I addressed all the comments (except the discussion re dynamic libraries), let me know what you think and if it looks good I'll rebase off the current main branch to make for a clean merge. Thanks!

dividedmind:
Are these failures caused by https://github.com/microsoft/CLRInstrumentationEngine/pull/428/commits/cbcd890caa9c69aaabc1884a06bafadb85deea16 ? It's weird that they built fine before.

Should I only conditionally change it so it's still `net461` on Windows, or do you just need to push updated net50 versions of dependencies into your repository?

delmyers:
> Are these failures caused by [cbcd890](https://github.com/microsoft/CLRInstrumentationEngine/commit/cbcd890caa9c69aaabc1884a06bafadb85deea16) ? It's weird that they built fine before.
> 
> Should I only conditionally change it so it's still `net461` on Windows, or do you just need to push updated net50 versions of dependencies into your repository?

It is likely due to a change in dotnet that requires us to update our nuget feed. We'll have to look into it on Monday or later in the week.

src/Dependencies/NativeDependencies.csproj
```diff
@@ -20,7 +20,7 @@
   <Import Project="$(MSBuildThisFileDirectory)..\..\build\Common.props" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
+    <TargetFramework>net50</TargetFramework>
```
jander-msft, line 23: If this really needs to be changed to `net50`, then condition it to be `net462` on Windows and `net50` on non-Windows. Otherwise, the builds will have `warning NU1701: Package 'Microsoft.ApplicationInsights.Agent.Intercept 2.4.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework 'net5.0'. This package may not be fully compatible with your project.` on Windows.

jander-msft (changes requested):

jander-msft:
> > Are these failures caused by [cbcd890](https://github.com/microsoft/CLRInstrumentationEngine/commit/cbcd890caa9c69aaabc1884a06bafadb85deea16) ? It's weird that they built fine before.
> > Should I only conditionally change it so it's still `net461` on Windows, or do you just need to push updated net50 versions of dependencies into your repository?
> 
> It is likely due to a change in dotnet that requires us to update our nuget feed. We'll have to look into it on Monday or later in the week.

I've pushed the dependent packages to the CoreClrPal feed.

src/Dependencies/NativeDependencies.csproj
delmyers: @dividedmind will you be able to make this update?

jakubch1:
Any update on this? We are also waiting for those changes and #429  :)

delmyers:
> Any update on this? We are also waiting for those changes and #429 :)

We are still waiting on @dividedmind to finish the requested changes.

src/Dependencies/NativeDependencies.csproj
dividedmind: Done!

dividedmind:
> Any update on this? We are also waiting for those changes and #429 :)

Sorry for being AWOL, I changed the .net reference to be conditional as requested and rebased off of current main.

src/Dependencies/NativeDependencies.csproj
delmyers: Thanks!

jander-msft (approved):
Please merge main into this branch again (to fix the change log conflicts) and then I think this is ready to go!

dividedmind:
> Please merge main into this branch again (to fix the change log conflicts) and then I think this is ready to go!

Done! :)

PR #433, opened 2021-09-08
Mark target tokens as volatile when importing assembly

jander-msft:
The InjectToMscorlib tests are failing on the CI machines whereas they are not failing on a local machine. This issue is that the target tokens (when importing an assembly into mscorlib) are different from the baseline. My theory is that the mscorlib on the CI machines is different (it has more or less TypeDefs, FieldDefs, etc compared to local machine) which causes the imported tokens to have values different from the baseline. I think its best to mark these as volatile so that we don't have to account for variations in mscorlib versions.

Additional changes were made to attach the output xml file when the tests fail to make it easier to determine the full difference between the output and the baseline.


wiktork (approved):

delmyers (approved):
:shipit:

PR #436, opened 2021-09-09
Fix XmlReader.Create to read from stream

WilliamXieMSFT:
Wide characters get URI escaped in XmlReader.Create call. The fix here is to pass to StreamReader (which does not escape the string) and pass the stream to XmlReader.

### DIFF
--- a/src/InstrumentationEngine.Attach/CommandHandler/AttachCommandHandler.cs
+++ b/src/InstrumentationEngine.Attach/CommandHandler/AttachCommandHandler.cs
@@ -109,6 +109,7 @@ namespace Microsoft.InstrumentationEngine
                     readerSettings.ValidationType = ValidationType.Schema;
 
                     // Read configuration source file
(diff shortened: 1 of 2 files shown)


src/InstrumentationEngine.Attach/CommandHandler/AttachCommandHandler.cs
```diff
@@ -109,7 +109,8 @@ internal class AttachCommandHandler
                     readerSettings.ValidationType = ValidationType.Schema;
 
                     // Read configuration source file
-                    using (XmlReader reader = XmlReader.Create(sourceInfo.ConfigSourceFilePath, readerSettings))
+                    using (var stream = new StreamReader(sourceInfo.ConfigSourceFilePath))
```
jander-msft, line 112: Can you add comments here as to why this is being done so that someone doesn't come along later and revert the behavior?

jander-msft (approved):

PR #445, opened 2021-10-04
Remove unused CComVariant from InstrumentationEngine.Lib

schmittjoseph:
- Remove an unused `CComVariant` variable that was resulting in `ClrInstrumentationEngine.so` exporting `clrie_VariantInit`.
- Remove `CComVariant` related code from unix headers so that future instances of this issue should result in a build break.
- Fix typo relating to the Linux build script.


WilliamXieMSFT (approved):
:shipit:

src/unix/inc/atl/atlcomcli.h
```diff
@@ -1470,1495 +1470,1495 @@ class CAdapt
 // CComVariant
```
WilliamXieMSFT, line 1470: >CComVariant [](http://example.com/codeflow?start=3&length=11)

maybe add a comment about not implemented in PAL

src/unix/inc/atl/atlcoll.h
```diff
@@ -238,42 +238,42 @@ class CElementTraits< GUID > :
...
-			return( 1 );
-		}
-		else
-		{
-			ATLASSERT( hr == static_cast<HRESULT>(VARCMP_EQ) || hr == static_cast<HRESULT>(VARCMP_NULL) );
-			return( 0 );
-		}
-	}
-};
+// template<>
```
delmyers, line 241: >//  [](http://example.com/codeflow?start=0&length=3)

Please add a comment indicating that the PAL does not implement CComVariant, so we do not allow the type to be referenced on non-windows platforms.

delmyers (approved):
:shipit:

Issue #443, opened 2021-09-30
HdiCloudK8sServiceTests System.InvalidProgramException : Common Language Runtime detected an invalid program


jakubch1:
We got report about System.InvalidProgramException when using CLR IE. I've checked that issue is reproducible even without any code coverage instrumentation (ShouldInstrument => true, InstrumentMethod => return S_OK). Issue is for method: GetHdiPolicyValidationRequest. Attaching dump logs and full project so you can run it if needed:

[gethdipolicyvalidationrequest.csproj.txt](https://github.com/microsoft/CLRInstrumentationEngine/files/7257796/gethdipolicyvalidationrequest.csproj.txt)
[Tests.cs.txt](https://github.com/microsoft/CLRInstrumentationEngine/files/7257793/Tests.cs.txt)
[ProfilerLog_33052.txt](https://github.com/microsoft/CLRInstrumentationEngine/files/7257798/ProfilerLog_33052.txt)

Issue:
```
      Error Message:
   System.InvalidProgramException : Common Language Runtime detected an invalid program.
  Stack Trace:
     at GetHdiPolicyValidationRequest.PolicyUtilities.GetHdiPolicyValidationRequest(HdiJobMetadata job)
   at GetHdiPolicyValidationRequest.Tests.Test1() in D:\vscodecoverage\artifacts\test\c\5679e3c908c47c930132f4c395c0b5c76a1469e6805df7b05fbb13d22a04525f\Tests.cs:line 14
```

jakubch1:
In this case I've changed our IM to not make any changes. So issue is on CLRIE side.

jander-msft:
There are other ways that an instrumentation method can affect the IL outside of the InstrumenMethod. It can define types, methods, fields, properties, stubs, p/invokes, etc and sometimes write method bodies whole-sale without ever going through InstrumentMethod.

Please compare the difference between the IL of the method from the original assembly vs the IL that CLRIE says it generated to determine what the actual difference is and get a better idea of what is causing the issue.

jakubch1:
This is original IL code:
```
.method public hidebysig static 
	class GetHdiPolicyValidationRequest.ValidationRequestDto GetHdiPolicyValidationRequest (
		class GetHdiPolicyValidationRequest.HdiJobMetadata job
	) cil managed 
{
	// Method begins at RVA 0x20bc
	// Code size 160 (0xa0)
	.maxstack 9
	.locals init (
[...]

PR #449, opened 2021-10-11
Take into account branches in case of stack max size calculation

jakubch1:
Fix for #443 

Few questions based on comment in code:
https://github.com/microsoft/CLRInstrumentationEngine/blob/main/src/InstrumentationEngine/InstructionGraph.cpp#L1094

1) should I remove this comment?
2) should I add similar logic for switches?
3) do you have any framework for testing this max stack size?

### DIFF
--- a/src/InstrumentationEngine/InstructionGraph.cpp
+++ b/src/InstrumentationEngine/InstructionGraph.cpp
@@ -1109,7 +1109,7 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::CalculateMaxStack(_Ou
         }
     };
 
-    // Heuristic to calculate maxstack. This may guess sligthly high.
+    // Heuristic to calculate maxstack. This may guess slightly high.
     // It makes a linear pass over the instructions looking at
     // the stack impact of each instruction and maintaining a theoretical maximum.
     // The actual maximum may be slightly smaller.
--- a/tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/MaxStackSizeExample.cs
+++ b/tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/MaxStackSizeExample.cs
@@ -10,13 +10,13 @@ namespace MaxStackSizeExample
     {
         static int Main(string[] args)
         {
-            GetHdiPolicyValidationRequest(new HdiJobMetadata()
+            GetRequest(new Job()
             {
-                ExperimentId = String.Empty,
+                Id = String.Empty,
                 File = String.Empty,
-                ModuleInfo = new ModuleInfoDto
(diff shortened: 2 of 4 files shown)


jakubch1:
@WilliamXieMSFT @wiktork  I've tested changes on roslyn and repo which discovered the initial issue and changes looks good. Not getting any InvalidProgramExceptions. Please review changes.

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -1091,28 +1091,89 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::CalculateMaxStack(_Ou
...
+                it->second = stackDepth;
+            }
+        }
+        else
+        {
+            stackDepthMap.insert(make_pair(offset, stackDepth));
+        }
+    };
+
+    // Heuristic to calculate maxstack. This may guess sligthly high.
```
wiktork, line 1112: nit: slightly

tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/MaxStackSizeExample.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace MaxStackSizeExample
+{
+    class Program
+    {
+        static int Main(string[] args)
+        {
+            GetHdiPolicyValidationRequest(new HdiJobMetadata()
```
wiktork, line 13: Are these type names from internal code? Did we want to rename them?

wiktork (approved):

src/InstrumentationEngine/InstructionGraph.cpp
jakubch1: fixed

src/InstrumentationEngine/InstructionGraph.cpp
```diff
@@ -1149,14 +1210,26 @@ HRESULT MicrosoftInstrumentationEngine::CInstructionGraph::IsFirstInstructionInC
         DWORD exceptionFlags;
         IfFailRet(pExceptionClause->GetFlags(&exceptionFlags));
 
-        if (exceptionFlags == COR_ILEXCEPTION_CLAUSE_NONE)
+        if (exceptionFlags == COR_ILEXCEPTION_CLAUSE_NONE || exceptionFlags == COR_ILEXCEPTION_CLAUSE_FILTER)
```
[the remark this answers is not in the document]
jakubch1: Filter contains both filter and handler. First if is changing stack size to 1 for handler and second if for filter. So this is needed in both places

PR #451, opened 2021-10-22
Add support for user created buffers for raw profiler hooks

wiktork:
Adds support for 3 different scenarios in raw profiler hooks:
- SetILFunctionBody is called right after DefineMethod during ModuleLoad.
- SetILFunctionBody is called right after DefineMethod during JitCompilationStarted.
- SetILFunctionBody is called with a custom buffer (one not allocated with our FunctionBodyAllocator)

TODO:

- [x] GetInstrumentationResults (test only) may result on having a non-zero buffer.


InstrumentationEngine.sln
```diff
@@ -306,8 +306,8 @@ Global
 		{EAD25B2C-E362-4660-A2FD-AD82B7C1FD77}.CodeAnalysis.Native|x64.ActiveCfg = Release|x64
 		{EAD25B2C-E362-4660-A2FD-AD82B7C1FD77}.CodeAnalysis.Native|x86.ActiveCfg = Release|x86
 		{EAD25B2C-E362-4660-A2FD-AD82B7C1FD77}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{EAD25B2C-E362-4660-A2FD-AD82B7C1FD77}.Debug|x64.ActiveCfg = Debug|x64
-		{EAD25B2C-E362-4660-A2FD-AD82B7C1FD77}.Debug|x64.Build.0 = Debug|x64
```
wiktork, line 310: Revert these back to be bitness specific

src/InstrumentationEngine/MethodInfo.cpp
```diff
@@ -1904,11 +1933,22 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::GetInstrumentationResults(
...
+            cbMethodSize = codeSize + sizeof(IMAGE_COR_ILMETHOD_TINY);
+        }
     }
     else
     {
         pMethodBody = (LPCBYTE)pMethodHeader + sizeof(IMAGE_COR_ILMETHOD_FAT);
         codeSize = ((COR_ILMETHOD_FAT*)pMethodHeader)->GetCodeSize();
+        if (cbMethodSize == 0)
+        {
+            cbMethodSize = codeSize + sizeof(COR_ILMETHOD_FAT);
```
wiktork, line 1950: Need to actually get the size of the header since it can have additional sections for exceptions

src/InstrumentationEngine/CorMethodMalloc.cpp
```diff
@@ -13,6 +13,8 @@ MicrosoftInstrumentationEngine::CCorMethodMalloc::CCorMethodMalloc() : m_cbBuffe
 
 PVOID MicrosoftInstrumentationEngine::CCorMethodMalloc::Alloc(_In_ ULONG cb)
 {
+    //TODO Any consecutive calls to Alloc will destroy the previous buffer.
```
wiktork, line 16: Consider failing subsequent calls or leaking the buffer

wiktork: leak + log

tests/RawProfilerHook/RawProfilerHook/CoRawProfilerHook.cpp
```diff
@@ -122,6 +167,54 @@ HRESULT STDMETHODCALLTYPE CCoRawProfilerHook::FunctionUnloadStarted(
...
+
+    ModuleID moduleId;
+    mdMethodDef methodToken;
+    HRESULT hr = S_OK;
+    IfFailRet(m_pRealProfilerInfo->GetFunctionInfo(functionId, nullptr, &moduleId, &methodToken));
+
+    if (moduleId == m_testModule)
+    {
+        if (m_firstJit)
+        {
```
wiktork, line 179: Extra tests:
- Double Alloc
- Order between IM and Raw maintained
- Order inversion with both modifying the method.

src/InstrumentationEngine.Lib/InstrumentationEngineFeatures.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
```
WilliamXieMSFT, line 1: nit Copyright commentblock

src/InstrumentationEngine/MethodInfo.cpp
```diff
@@ -1211,15 +1212,23 @@ HRESULT MicrosoftInstrumentationEngine::CMethodInfo::SetFinalRenderedFunctionBod
 
     m_bIsInstrumented = true;
 
-    if (m_pFinalRenderedMethod.empty())
+    if (cbMethodSize > 0)
     {
-        CLogging::LogError(_T("CMethodInfo::SetFinalRenderedFunctionBody - final method body should only be called once."));
-        return E_FAIL;
-    }
+        if (!m_pFinalRenderedMethod.empty())
```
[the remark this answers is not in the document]
wiktork: We have very limited raw profiler hook tests at the moment.

src/InstrumentationEngine/CorMethodMalloc.cpp
wiktork: It looks like this is the default code flow; we don't have independent cleanup for this buffer, we rely on subsequent AllocCalls or ref counting to cleanup. I will leave this as is for now.

WilliamXieMSFT (approved):
:shipit:

src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
```diff
@@ -442,32 +442,74 @@ HRESULT MicrosoftInstrumentationEngine::CCorProfilerInfoWrapper::SetILFunctionBo
...
+
+                if (pbNewILMethodHeader != pBuffer)
+                {
+                    CLogging::LogError(_T("Incorrect buffer passed to SetFunctionBodyMalloc"));
+                    return E_FAIL;
+                }
+
+                // NOTE: Relying on the size of the buffer that was handed out by the IMethodMalloc.
+                // this relies on the serialization of jit events by the profiler manager.
+                // Calculating the actual method size would require a fourth parse of header.
```
WilliamXieMSFT, line 474: >fourth [](http://example.com/codeflow?start=70&length=6)

force?

tests/RawProfilerHook/RawProfilerHook/CoRawProfilerHook.cpp
```diff
@@ -122,6 +167,54 @@ HRESULT STDMETHODCALLTYPE CCoRawProfilerHook::FunctionUnloadStarted(
...
+            ASSEMBLYMETADATA asmMeta = { 0 };
+            ZeroMemory(&asmMeta, sizeof(asmMeta));
+            asmMeta.usMajorVersion = 4;
+            mdAssemblyRef mscorlibRef = 0;
+            IfFailRet(m_assemblyEmit->DefineAssemblyRef(publicKeyToken, sizeof(publicKeyToken), L"mscorlib", &asmMeta, nullptr, 0, 0, &mscorlibRef));
+
+            mdTypeRef objectRef = 0;
+            IfFailRet(m_emit->DefineTypeRefByName(mscorlibRef, L"System.Object", &objectRef));
+
+            IfFailRet(m_emit->DefineTypeDef(L"CustomTestType", tdClass, objectRef, nullptr, &m_createdTestType));
```
WilliamXieMSFT, line 193: >tdClass [](http://example.com/codeflow?start=63&length=7)

This is not defined anywhere?

src/InstrumentationEngine/CorProfilerInfoWrapper.cpp
wiktork: Fourth was intended. We parse the header several times but don't need to here since we know the size of the buffer that was allocated.

tests/RawProfilerHook/RawProfilerHook/CoRawProfilerHook.cpp
WilliamXieMSFT: this is part of CorTypeAttr from CorHdr.h

PR #461, opened 2022-01-12
Fix ApiScan 

WilliamXieMSFT:
Update node to 14.x which is required by the latest apiscan task.


WilliamXieMSFT:
(internal) build link: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5620641&view=results

schmittjoseph (approved):

PR #465, opened 2022-02-23
Fix build scripts to use v142 toolsets

WilliamXieMSFT:
Builds were working for VS2019, however there is a bug with having that + VS2022 installed side-by-side where atlbase.h is not found. This is likely due to some intrinsic lookup by MSVC toolsets that conflict or gets confused. Updating the build scripts to point directly to v142 helps mitigate this for users in the future.

Also fixed managed tests to use x64 runsettings as C:\hostedtoolcache\windows\dotnet\dotnet.exe is now failing when using x86.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Please format the changes as follows:
 + Updates:
 
 ## 1.0.44
++ BugFixes: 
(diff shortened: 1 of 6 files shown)


WilliamXieMSFT:
Future considerations include upgrading to v143 toolsets and using latest Windows 10 (or 11) SDK

PR #471, opened 2022-03-01
Support ARM64 for internal pipeline builds

WilliamXieMSFT:
###### Summary 

This change brings in the ability for our internal pipelines to build ARM64 binaries and packages.

### Change Log  

- Add `Nuget.internal.config` which provides an internal feed link and ARM64 sdk
- Update `pipelines.yaml` with ARM64 build if `IsMicroBuildInternal` is set
- Differentiate dotnet restore based on ARM64 platform
- Fixed Packages.sln Release configuration to build only NuGet projects for ARM64

###### Test Methodology

Running the internal build - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5819838&view=results


NuGet.internal.config
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+<configuration>
+    <packageSources>
+        <clear />
+        <add key="DNCENG-CoreClrPal" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/CoreClrPal/nuget/v3/index.json" />
+        <add key="privateDependency" value="https://pkgs.dev.azure.com/devdiv/DevDiv/_packaging/ClrInstrumentationEngine-PrivateDependency/nuget/v3/index.json" />
```
wiktork, line 8: Will this cause issues since we need to have only 1 feed?

src/InstrumentationEngine.Packages.sln
```diff
@@ -64,19 +64,18 @@ Global
...
-		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Debug|ARM64.ActiveCfg = Debug|arm64
-		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Debug|ARM64.Build.0 = Debug|arm64
+		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Debug|ARM64.Build.0 = Debug|ARM64
 		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Debug|x64.ActiveCfg = Debug|x64
 		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Debug|x86.ActiveCfg = Debug|x86
 		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Release|Any CPU.Build.0 = Release|Any CPU
-		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Release|ARM64.ActiveCfg = Release|Any CPU
-		{3D3BA42C-4A1E-4135-B750-4D8CB0A7117F}.Release|ARM64.Build.0 = Release|Any CPU
```
wiktork, line 79: VS tends to change these solution entries automatically and not always to the correct value. Was the conversion from Any CPU to ARM64 intended here?

wiktork (approved):

NuGet.internal.config
WilliamXieMSFT: From https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/how-to-securely-configure-package-source-files, open source projects can have multiple feeds as long as they don't upstream to nuget.org, are within our control, and the packages use reserved namespaces (Microsoft.*) or are trusted authors.

src/InstrumentationEngine.Packages.sln
WilliamXieMSFT: This change is intended; we are building the Headers package differently for AnyCPU (which contains x86 and x64 folders) and ARM64 (which contains an ARM64 folder) since we will publish the AnyCPU Headers package publicly to NuGet.org while keeping the ARM64 headers package internal since we are not ready to announce that CLRIE officially supports ARM64 yet.

build/yaml/steps/windows/dotnetrestore.yaml
```diff
@@ -22,4 +25,18 @@ steps:
       --configfile
       $(Build.SourcesDirectory)\NuGet.Config
       -bl:"$(Build.ArtifactStagingDirectory)\logs\${{ parameters.BinLogCategory }}_Windows_AnyCPU_$(Configuration)_Restore.binlog"
-  condition: ${{ parameters.Condition }}
+  condition: and(succeeded(), eq(variables['Platform'], 'Any CPU'))
+
+- task: DotNetCoreCLI@2
```
jander-msft, line 30: Please conditionally only run this for internal builds. Possibly using IsMicroBuildInternal would be an okay differentiation for this.

jander-msft: And then conditionally DON'T run the prior step. These steps should be either-or, but not both (although running both probably doesn't cause any issues when running internally).

WilliamXieMSFT: The condition will naturally do an either-or due to the platform being ARM64 or AnyCPU (which we do on separate agents). I'll add an additional IsMicroBuildInternal check here

PR #477, opened 2022-03-03
Add SBOM generation

WilliamXieMSFT:
###### Summary 

This change supports generating the Software Bill of Materials (SBOM) on internal builds.

###### Change Log  

- Add PostBuild stage after Build stage that pulls down the Windows-Binaries-Release artifact to generate SBOM
- Upload SBOM as a build artifact

###### Test Methodology

Internal Build - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5827380&view=results


WilliamXieMSFT:
(Relevant to Internal only) Due to our setup with VS we don't actually need these SBOMs; our nuget packages get ingested by the core vsmanifest which generates an SBOM on our behalf. The SBOM generated by these changes is mainly for audit purposes.

build/yaml/pipelines/signed.yaml
```diff
@@ -10,6 +10,7 @@ name: $(date:yyyyMMdd)$(rev:rr)
 
 variables:
   TeamName: ClrInstrumentationEngine
+  Packaging.EnableSBOMSigning: true
```
jander-msft, line 13: Should this just be a variable on the postbuild job instead of all of the jobs in this stage?

build/yaml/jobs/windows/postbuild.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Job Template: PostBuild steps
+
+parameters:
+  SignType: None
```
jander-msft, line 7: Not used in this job.

build/yaml/pipelines/signed.yaml
```diff
@@ -22,11 +23,17 @@ stages:
       Configuration: Release
       IsMicroBuildInternal: true
 
+- stage: PostBuild
+  jobs:
+  - template: ../jobs/windows/postbuild.yaml 
+    parameters:
```
jander-msft, line 29: None of these parameters are part of the postbuild.yaml template except SignType (which isn't used).

jander-msft: Maybe these were accidentally moved from the packages.yaml template inclusion?

build/yaml/steps/microbuild/codesignverify.yaml
```diff
@@ -4,7 +4,8 @@
 # Step Template: MicroBuild Signing and Verification
 
 parameters:
-    SignType: None
+  SignType: None
+  TargetFolder: $(Build.StagingDirectory)
```
jander-msft, line 8: I don't see this parameter being used, Maybe revert this change.

build/yaml/jobs/windows/postbuild.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+- job: PostBuild
```
jander-msft, line 10: My rough reading of the requirements for SBOM generation is that this has to contain everything that is publicly shipped. Arguably, this would include all packages whether they are NuGet, MSI, MSM, etc. This probably should be a post-package job and may only need to consider the packages themselves rather than all binaries; the requirements state that only top-level dependencies are required.

jander-msft (changes requested):

build/yaml/pipelines/signed.yaml
WilliamXieMSFT: Oops, fixing

build/yaml/jobs/windows/postbuild.yaml
WilliamXieMSFT: Leftover artifact, removing

build/yaml/jobs/windows/postbuild.yaml
WilliamXieMSFT: From the guidance:
```
You should run the generator after you have built your binaries but before you have built your installer manifest and setup packages. The SBOM generator task takes an output location (BuildDropPath) as input and generates an SBOM file for that output. The output directory you specify should contain every file contained the in the packages you are planning on inserting into Visual Studio. 
```

I'll refactor this to pull in only the files we put into our packages.

jander-msft: Other guidance that I've seen is `it is okay to generate an SBOM that just lists the NuGet package and not the contents.`, which tells me that the package is what's important, not the individual files within the package. I've sent some email out to hopefully get some clarification on this.

build/yaml/jobs/windows/postbuild.yaml
```diff
@@ -0,0 +1,36 @@
...
+  displayName: PostBuild
+
+  variables:
+    Packaging.EnableSBOMSigning: true
+
+  pool:
+    name: VSEngSS-MicroBuild2019-1ES
+    demands:
```
jander-msft, line 15: I don't think any of these demands are necessary.

build/yaml/jobs/windows/postbuild.yaml
```diff
@@ -0,0 +1,36 @@
...
+    clean: outputs
+
+  steps:
+  - checkout: none
+
+  - template: ../../steps/azuredevops/downloadbuildartifacts.yaml
+    parameters:
+      ArtifactName: binaries-windows-Release
+
+  - template: ../../steps/microbuild/sbom.yaml
```
jander-msft, line 29: We need more than just the Windows binaries; we definitely need the Linux binaries.

jander-msft (changes requested):

build/yaml/jobs/windows/postbuild.yaml
WilliamXieMSFT: removing

build/yaml/jobs/windows/postbuild.yaml
WilliamXieMSFT: assuming we just need to target packages, then I can pull down packages-all which contains both platform files.

WilliamXieMSFT:
SBOM validation: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5848546&view=results

jander-msft (approved):
nit: Rename everything from "PostBuild" to "PostPackage"

WilliamXieMSFT:
Updated SBOM validation: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5849129&view=results

PR #480, opened 2022-03-03
Support Windows ARM64 

WilliamXieMSFT:
###### Summary 

Add support to CLRIE for ARM64 on Windows.

###### Change Log  

- Update the following projects to build with ARM64
  * InstrumentationEngine
  * InstrumentationEngine.Lib
  * ExtensionsCommon
  * InstrumentationEngine.Attach
  * Common.Lib
  * NativeDependencies
  * InstrumentationEngine.Api
- Use internal nuget config to reference private feed hosting an internal ARM64 sdk
- Update build scripts to support ARM64 build (internal only)
- Update internal signed pipeline to support ARM64 
- Create InstrumentationEngine ARM64 nuget package and ARM64-only headers package

###### Test Methodology

Local & pipeline builds


build.ps1
```diff
@@ -210,24 +220,33 @@ if (!$SkipBuild)
...
+        if ($ARM64)
+        {
+            $dotnetRestoreArgs = "$dotnetRestoreArgs /p:IncludeARM64='True'"
+        }
+
+        Invoke-ExpressionHelper -Executable "dotnet" -Arguments $dotnetRestoreArgs -Activity 'dotnet Restore Solutions'
+
+        # NuGet restore disregards platform/configuration
+        $nugetRestoreArgs = "restore `"$repoPath\NativeNugetRestore.sln`" -configfile `"$repoPath\NuGet.config`""
+        Invoke-ExpressionHelper -Executable "nuget" -Arguments $nugetRestoreArgs -Activity 'nuget Restore Solutions'
```
jander-msft, line 234: Why are you using `nuget` instead of `dotnet`? Doesn't the restoration for the `x64` and `x86` platforms still need to take place?

build.ps1
```diff
@@ -210,24 +220,33 @@ if (!$SkipBuild)
...
-        # NuGet restore disregards platform/configuration
         # dotnet restore defaults to Debug|Any CPU, which requires the /p:platform specification in order to replicate NuGet restore behavior.
-        $restoreArgsInit = "restore $repoPath\InstrumentationEngine.sln --configfile $repoPath\NuGet.config"
-        $restoreArgs = @(
-            "$restoreArgsInit /p:platform=`"x86`""
-            "$restoreArgsInit /p:platform=`"x64`""
-            "$restoreArgsInit /p:platform=`"Any CPU`""
-        )
-        Invoke-ExpressionHelper -Executable "dotnet" -Arguments $restoreArgs -Activity 'dotnet Restore Solutions'
+        $dotnetRestoreArgs = "restore `"$repoPath\InstrumentationEngine.sln`" --configfile `"$repoPath\NuGet.config`" /p:platform=`"Any CPU`""
```
jander-msft, line 224: This is now different than what is at line ~270 (other than the solution name). Why is that?

build/CommonBuild.Global.props
```diff
@@ -89,6 +89,7 @@
         <ArchitectureFolderName Condition=" '$(Platform)' == 'x86' ">x86</ArchitectureFolderName>
         <ArchitectureFolderName Condition=" '$(Platform)' == 'Any CPU' ">Any CPU</ArchitectureFolderName>
         <ArchitectureFolderName Condition=" '$(Platform)' == 'AnyCPU' ">AnyCPU</ArchitectureFolderName>
+        <ArchitectureFolderName Condition=" '$(Platform)' == 'ARM64' ">ARM64</ArchitectureFolderName>
```
jander-msft, line 92:
```suggestion
        <ArchitectureFolderName Condition=" '$(Platform)' == 'ARM64' ">arm64</ArchitectureFolderName>
```

src/ExtensionsCommon/NativeInstanceMethodsCodeInjector.cpp
```diff
@@ -103,7 +103,10 @@ namespace Instrumentation
 #if defined(_M_IX86)
 		const auto NativeObjThisPointerElementType = ELEMENT_TYPE_I4;
 #endif
-#if (!defined(_M_X64)) && !defined(_M_IX86)
+#if defined(_M_ARM64)
+        const auto NativeObjThisPointerElementType = ELEMENT_TYPE_I8;
```
jander-msft, line 107: nit: Please line up with rest of code (add more space at beginning).

src/ExtensionsCommon/NativeInstanceMethodsCodeInjector.cpp
```diff
@@ -103,7 +103,10 @@ namespace Instrumentation
...
+#endif
+#if (!defined(_M_X64)) && !defined(_M_IX86) && !defined(_M_ARM64)
```
jander-msft, line 109: Just do `#else`

src/InstrumentationEngine.NuGet.Headers/InstrumentationEngine.Nuget.Headers.csproj
```diff
@@ -2,13 +2,21 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved.
      Licensed under the MIT License. -->
 <Project>
+  <PropertyGroup Condition="'$(Platform)'=='ARM64'">
+    <PackageId>Microsoft.InstrumentationEngine.Headers.ARM64</PackageId>
```
jander-msft, line 6: Are we ever going to publish this package to nuget.org? Is this just a temporary internal package until everything is validated E2E?

src/InstrumentationEngine.NuGet.Headers/InstrumentationEngine.Nuget.Headers.csproj
```diff
@@ -2,13 +2,21 @@
...
+    <NuspecFile>InstrumentationEngine.Headers.nuspec</NuspecFile>
+  </PropertyGroup>
   <PropertyGroup>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <TargetExt>.nupkg</TargetExt>
-    <PackageId>Microsoft.InstrumentationEngine.Headers</PackageId>
     <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
     <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <Platforms>AnyCPU;x64;x86;ARM64</Platforms>
```
jander-msft, line 19:
```suggestion
    <Platforms>AnyCPU;x64;x86;arm64</Platforms>
```

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
```diff
@@ -27,7 +27,7 @@
     <RootNamespace>Microsoft.Diagnostics.InstrumentationEngine.NuGet</RootNamespace>
     <!-- NU5100: Package is used as a means to acquire the engine and not to have its assets referenced directly. -->
     <NoWarn>$(NoWarn);NU5100</NoWarn>
-    <Platforms>AnyCPU;x64;x86</Platforms>
+    <Platforms>AnyCPU;x64;x86;ARM64</Platforms>
```
jander-msft, line 30:
```suggestion
    <Platforms>AnyCPU;x64;x86;arm64</Platforms>
```

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
```diff
@@ -42,7 +42,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
-  <ItemGroup Condition="'$(PlatformOS)'!='Linux'">
+  <ItemGroup Condition="'$(PlatformOS)'!='Linux' and '$(Platform)'!='ARM64'">
```
jander-msft, line 45:
```suggestion
  <ItemGroup Condition="'$(PlatformOS)'!='Linux' and '$(Platform)'!='arm64'">
```

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
```diff
@@ -63,6 +63,8 @@
       <Link>InstrumentationEngine\$(Platform)\Microsoft.Diagnostics.Instrumentation.Extensions.Base.pdb</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+  </ItemGroup>
+  <ItemGroup Condition="'$(PlatformOS)'!='Linux'">
```
jander-msft, line 67: Is arm64 only built for Windows at this time?

src/InstrumentationEngine/InstrumentationEngine.vcxproj
```diff
@@ -37,6 +45,9 @@
   <PropertyGroup Condition="'$(Platform)'=='x64'">
     <TargetName>MicrosoftInstrumentationEngine_x64</TargetName>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Platform)'=='ARM64'">
+    <TargetName>MicrosoftInstrumentationEngine_ARM64</TargetName>
```
jander-msft, line 49:
```suggestion
    <TargetName>MicrosoftInstrumentationEngine_arm64</TargetName>
```

jander-msft: In the long term, we should consolidate the naming so that the architecture is not in the name of the assembly/shared library. No one else does this differentiation. Same with the nuget packages; we should just have one headers package and one impl package with all of the architectures inside them.

jander-msft (changes requested):

src/ExtensionsCommon/NativeInstanceMethodsCodeInjector.cpp
WilliamXieMSFT: Gah tabs vs spaces

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
WilliamXieMSFT: That's correct

src/InstrumentationEngine.NuGet.Headers/InstrumentationEngine.Nuget.Headers.csproj
WilliamXieMSFT: Sort of, we didn't want to modify the currently public headers package due to the blocking internal ARM64 sdk requirement. This header is in case the clients in VS need a separate headers package but long term we hope to integrate it with the currently public headers package once everything gets plumbed through.

build.ps1
WilliamXieMSFT: I've added nuget to mirror our build pipeline, but seems I left off x86/x64 restore. I'll get those readded.

WilliamXieMSFT: Actually, the dotnet restore is only done for AnyCPU so we don't need x86/x64 differentiation.

build.ps1
WilliamXieMSFT: These changes to the build scripts are to mirror the pipeline build. It also looks like the package solution is only using nuget restore, not dotnet restore so I'll modify this accordingly

WilliamXieMSFT:
>         ($anyCpuTests -join ' ') + " /Settings:`"$repoPath\tests\TestSettings\x86.runsettings`"" # AnyCPU

x64

---
Refers to: build.ps1:324 in 70c7300. [](commit_id = 70c7300a34201e075a89c63a69366d13a7e91893, deletion_comment = False)

InstrumentationEngine.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28727.205
+# Visual Studio Version 17
+VisualStudioVersion = 17.2.32201.480
```
wiktork, line 4: Is this intended? I found that it was not possible to update to vs2022 without also updating extensions.base since vs 2022 does not want to build anything below 4.5.2.

We also haven't released 17.2 yet publicly

InstrumentationEngine.sln
```diff
@@ -92,535 +92,719 @@ Global
...
+		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Debug|ARM64.ActiveCfg = Debug|Any CPU
 		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Debug|x64.ActiveCfg = Debug|Any CPU
 		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Debug|x86.ActiveCfg = Debug|Any CPU
 		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Release|Any CPU.Build.0 = Release|Any CPU
+		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Release|ARM64.ActiveCfg = Release|Any CPU
 		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Release|x64.ActiveCfg = Release|Any CPU
 		{DFC9E517-AF1A-4087-BB27-AD1A09602F4F}.Release|x86.ActiveCfg = Release|Any CPU
 		{939151FA-8324-4197-B5B4-EAE927BD2AEE}.CodeAnalysis.Managed|Any CPU.ActiveCfg = Release|Win32
+		{939151FA-8324-4197-B5B4-EAE927BD2AEE}.CodeAnalysis.Managed|ARM64.ActiveCfg = Debug|x64
```
wiktork, line 134: This doesn't seem right.

build.ps1
```diff
@@ -155,13 +159,19 @@ if (-not ($repoPath))
 ###
 # Picks up msbuild from vs2019 installation
 ###
-$VsRequirements = @(
+$VsRequirements = [System.Collections.ArrayList]@(
     'Microsoft.Component.MSBuild'
     'Microsoft.VisualStudio.Workload.NativeDesktop'
     'Microsoft.VisualStudio.Component.VC.14.29.16.11.ATL.Spectre'
```
wiktork, line 165: Any chance we could point these at less specific version numbers? That way if you have something later installed it doesn't fail.

WilliamXieMSFT: VS component ids are quite annoying. These ids are the only ids for v142 toolset in VS2022 (which require versioning). 

If we use the "latest toolset" id, then it's v142 for VS2019 and v143 for VS2022. I could update VSwhere to check different ids for VS versions. That work can be done in a separate PR.

InstrumentationEngine.sln
WilliamXieMSFT: This version header gets updated automatically whenever a new major VS version builds the solution and doesn't actually do anything besides update the icon for the .sln file.

Spoke offline, we'll continue to expect support on VS2019 (our build pools won't change) and will make a larger move to VS2022 separately.

InstrumentationEngine.sln
WilliamXieMSFT: Good point, we should not be building the CodeAnalysis for ARM64 either, I'll fix this

InstrumentationEngine.sln
```diff
@@ -92,535 +92,669 @@ Global
...
 		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Debug|x86.Build.0 = Debug|Win32
 		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Release|Any CPU.ActiveCfg = Release|Win32
+		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Release|ARM64.ActiveCfg = Release|ARM64
 		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Release|x64.ActiveCfg = Release|x64
 		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Release|x64.Build.0 = Release|x64
 		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Release|x86.ActiveCfg = Release|Win32
 		{DAB53F24-D55A-4759-B100-5B7FEFBF6D08}.Release|x86.Build.0 = Release|Win32
 		{8F446E35-5683-4ABF-8F42-1C8EB5797A5B}.CodeAnalysis.Managed|Any CPU.ActiveCfg = Release|Any CPU
 		{8F446E35-5683-4ABF-8F42-1C8EB5797A5B}.CodeAnalysis.Managed|Any CPU.Build.0 = Release|Any CPU
+		{8F446E35-5683-4ABF-8F42-1C8EB5797A5B}.CodeAnalysis.Managed|ARM64.ActiveCfg = Debug|Any CPU
```
WilliamXieMSFT, line 607: >Any CPU [](http://example.com/codeflow?start=86&length=7)

VS Configuration Manager does not like these projects having ARM64 and keeps reverting them so I'll keep these as `Any CPU`

src/InstrumentationEngine/InstrumentationEngine.vcxproj
```diff
@@ -67,7 +78,10 @@
       <SubSystem>Windows</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalLibraryDirectories>$(OutDir);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
-      <AdditionalDependencies>InstrumentationEngine.Lib.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;msxml6.lib;version.lib;mscoree.lib;Faultrep.lib;Shlwapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalLibraryDirectories Condition="'$(Platform)'=='ARM64'">$(WindowsSDK_LibraryPath_arm64);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalDependencies>InstrumentationEngine.Lib.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;msxml6.lib;version.lib;Faultrep.lib;Shlwapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies Condition="'$(Platform)'!='ARM64'">mscoree.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
jander-msft, line 83: Are you sure this works correctly? The previous line has `%(AdditionalDependencies)` and this line has `%(AdditionalDependencies)`. I would find in strange that MSBuild is going to include the `<AdditionalDependencies>` for each defined item on the previous line but only include the redefined one on this line.

Maybe the one on line 82 needs to be renamed to something else and then included in lines 83 and 84 to avoid the ambiguity.

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
```diff
@@ -5,8 +5,8 @@
   <PropertyGroup>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <TargetExt>.nupkg</TargetExt>
-    <NugetSuffix Condition="'$(PlatformOS)'=='Linux'">$(Platform)_$(LinuxDistro)</NugetSuffix>
-    <NugetSuffix Condition="'$(NugetSuffix)'==''">$(Platform)</NugetSuffix>
+    <NugetSuffix Condition="'$(PlatformOS)'=='Linux'">$(Platform.ToLower())_$(LinuxDistro)</NugetSuffix>
```
jander-msft, line 8: If this and PackageId were moved down to the second PropertyGroup, could we use `$(ArchitectureFolderName)` instead of special casing the platform in certain places with `$(Platform.ToLower())`?

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
```diff
@@ -27,7 +27,7 @@
     <RootNamespace>Microsoft.Diagnostics.InstrumentationEngine.NuGet</RootNamespace>
     <!-- NU5100: Package is used as a means to acquire the engine and not to have its assets referenced directly. -->
     <NoWarn>$(NoWarn);NU5100</NoWarn>
-    <Platforms>AnyCPU;x64;x86</Platforms>
+    <Platforms>AnyCPU;x64;x86;arm64</Platforms>
```
jander-msft, line 30: I think this should be `ARM64` to be consistent with the idea that Platform is set as upper case by is lowercase when used in file system naming.

src/InstrumentationEngine.NuGet.Headers/InstrumentationEngine.Nuget.Headers.csproj
```diff
@@ -2,13 +2,21 @@
...
+    <NuspecFile>InstrumentationEngine.Headers.nuspec</NuspecFile>
+  </PropertyGroup>
   <PropertyGroup>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <TargetExt>.nupkg</TargetExt>
-    <PackageId>Microsoft.InstrumentationEngine.Headers</PackageId>
     <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
     <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <Platforms>AnyCPU;x64;x86;arm64</Platforms>
```
jander-msft, line 19: I think this should be `ARM64` to be consistent with the idea that Platform is set as upper case by is lowercase when used in file system naming.

InstrumentationEngine.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28727.205
+# Visual Studio Version 17
+VisualStudioVersion = 17.2.32201.480
 MinimumVisualStudioVersion = 10.0.40219.1
```
WilliamXieMSFT, line 5: >MinimumVisualStudioVersion [](http://example.com/codeflow?start=0&length=26)

Maybe we should require 16 as minimum

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
WilliamXieMSFT: Sure, will fix

src/InstrumentationEngine/InstrumentationEngine.vcxproj
WilliamXieMSFT: I can differentiate for clarity

jander-msft (approved):

InstrumentationEngine.sln
WilliamXieMSFT: We can make that fix in QFW with the other changes to build with VS2022.

src/InstrumentationEngine.NuGet/InstrumentationEngine.Nuget.csproj
```diff
@@ -21,13 +18,16 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <NuspecFile>InstrumentationEngine.nuspec</NuspecFile>
     <NuspecProperties>$(NuspecProperties);scriptdir=$(MSBuildProjectDirectory)</NuspecProperties>
+    <NugetSuffix Condition="'$(PlatformOS)'=='Linux'">$(ArchitectureFolderName)_$(LinuxDistro)</NugetSuffix>
+    <NugetSuffix Condition="'$(NugetSuffix)'==''">$(ArchitectureFolderName)</NugetSuffix>
+    <PackageId>Microsoft.InstrumentationEngine_$(NugetSuffix)</PackageId>
```
WilliamXieMSFT: >PackageId [](http://example.com/codeflow?start=63&length=9)

PackageId needs to be set before the Packaging.props

wiktork (approved):

PR #487, opened 2022-03-17
Fix package restore issues in build.ps1

WilliamXieMSFT:
###### Summary 

Address $env:NUGET_PACKAGES overriding the globalPackagesFolder set in our nuget.config.

###### Change Log  

- Clear the global variable before package restoring
- Fix single-quotes in msbuild properties

###### Test Methodology

Local build

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -252,34 +252,40 @@ if (!$SkipBuild)
         # The $env:NUGET_PACKAGES variable overrides the globalPackagesFolder set in nuget.config.
         # This clears it out temporarily for proper nuget restore behavior.
         $tempGlobalCache = $env:NUGET_PACKAGES
-        $env:NUGET_PACKAGES = $null
 
-        # Clean up bin & obj folder if exists
-        if (Test-Path "$repoPath\bin\$configuration")
+        try
         {
-            Remove-Item -Force -Recurse "$repoPath\bin\$configuration"
-        }
+            $env:NUGET_PACKAGES = $null
 
-        if (Test-Path "$repoPath\obj\")
+            # Clean up bin & obj folder if exists
+            if (Test-Path "$repoPath\bin\$configuration")
+            {
+                Remove-Item -Force -Recurse "$repoPath\bin\$configuration"
+            }
+
+            if (Test-Path "$repoPath\obj\")
+            {
+                Remove-Item -Force -Recurse "$repoPath\obj\"
+            }
+
+            # dotnet restore defaults to Debug|Any CPU, which requires the /p:platform specification in order to replicate NuGet restore behavior.
+            $dotnetRestoreArgs = "restore `"$repoPath\InstrumentationEngine.sln`" --configfile `"$configFile`""
+            if ($ARM64)
+            {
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -242,13 +242,18 @@ Verify-DotnetExists
...
-    $configFile = "$repoPath\NuGet.internal.config"   
+    $configFile = "$repoPath\NuGet.internal.config"
 }
 
 if (!$SkipBuild)
 {
     if (!$SkipCleanAndRestore)
     {
+        # The $env:NUGET_PACKAGES variable overrides the globalPackagesFolder set in nuget.config.
+        # This clears it out temporarily for proper nuget restore behavior.
```
jander-msft, line 253: Why? The point of NUGET_PACKAGES is that a user can direct where the packages are restored. What's the specific error that happens without this change?

WilliamXieMSFT: The problem is that we point $(PackagesDir) to the $(EnlistmentRoot)\packages so builds will fail to resolve when reading from that property. The alternative fix wouild be to respect using NUGET_PACKAGES everywhere.

WilliamXieMSFT: The error we get when building ARM64 is:

LINK : fatal error LNK1104: cannot open file 'E:\ClrInstrumentationEngineGitHub\packages\microsoft.visualstud
       io.arm64.sdk\17.0.0\lib\mscoree.lib' [E:\ClrInstrumentationEngineGitHub\src\InstrumentationEngine\Instrumenta
       tionEngine.vcxproj]

build.ps1
```diff
@@ -264,14 +269,17 @@ if (!$SkipBuild)
...
         }
 
         Invoke-ExpressionHelper -Executable "dotnet" -Arguments $dotnetRestoreArgs -Activity 'dotnet Restore Solution'
 
         # NuGet restore disregards platform/configuration
         $nugetRestoreArgs = "restore `"$repoPath\NativeNugetRestore.sln`" -configfile `"$configFile`""
         Invoke-ExpressionHelper -Executable "nuget" -Arguments $nugetRestoreArgs -Activity 'nuget Restore Solution'
+
+        # Restore global variable
+        $env:NUGET_PACKAGES = $tempGlobalCache
```
schmittjoseph, line 282: Since we're setting `$ErrorActionPreference = 'Stop'` in this script, if any of the above commands fial, such as `dotnet restore`, this will never run and the user's set environment variables will be altered outside of the script's scope.

It'd also conflict with multiple instances of this script running at the same time (even if in different file paths).

I'm not sure if this exists, but if powershell has a way of defining environment variables for child processes that'd be best. If not, consider using a `trap` statement instead to guarantee the user's environment variables are always restored.

wiktork (approved):

build.ps1
WilliamXieMSFT: Good suggestion, from [https://devblogs.microsoft.com/powershell/traps-vs-trycatch/](https://devblogs.microsoft.com/powershell/traps-vs-trycatch/) I'll use a `try-finally` since that provides more granular control. Seems they do similar things.

schmittjoseph (approved):

PR #488, opened 2022-03-17
Fix duplicate instrumentation method check

WilliamXieMSFT:
###### Summary 

CLR IE gracefully skips over instrumentation methods that fail to load. However, our duplicate classid checks on the config file happen before loading method dlls. This causes problems if methods are compiled for different architectures (eg. x64 and ARM64) and only the first config environment variable encountered will be used which may be the wrong arch.

###### Change Log  

- Refactor duplicate check to happen while loading instrumentation methods, rather than on config file

###### Test Methodology

Local debug

### DIFF
--- a/src/InstrumentationEngine/ProfilerManager.cpp
+++ b/src/InstrumentationEngine/ProfilerManager.cpp
@@ -1103,7 +1103,7 @@ HRESULT CProfilerManager::InitializeCore(
     IfFailRet(DetermineClrVersion());
 
     //set event mask to be default one. Later host and Instrumentation methods may call it again
-    IfFailRet(this->SetEventMask(m_dwEventMask));
+    IfFailRet(SetEventMask(m_dwEventMask));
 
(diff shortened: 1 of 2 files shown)


src/InstrumentationEngine/ProfilerManager.cpp
```diff
@@ -860,7 +855,7 @@ HRESULT CProfilerManager::AddInstrumentationMethod(
     CComPtr<IEnumInstrumentationMethodSettings> pSettingsEnum;
     IfFailRet(pSource->EnumSettings(&pSettingsEnum));
 
-    return this->AddInstrumentationMethod(pInstrumentationMethod.release(), pSettingsEnum, ppInstrumentationMethod);
```
cablooshe, line 863: If you're removing `this->` calls can you also remove the one for `SetEventMask` on line 1111?

cablooshe (changes requested):

cablooshe (approved):

src/InstrumentationEngine/ProfilerManager.cpp
WilliamXieMSFT: Yep

PR #500, opened 2022-04-05
removed packages-all artifact

cablooshe:
###### Summary 

Removed references to packages-all in the yaml builds, as we don't need those artifacts anymore.

###### Change Log  

- updated yaml files to remove tasks to move/publish packages-all

###### Test Methodology

testing with pipelines in this pr

### DIFF
--- a/build/yaml/jobs/postpackage.yaml
+++ b/build/yaml/jobs/postpackage.yaml
@@ -19,15 +19,6 @@ jobs:
   steps:
   - checkout: none
 
-  - template: ../steps/azuredevops/downloadbuildartifacts.yaml
-    parameters:
-      ArtifactName: packages-all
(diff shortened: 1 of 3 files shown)


build/yaml/jobs/postpackage.yaml
```diff
@@ -19,15 +19,6 @@ jobs:
   steps:
   - checkout: none
 
-  - template: ../steps/azuredevops/downloadbuildartifacts.yaml
```
WilliamXieMSFT, line 22: We'll need to download from Windows/Linux artifacts here

WilliamXieMSFT (changes requested):

WilliamXieMSFT (approved):
:shipit:

PR #503, opened 2022-04-11
Added debug build and test to signed builds for tests

cablooshe:
###### Summary 

Added build and test stages for debug builds when running signed builds. While it'd be best to just test release builds, there's a problem with some file generation there right now so this is the temporary solution to allow us to test something during the signed builds.

###### Change Log  

- Added BuildDebug and TestDebug steps in signed.yaml, as extra unsigned builds.
- updated tests to always have _MicrosoftInstrumentationEngine_DisableCodeSignatureValidation = 1_ flag when running debug builds as the lack of this flag was causing test failures.
- As a side note: the _SignType: None_ piece of the yaml file is more formality as by default it's already set to none, but figured it would be good to be specific there.

### DIFF
--- a/build/Managed.props
+++ b/build/Managed.props
@@ -21,6 +21,7 @@
 
   <PropertyGroup>
     <DefineConstants Condition="'$(PublicRelease)' != 'True'">$(DefineConstants);ALLOWNOTSIGNED</DefineConstants>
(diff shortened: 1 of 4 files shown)


WilliamXieMSFT (approved):
:shipit:

WilliamXieMSFT:
Consider updating ChangeLog.md with a small description

PR #508, opened 2022-05-23
Assemble IL in Embedded Resources

delmyers:
###### Summary 

Assemble IL in Embedded Resources

###### Change Log  

1) Add ILASM to test project
2) Copy ilasm as an embedded resource into the InstrEngineTests assembly
3) Add il files to embedded resources
4) Refactor the compiler helpers to extract embedded resources as files
5) Invoke ilasm to assemble the embedded il

Note: I think that ILASM is only available for Windows, so we may need
to update our build steps, in the future, to run ILASM for assembly il
files, and embed them into the instrumentation engine tests, or bundle
them as part of the deployment.

###### Test Methodology

Local build, local unit tests. No production code has changed. I manually verified that ILASM is running during tests.

### DIFF
--- a/src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/ByRefLikeInvalidCSharp.il
+++ b/src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/ByRefLikeInvalidCSharp.il
@@ -1,0 +1,631 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+.assembly extern System.Console { .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A ) }
+.assembly extern System.Runtime { .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A ) }
+
+.assembly InvalidCSharp { }
+
+//
+// Begin invalid
+//
+
+.class public sequential ansi sealed beforefieldinit InvalidCSharp.GenericClass_Invalid`1<T>
+    extends [System.Runtime]System.Object
+{
+    .method public hidebysig specialname rtspecialname
+        instance void .ctor () cil managed
+    {
+        ldarg.0
+        call instance void [System.Runtime]System.Object::.ctor()
+        ret
+    }
+}
+
+.class interface public auto ansi abstract InvalidCSharp.GenericInterface_Invalid`1<T>
+{
+}
+
+.class public sequential ansi sealed beforefieldinit InvalidCSharp.GenericValueType_Invalid`1<T>
+    extends [System.Runtime]System.ValueType
+{
+}
+
+.class public sequential ansi sealed beforefieldinit InvalidCSharp.GenericByRefLike_Invalid`1<T>
+    extends [System.Runtime]System.ValueType
+{
+    .custom instance void [System.Runtime]System.Runtime.CompilerServices.IsByRefLikeAttribute::.ctor() = (
+        01 00 00 00
+    )
+}
+
+//
+// End invalid
+//
+
+.class public sequential ansi sealed beforefieldinit InvalidCSharp.GenericClass_Over`1<byreflike T>
+    extends [System.Runtime]System.Object
+{
+    .method public hidebysig specialname rtspecialname
+        instance void .ctor () cil managed
+    {
+        ldarg.0
+        call instance void [System.Runtime]System.Object::.ctor()
+        ret
+    }
+}
+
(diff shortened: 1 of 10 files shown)


delmyers:
@kkeirstead @wiktork @jander-msft Is someone available to review this PR? Thanks.

src/Tests/InstrEngineTests/InstrEngineTests/TestAppAssembler.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace InstrEngineTests
+{
+    internal class TestAppAssembler
+    {
+        internal static bool AssembleFile(IEmbeddedResourceFile embeddedResource, string directoryPath, string assemblyName, bool isDebug, bool is64Bit, bool isExe = false)
+        {
+            ProcessStartInfo psi = new ProcessStartInfo();
+            IEmbeddedResourceFile ilasm = EmbeddedResourceUtils.GetTestResourceFile("ilasm.exe");
+            if (ilasm == null)
```
wiktork, line 22: Should we assert if we can't find ilasm?

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/ByRefLikeInvalidCSharp.il
```diff
@@ -0,0 +1,631 @@
...
+    .method public hidebysig specialname rtspecialname
+        instance void .ctor () cil managed
+    {
+        ldarg.0
+        call instance void [System.Runtime]System.Object::.ctor()
+        ret
+    }
+}
+
+.class interface public auto ansi abstract InvalidCSharp.GenericInterface_Over`1<byreflike T>
```
wiktork, line 58: Do we need to do this in IL, or just use refstructs?

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/ByRefLikeInvalidCSharp.il
```diff
@@ -0,0 +1,631 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+.assembly extern System.Console { .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A ) }
+.assembly extern System.Runtime { .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A ) }
+
+.assembly InvalidCSharp { }
```
wiktork, line 7: Was the name InvalidCSharp meant to suggest this can't be expressed in csharp or is actually invalid? I found this confusing.

src/Tests/InstrEngineTests/InstrEngineTests/TestAppAssembler.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                return false;
+            }
+
+            psi.FileName = ilasm.FilePath;
+            psi.RedirectStandardError = true;
+            psi.RedirectStandardOutput = true;
+            psi.UseShellExecute = false;
+            string debugSwitch = isDebug ? "/DEBUG=OPT" : string.Empty;
+            string bitSwitch = is64Bit ? "/X64" : "/32BITPREFERRED";
```
wiktork, line 32: nit: consider making this more generic architecture, given arm64 support.

wiktork (approved):

src/Tests/InstrEngineTests/InstrEngineTests/TestAppAssembler.cs
delmyers: We can. I think it also asserts on return, but having a finer-grained one would be good, too.

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/ByRefLikeInvalidCSharp.il
delmyers: I inherited the name from Aaron's code. Some of the stuff cannot be expressed in CSharp. Some of it is actually invalid. Some of the tests that are written are to ensure that the jitter throws an exception when it attempts to execute. We could potentially remove those, since we aren't concerned with invalid IL for CLRIE

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/ByRefLikeInvalidCSharp.il
delmyers: Quite possibly, we could just do it in C#. I didn't go to heavily over the actual implementation of these because I just went with what was recommended to me, but we could do a closer audit.

PR #509, opened 2022-06-22
Added Entrypoint to Exercise InvalidCSharp IL

kkeirstead:
###### Summary 
Added C# code to exercise the IL; can now assemble multiple IL files together and compile it with the corresponding C# code.
Builds on #508.

###### Change Log  

- Removed unused IL code paths (not validating that exceptions are thrown).
- Added C# to exercise the InvalidCSharp IL paths.
- Can assemble multiple IL files together and compile with corresponding C# code.
- Miscellaneous clean-up and refactoring.
- Added credits for original InvalidCSharp IL files and C# validation code. 

###### Test Methodology

Local build, local unit tests. No production code has changed. I manually verified that ILASM is running during tests.


src/Tests/InstrEngineTests/InstrEngineTests/TestAppAssembler.cs
```diff
@@ -46,5 +43,46 @@ internal static bool AssembleFile(IEmbeddedResourceFile embeddedResource, string
...
+            string outputType = isExe ? "/EXE" : "/DLL";
+            string extension = isExe ? "exe" : "dll";
+            string outputFile = Path.Combine(directoryPath, FormattableString.Invariant($"{assemblyName}.{extension}"));
+
+            string[] embeddedResourceFilePaths = embeddedResources.Select(x => x.FilePath).ToArray();
+            string ilFileNames = string.Join("\" \"", embeddedResourceFilePaths);
+
+            psi.Arguments = FormattableString.Invariant($"{debugSwitch} {bitSwitch} {outputType} /OUTPUT=\"{outputFile}\" \"{ilFileNames}\"");
+            Process p = Process.Start(psi);
+            p.WaitForExit(5000);
```
WilliamXieMSFT, line 76: >5000 [](http://example.com/codeflow?start=26&length=4)

Do we need this timeout?

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResourceUtils.cs
```diff
@@ -141,11 +141,13 @@ public static void CleanTestResourceFiles()
...
 
-        private static Stream GetEmbeddedResource(string fileName, bool required = true)
+        private static Stream GetEmbeddedResource(string fileName, bool required = true, string alternateResourcesPath = null)
         {
             Assert.IsNotNull(fileName);
 
-            var fullPath = FormattableString.Invariant($"{EmbeddedResourcesPath}.{fileName}");
+            string resourcesPath = string.IsNullOrEmpty(alternateResourcesPath) ? EmbeddedResourcesPath : alternateResourcesPath;
+
+            var fullPath = FormattableString.Invariant($"{resourcesPath}.{fileName}");
```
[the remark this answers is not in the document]
kkeirstead: I think there's something else going on - all the test failures are specifically on net4*, so I'm going to do some investigating to see why those tests are having the issue.

src/Tests/InstrEngineTests/InstrEngineTests/TargetAppCompiler.cs
```diff
@@ -68,50 +69,69 @@ internal class TargetAppCompiler
 
         internal static void CompileTestCode(string directoryPath)
         {
-            CompileCSharpTestCode(directoryPath);
+#if NETCOREAPP
```
jander-msft, line 72: I think this needs to be even stricter... ref fields are not supported in any runtime version other than .NET 7. So this should be `NET7_0_OR_GREATER`.

src/Tests/InstrEngineTests/InstrEngineTests/InstrEngineTests.csproj
```diff
@@ -19,6 +19,9 @@
     <Compile Remove="EmbeddedResources\**" />
     <EmbeddedResource Include="EmbeddedResources\**" />
   </ItemGroup>
+  <ItemGroup>
```
jander-msft, line 22: This project should be updated to include the net70 target framework and only compile or run any of the new ref field testing for .NET7 or newer.

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResourceUtils.cs
```diff
@@ -141,11 +141,13 @@ public static void CleanTestResourceFiles()
             }
         }
 
-        private static Stream GetEmbeddedResource(string fileName, bool required = true)
+        private static Stream GetEmbeddedResource(string fileName, bool required = true, string alternateResourcesPath = null)
```
delmyers, line 144: >string alternateResourcesPath = null [](http://example.com/codeflow?start=89&length=36)

You can just do string resourcesPath = EmbeddedResourcePath. That is probably better, so that way we just end up getting a test failure if someone accidentally passes null or empty (same above).

src/Tests/InstrEngineTests/InstrEngineTests/TestAppAssembler.cs
delmyers: It only happened during debugging. I think it is probably an issue with .NET 6/7 debugging and the Process class. The CLR has special code that runs in the process to handle certain debugging scenarios, and I think they may have introduced a hanging issue here, probably somewhere in the event source code base. That kind of thing has happened before, where the event source takes a lock and gets hung because it is waiting for the debugger to continue. We didn't have the time to look into it, though.

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/InvalidCSharp/readme.txt
```diff
@@ -0,0 +1,6 @@
+﻿-   ByRefLikeInvalidCSharp.il
+    Original source code from the dotnet runtime GitHub repo (https://github.com/dotnet/runtime/blob/4881a639e7c3f27b5a8d2d160e234d8055333cda/src/tests/Loader/classloader/generics/ByRefLike/InvalidCSharp.il)
+-   RefInvalidCSharp.il
+    Original source code from the dotnet runtime GitHub repo (https://github.com/dotnet/runtime/blob/053a472be0143bda0891c715d7b2ce7398538127/src/tests/Loader/classloader/RefFields/InvalidCSharp.il)
+-   RefStructsTests.cs
```
delmyers, line 5: >RefStructsTests.cs [](http://example.com/codeflow?start=4&length=18)

This link is broken. Also, the license in the file indicates that it is from Microsoft, not the dotnet foundation.

src/Tests/InstrEngineTests/InstrEngineTests/InstrEngineTests.csproj
delmyers: That can be done in the next iteration PR, where we actually introduce the tests.

delmyers (changes requested):
:clock1:

src/Tests/InstrEngineTests/InstrEngineTests/EmbeddedResources/InvalidCSharp/readme.txt
kkeirstead: Which link do you think is broken? I just checked both and didn't have issues with either of them.

kkeirstead: My guess is you might've been getting the comma as part of the link; I switched to using `and` just in case that's an issue.

delmyers (approved):
:shipit:

PR #516, opened 2022-07-18
Add tests to execute ref/struct examples

kkeirstead:
###### Summary 

Added tests to execute ref/struct examples; these tests only run for .NET 7 and greater.

###### Change Log  

- Added seven baselines/testscripts for validating ref structs
- Integrated Del's fix so that TestAppRunner finds all the needed assemblies (our IL assembly was previously not being added)

###### Test Methodology

The added tests are passing (using baselines generated by a preliminary run of the test).


delmyers:
> 

There should be no changes to the APIs in this version. Please remove the changes to changes in this directory.

---
Refers to: src/InstrumentationEngine.Api/InstrumentationEngine.h:4 in 903e2be. [](commit_id = 903e2be399d0a588f8835c0b6f30f31768910751, deletion_comment = False)

src/Tests/InstrEngineTests/InstrEngineTests/RefStructsTests.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace InstrEngineTests
+{
+#if NET7_0_OR_GREATER
+    [TestClass]
+    // "dotnet test" runs tests in-place (instead of copying them to a deployment
+    // directory). DeploymentItems that copy from a source path to a destination
+    // that is the same location are not necessary and cause file locking issues
+    // when initializing the tests. Only use DeploymentItem for paths with different
+    // destinations for .NET Core test assemblies.
+#if !NETCOREAPP
```
delmyers, line 15: > !NETCOREAPP [](http://example.com/codeflow?start=3&length=12)

This should never be true, given the above condition.

src/Tests/InstrEngineTests/InstrEngineTests/RefStructsTests.cs
```diff
@@ -0,0 +1,106 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace InstrEngineTests
+{
+#if NET7_0_OR_GREATER
```
delmyers, line 8: >NET7_0_OR_GREATER [](http://example.com/codeflow?start=4&length=17)

Is there a way that we can put this condition in the csproj instead? Just exclude these files from the build?

src/Tests/InstrEngineTests/InstrEngineTests/InstrEngineTests.csproj
```diff
@@ -21,6 +21,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Remove="EmbeddedResources\InvalidCSharp\readme.txt" />
+    <None Remove="EmbeddedResources\InvalidCSharp\RefStructsTestsImpl.il" />
```
delmyers, line 24: ><None Remove="EmbeddedResources\InvalidCSharp\RefStructsTestsImpl.il" /> [](http://example.com/codeflow?start=4&length=72)

Why?

src/Tests/InstrEngineTests/InstrEngineTests/RefStructsTests.cs
delmyers: Looks like you can do:

```xml
<ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">
  <Compile Remove="RefStructsTests.cs" />
<ItemGroup>
```

src/Tests/InstrEngineTests/InstrEngineTests/InstrEngineTests.csproj
kkeirstead: I believe this was added automatically; I can remove it.

src/Tests/InstrEngineTests/InstrEngineTests/TargetAppCompiler.cs
```diff
@@ -103,14 +105,14 @@ private static void AssembleILTestCode(string directoryPath)
 
                 List<IEmbeddedResourceFile> embeddedResources = ilPrefixes.Select(prefix => EmbeddedResourceUtils.GetTestResourceFile(FormattableString.Invariant($"{prefix}.il"), resourcesPath: EmbeddedResourceUtils.InvalidCSharp_EmbeddedResourcesPath)).ToList();
 
+                string assemblyName = ilPrefixes[0];
```
delmyers, line 108: >string assemblyName = ilPrefixes[0] [](http://example.com/codeflow?start=16&length=35)

Please add some comments about why we build only one, and how we choose the name of the assembly.

src/Tests/InstrEngineTests/TestAppRunner/Program.cs
```diff
@@ -23,12 +25,10 @@ static int Main(string[] args)
...
             string assemblyPath = args[0];
+            _resolutionBasePath = Path.GetDirectoryName(assemblyPath);
 
-            using FileStream assemblyStream = new FileStream(assemblyPath, FileMode.Open, FileAccess.Read);
-            using BinaryReader assemblyReader = new BinaryReader(assemblyStream);
-            byte[] assemblyData = assemblyReader.ReadBytes((int)assemblyStream.Length);
-
-            Assembly assembly = AppDomain.CurrentDomain.Load(assemblyData);
+            Assembly assembly = LoadAssemblyFromDisk(assemblyPath);
+            AppDomain.CurrentDomain.AssemblyResolve += OnAssemblyResolve;
```
delmyers, line 31: >AppDomain.CurrentDomain.AssemblyResolve += OnAssemblyResolve [](http://example.com/codeflow?start=12&length=60)

Did you try Assembly.LoadFrom() instead, to see what it does?

src/Tests/InstrEngineTests/TestAppRunner/Program.cs
```diff
@@ -60,7 +60,26 @@ static int Main(string[] args)
...
+
+            assemblyStream = new FileStream(assemblyPath, FileMode.Open, FileAccess.Read);
+            using (assemblyReader = new BinaryReader(assemblyStream))
+            {
+                byte[] assemblyData = assemblyReader.ReadBytes((int)assemblyStream.Length);
+                return AppDomain.CurrentDomain.Load(assemblyData);
+            }
+        }
+
+        private static Assembly OnAssemblyResolve(object sender, ResolveEventArgs args)
```
delmyers, line 78: >OnAssemblyResolve [](http://example.com/codeflow?start=32&length=17)

Please add some comments about why we do this.

src/Tests/InstrEngineTests/TestAppRunner/Program.cs
delmyers: Also, you might be able to set AppDomain.CurrentDomain.PrivateBinPath to _resolutionPath for the assembly loader to automatically load the assemblies.

delmyers (changes requested):
:clock1:

src/Tests/InstrEngineTests/TestAppRunner/Program.cs
kkeirstead: I did experiment with it a little bit, but I wasn't able to get it to work. It's possible that was an error on my part, but I didn't think it was worth investing too much time into it since we have a solution that works.

delmyers: I would really like to know if setting PrivateBinPath would work as well. That way, we wouldn't have to resort to loading the file directly.

src/Tests/InstrEngineTests/TestAppRunner/Program.cs
delmyers: This is OK, but some explanation of *why* the dependent assemblies aren't getting loaded would be good.

delmyers (approved):
:shipit:

PR #521, opened 2022-09-15
Fix Static Analysis issues

WilliamXieMSFT:
###### Summary 

Static Analysis builds were failing for two reasons:
* Roslyn Analyzers reported several issues:
  * CA1014 - Mark assemblies with CLSCompliant
    * Fixed by adding AssemblyInfo.cs to every managed project with `[assembly: CLSCompliant(true)]` attribute 
  * CA1502 - 'ApplicationInsights_RemoveCallbacks' has a cyclomatic complexity of '29'. Rewrite or refactor the code to decrease its complexity below '26'.
    * Fixed by pragma ignore; this code is autogenerated.
  * CA1506 - 'Attach' is coupled with '42' different types from '17' different namespaces. Rewrite or refactor the code to decrease its class coupling below '41'.
    * Fixed by renaming the function to `AttachInstrumentationEngine` 
  * CA1508 - 'stream' is never 'null'. Remove or refactor the condition(s) to avoid dead code. 
    * [Issue with the analyzer](https://github.com/dotnet/roslyn-analyzers/issues/4526)
    * Fixed by pragma ignore
  * Warning: Could not find ruleset file DevDivRoslynRequired.ruleset
    * Added a copy of the ruleset file
* APIScan suppression baselines are no longer respected as of June 30, 2022.
  * Filed false positive ICM tickets (internal)
  * The build will still be broken on APIScan until the tickets are addressed/resolved.

###### Change Log  

- Fixed Static Analysis builds.


delmyers:
>       softwareVersionNum: '$(Build.BuildNumber)'

Now that we are uploading to TSA, this has to actually match the version number of the software, not the build number. We will have to read it from the version text files that are dropped.

---
Refers to: build/yaml/jobs/codeanalysis/apiscan.yaml:66 in 00fc194. [](commit_id = 00fc1943d468b3a8dd64ecf229a3b3bd8df2ab15, deletion_comment = False)

delmyers (changes requested):
:clock1:

WilliamXieMSFT:
>       softwareVersionNum: '$(Build.BuildNumber)'

Before:
    Software: ClrInstrumentationEngine
    Version : 2022091506
    Build   : 6692351

After:
    Software: ClrInstrumentationEngine
    Version : 1.0.46
    Build   : 6699123

---
In reply to: [1248737359](https://github.com/microsoft/CLRInstrumentationEngine/pull/521#issuecomment-1248737359) [](http://example.com/codeflow?ancestors=1248737359,1248737359)

---
Refers to: build/yaml/jobs/codeanalysis/apiscan.yaml:66 in 00fc194. [](commit_id = 00fc1943d468b3a8dd64ecf229a3b3bd8df2ab15, deletion_comment = False)

src/Extensions.Base.Api/BeginCallbacks.cs
```diff
@@ -45,9 +45,11 @@ internal class BCB<T>
 
 }
 
-//Underscore _System assembly will be treated as regular System namespace by ImportModule function of Intstrumentation Engine.
+//Underscore _System assembly will be treated as regular System namespace by ImportModule function of Instrumentation Engine.
 //This is done to avoid conflicts during compilation
+#pragma warning disable CS3008 // Identifier is not CLS-compliant
 namespace _System.Diagnostics
```
delmyers, line 51: So, in other words, the assembly is *not* CLS compliant. So, we shouldn't really mark it as-such should we?

WilliamXieMSFT: Good point, I'll remote the pragma warning

delmyers (approved):
:shipit:

Issue #361, opened 2021-02-09
Build NaglerInstrumentationMethod on Linux


delmyers:

delmyers:
depends on #360 #359 #358 and #357

oakcode:
Hi all, any news here? I am very keen on to try CLRIE on Linux (I already did it on Windows) but I do not know how to do that without NaglerInstrumentationMethod on Linux. Ok, CLRIE is stared on Linux but how I do not have built plugin for CLRIE further testing is impossible.
Are you planning to enable build on Linux soon? Also, if you are open to contributing maybe I can help somehow but probably I will need some information from you first.

Thanks.
All the best

delmyers:
@oakcode There is currently a PR out to get the build portion working. It isn't perfect, because it still has a dependency on a common library, but it is a start. 

Also, we don't have any code that actually exercises the Nagler profiler on Linux, so we don't know for sure whether it works. But, the code that is in the PR #523 should be a good starting point for you.

oakcode:
Hi @delmyers, thanks

I have created empty plugin for CLR IE, just for test, and now it does not look good. Really simple plugin cannot be loaded on linux, so I have to fix it first. But, Nagler is definitely something what I need at the end because it contains many useful examples. 

Thanks again, all the best

delmyers:
@oakcode 

After the PR goes through, the header files in src/unix/inc/mincom and src/unix/inc/mincor will help you a lot to get off the ground. They contain the header files that define the layouts for many CLR/COM types that you need to get started with building instrumentation methods, without needing to get the whole windows SDK

WilliamXieMSFT:
@oakcode #524 may also be a resource for you in building on Linux. 

If you would like to contribute fixes or updates, feel free to fork the repo or open new issues. Thank you for your interest in using CLRIE and please continue reaching out if you have any further questions or suggestions.

oakcode:
Thank you @delmyers and @WilliamXieMSFT 🥇 
You have helped me a lot. I have changed Nagler on the way I needed and now everything works fine on Linux. 
Thank you once again and I hope that I will be able to contribute soon.

delmyers:
@oakcode, glad to help!

PR #523, opened 2022-10-05
Get Nagler building on Linux

delmyers:
###### Summary 

Includes all changes needed to make sure that Nagler builds on Linux. Note: Nagler is not exercised on Linux yet, so there may be bugs.

###### Change Log  

1) Removes dependency on the PAL for Nagler
  1) Copies/modifies a number of headers from the dotnet PAL
  2) Reimplements a couple of standard Windows crt apis
  3) Adds platform independent environment variable reading
  4) Remove unneeded reference to atl in InitOnce
  5) Replace ATL asserts with custom assert code.
2) Update Nagler to run on Linux
  1) Fix string to use "tstring"  and _T( instead of L"
  2) Update XML loading to only multithread on Windows
     (since Windows uses COM for xml reading)
  3) Disable "Inject Assembly" tests on Linux.
3) Add Nagler to the Linux CMake
  1) Update CMake to allow libraries to declare
     when they link to the PAL
  2) Add Nagler to the root cmake.
4) Fix some of the local build scripts

###### Test Methodology

Local build to make sure that compilation occurs on Linux. Pipeline build will run required tests.


delmyers:
fixes #361

src/CMakeLists.txt
```diff
@@ -183,13 +185,17 @@ endmacro(add_lib lib_name type use_redefines hide_symbols sources)
 
 # Adds an executable and optionally adds dependency on generated file pal_redefines.h for
 # projects that depend on the PAL.
-macro(add_exe exe_name use_redefines)
+macro(add_exe exe_name no_pal use_redefines)
```
delmyers, line 188: >no_pal [](http://example.com/codeflow?start=23&length=6)

This doesn't match the code below

src/Tests/InstrEngineTests/NaglerInstrumentationMethod/assertions.h
```diff
@@ -0,0 +1,17 @@
...
+
+// Note, we use const char* because it is easier for cross-plat
+inline void AssertFailed(_In_ const char* message)
+{{
+    #ifdef PLATFORM_UNIX
+    // Just write to standard error
+    fprintf(stderr, "%s", message);
+    #else
+        _ASSERT_EXPR(0, message);
+    #endif
```
delmyers, line 16: The previous implementation threw. We should probably do that.

src/Tests/InstrEngineTests/NaglerInstrumentationMethod/dllmain.cpp
```diff
@@ -80,4 +81,20 @@ STDAPI DLLEXPORT(DllGetClassObject, 12)(_In_ REFCLSID rclsid, _In_ REFIID riid,
     }
 
     return E_NOINTERFACE;
+}
+
+void AssertLogFailure(_In_ const WCHAR* wszError, ...)
+{
```
delmyers, line 87: This is still necessary because of the dependency on Common.Lib/Macros.h for xml.

src/Tests/InstrEngineTests/NaglerInstrumentationMethod/NaglerInstrumentationMethod.cpp
```diff
@@ -59,15 +83,21 @@ HRESULT CInstrumentationMethod::Initialize(_In_ IProfilerManager* pProfilerManag
     m_pProfilerManager = pProfilerManager;
     IfFailRet(m_pProfilerManager->QueryInterface(&m_pStringManager));
 
+#ifdef PLATFORM_UNIX
+    DWORD retVal = LoadInstrumentationMethodXml(nullptr);
+#else
+    // On Windows, xml reading is done in a single-threaded appartment using
```
WilliamXieMSFT, line 89: >appartment [](http://example.com/codeflow?start=60&length=10)

nit apartment

WilliamXieMSFT (dismissed):
:shipit:

src/Tests/InstrEngineTests/NaglerInstrumentationMethod/assertions.h
```diff
@@ -0,0 +1,13 @@
...
+
+
+// Note, we use const char* because it is easier for cross-plat
+inline void AssertFailed(_In_ const char* message)
+{{
+    #ifdef PLATFORM_UNIX
+    // Just write to standard error
+    fprintf(stderr, "%s", message);
+    #else
+        _ASSERT_EXPR(0, message);
```
WilliamXieMSFT, line 11: >_ASSERT_EXPR [](http://example.com/codeflow?start=8&length=12)

This is defined as writing to wide string:
```
#define _ASSERT_EXPR(expr, msg) \
(void)(                                                                                     \
    (!!(expr)) ||                                                                           \
    (1 != _CrtDbgReportW(_CRT_ASSERT, _CRT_WIDE(__FILE__), __LINE__, NULL, L"%ls", msg)) || \
    (_CrtDbgBreak(), 0)                                                                     \
)
```

WilliamXieMSFT: Should we use `_RPT0(0, message)` instead? that maps to `_CrtDbgReport`

src/Tests/InstrEngineTests/NaglerInstrumentationMethod/assertions.h
```diff
@@ -0,0 +1,19 @@
...
+inline void AssertFailed(_In_ const char* message)
+{
+#ifdef DEBUG
+
+#ifdef PLATFORM_UNIX
+    // Just write to standard error
+    fprintf(stderr, "%s", message);
+#else
+    tstring wmessage;
+    SystemString::Convert(message, wmessage);
```
WilliamXieMSFT, line 14: >SystemString::Convert(message, wmessage); [](http://example.com/codeflow?start=4&length=41)

Missing `using "systemstring.h`

delmyers: That's not the issue. "systemstring.h" is included in stdafx. I forgot to qualify SystemString with the namespace.

WilliamXieMSFT (approved):

PR #557, opened 2023-06-01
Increase maxargs

kylereedmsft:
###### Summary 

Increase the max number of args for the Interception Engine callbacks to support functions with more than 6 arguments.
This increases maxargs as much as is possible for Framework 4.6.

###### Change Log  

- Fix typos and add comments to Extensions.Base.API templates
- Change maxargs to 15

###### Test Methodology

Local build. This change should not affect existing callers because they will not have been updated to call them. Existing functions did not change.


delmyers (approved):
:shipit:

WilliamXieMSFT (approved):

kylereedmsft:
Hmm, maybe this isn't going to work after all. I'll take a look, but I'm starting to think that I might just need to support only loading the first N args in the intercept engine instead.

PR #582, opened 2024-06-17
User Azure CLI for ACR login

jander-msft:
###### Summary 
Update Docker Login step to use Azure CLI for ACR login

###### Test Methodology
Test via PR build

### DIFF
--- a/build/yaml/steps/linux/binaries.yaml
+++ b/build/yaml/steps/linux/binaries.yaml
@@ -5,11 +5,13 @@
 
 steps:
 # Login to ACR
-- task: Docker@2
+- task: AzureCLI@2
   displayName: Docker Login
   inputs:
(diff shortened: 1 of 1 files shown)


build/yaml/steps/linux/binaries.yaml
```diff
@@ -5,11 +5,13 @@
 
 steps:
 # Login to ACR
-- task: Docker@2
+- task: AzureCLI@2
   displayName: Docker Login
```
WilliamXieMSFT, line 9: >Docker [](http://example.com/codeflow?start=15&length=6)

nit rename

WilliamXieMSFT (approved):
:shipit:

build/yaml/steps/linux/binaries.yaml
jander-msft: I'm going to leave as-is since it is stilling providing credentials to the docker service.

PR #584, opened 2024-10-17
Update feeds

jander-msft:
###### Summary 

Replace removed feed with existing feed to allow NuGet packages to be restored on Unix builds.

###### Test Methodology

Internal build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10408402&view=results

### DIFF
--- a/src/unix/dependencies/nuget.config
+++ b/src/unix/dependencies/nuget.config
@@ -4,7 +4,8 @@
 <configuration>
   <packageSources>
     <clear />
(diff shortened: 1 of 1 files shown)


wiktork (approved):

clguiman (approved):

WilliamXieMSFT:
There used to be a security requirement to only expose one public feed in a nuget config or only private feeds with public upstream dependencies. Seems this requirement is no longer applicable?

PR #592, opened 2024-12-24
Update dotnetcli domain

richlander:
I don't know if these Dockerfiles are still in use.

Related: https://github.com/dotnet/core/issues/9671


WilliamXieMSFT:
Yep, we have a separate PR out to fix this: https://github.com/microsoft/CLRInstrumentationEngine/pull/588/files

PR #598, opened 2026-06-25
Update TUF keys to fix scorecard analysis

plnelson:
###### Summary 
Fix failure in score card analysis.  This is happening because it was using old TUF keys.  Update to use the latest.

###### Test Methodology
Change is needed to successfully test scorecard analysis


plnelson:
@WardenGnaw

WardenGnaw (approved):

PR #601, opened 2026-07-21
Pin BinSkim task to latest 4.x engine instead of legacy 1.9.5

plnelson:
###### Summary 
Fix BinSkim step to use the latest BinSkim tool

The BinSkim@4 task defaults to downloading Microsoft.CodeAnalysis.BinSkim 1.9.5, a deprecated engine. Add toolVersion: 'Latest' so the current 4.x BinSkim CLI is used.


plnelson:
@wiktork

wiktork (approved):

PR #602, opened 2026-07-21
Change release pipeline to use push command instead of having outputs

plnelson:
###### Summary
Change release pipeline to use push command instead of having outputs


plnelson:
@wiktork

wiktork (approved):

PR #604, opened 2026-07-24
Fix link check - switched to using lychee

plnelson:
###### Summary 
Link check step was randomly failing and was using a deprecated link checker.

Switched to use lychee instead.


plnelson:
@wiktork

wiktork (approved):

PR #608, opened 2026-08-27
Fix signed build

plnelson:
###### Summary 
Fixed signed build

###### Change Log 
Update signing step to use PME workload-identity authentication instead of the legacy ESRP certificate it was using before

###### Test Methodology
Validated signed build succeeds with these changes


plnelson:
@wiktork

wiktork (approved):

wiktork (approved):

